PR #6, opened 2019-03-19
fix issues with current code not compiling

rafael-azevedo:


pkg/controller/account/account_controller.go
```diff
@@ -192,11 +192,11 @@ func getAwsAccountId(client awsclient.Client, awsAccountName string) (*string, e
 			id = accountStatus.Id
 		}
 	}
-
+	return id, nil
 }
 
 // getStsCredentials returns sts credentials for the specified account ARN
-func getStsCredentials(client awsclient.Client, awsAccountId string) (*sts.AssumeRoleOutput, error) {
+func getStsCredentials(client awsclient.Client, awsAccountID string) (*sts.AssumeRoleOutput, error) {
```
jharrington22: I know that the linting complains about this but I don't think teams like UHC are respecting this. Can you check quickly? I'd like to keep things consistent

rafael-azevedo: they use uppercase ID https://github.com/openshift/hive/blob/4c65148a0134dd4d6945d7c6db6099797528730e/pkg/install/generate_test.go

jharrington22:
LGTM :shipit:

PR #9, opened 2019-03-20
Add spec and remove boilerplate code from controller

jharrington22:
This PR adds the Spec for the AccountClaim controller and removes the boilerplate code from the controller code. At the moment this controller will perform no action when there is a change to the accountclaim CRD.


jharrington22:
Merging! :ship:

PR #14, opened 2019-03-21
Account pool controller bug fixes

rafael-azevedo:
This pr removes status from the account object creation. It also sets the accountpool as the owner of the account object. Lastly it adds error handling to the client.create call.


rafael-azevedo:
Added a bug fix to a logging issue in the account controller.

PR #15, opened 2019-03-21
Add account status to AccountPool

jharrington22:
This PR adds 3 status fields to the AccountPool schema so that a user can easily see that status of claimed/unclaimed accounts and the desired pool size.


jharrington22:
Merging! :shipit:

PR #20, opened 2019-03-22
This should be the AWS account ID numeric value

jharrington22:
This PR removes the AWS account name from the AwsAccountID field in the Spec. This field should be filled in by the account controller with the numeric ID for the AWS account once its created.


jharrington22:
Merging!

PR #22, opened 2019-03-25
Paginate through accounts, return err if none found

jharrington22:
The PR ensure we paginate through all AWS accounts and finally that we check to see that we actually found and account or else return an error.


jharrington22:
@rafael-azevedo @fahlmant could you both please re review? This should be quicker though doesn't address the duplicate account issue at the moment.

rafael-azevedo:
LGTM

jharrington22:
@fahlmant @rafael-azevedo sorry another review, obviously slower since we're going to still paginate all accounts it will now return an error if there is duplicate account names.

rafael-azevedo:
LGTM , should the commits be squashed ?

jharrington22:
@rafael-azevedo I'll squash and merge

PR #25, opened 2019-03-27
Finished MVP logic for reconcile function

rafael-azevedo:
This PR ties the logic together in the reconcile function. It also refactors other functions used in the reconcile function.

The `NewClient()` from the `awsclient/client.go` file has been modified to include a token field. 

For now the initial aws credentials used are taken from the environment variables. 
```
"aws_access_key_id"
"aws_secret_access_key"
```
For creating accounts in testing comment line 172 out of the account pool controller and add this line
```
accountName := "EmailUserName-" + uuid
Example:
accountName := "razevedo-" + uuid
```

The following items are hard coded right now.

- IAM user name to be created is "osdManagedAdmin"
- email suffix "redhat.com"
- region "us-east-1"

The IAM user secrets are created in the namespace the operator is running in as "accountName-secret"
```
apiVersion: v1
data:
  aws_access_key_id: "some secret"
  aws_secret_access_key: "some secret"
  aws_user_name: b3NkTWFuYWdlZEFkbWlu
kind: Secret
metadata:
  creationTimestamp: 2019-03-26T22:15:46Z
  name: razevedo-gjqmqd-secret
  namespace: default
  resourceVersion: "425657"
  selfLink: /api/v1/namespaces/default/secrets/razevedo-gjqmqd-secret
  uid: b4233903-5014-11e9-8c8e-0800278a78dc
type: Opaque
```

One question moving forward is should the aws_access_key_id and aws_secret_access_key be stored as plain text in those secrets?

What still needs to be done.
- We need to grab the aws credentials from a secret instead of a env variable
- Refactor logic so that reconcile looping makes more sense and items can come back into the loop at different stages
- Remove hard coded items or move them into constants 
- Create EC2 instances 
- Better testing before attempting to assume role in a new account instead of a sleep function

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -4,6 +4,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
@@ -14,6 +20,7 @@ import (
 	"github.com/openshift/aws-account-operator/pkg/awsclient"
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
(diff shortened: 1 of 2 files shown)


jharrington22:
@rafael-azevedo secrets should have only base64 encoded values

pkg/controller/account/account_controller.go
```diff
@@ -81,44 +104,210 @@ type ReconcileAccount struct {
...
+			return nil, fmt.Errorf("AWS credentials secret %v did not contain key %v",
+				input.secretName, awsCredsSecretIDKey)
+		}
+		secretAccessKey, ok := secret.Data[awsCredsSecretAccessKey]
+		if !ok {
+			return nil, fmt.Errorf("AWS credentials secret %v did not contain key %v",
+				input.secretName, awsCredsSecretAccessKey)
+		}
+
+		fmt.Printf("STRINGS::: %s ::::: %s\n", accessKeyID, secretAccessKey)
```
jharrington22: Probably can remove this for the time being

rafael-azevedo: I have the fix , turns out they are already base64 encoded. Fixing this now

jharrington22:
@rafael-azevedo if you could make that one change well get this merged

PR #27, opened 2019-03-27
Add finalizer to the AWS Account objects

fahlmant:
To ensure the account CRs don't get deleted, this PR adds a finalizer to them on creation


pkg/controller/accountpool/accountpool_controller.go
```diff
@@ -138,6 +139,7 @@ func (r *ReconcileAccountPool) Reconcile(request reconcile.Request) (reconcile.R
 
 	// Create Account CR
 	newAccount := newAccountForCR(request.Namespace)
+	addFinalizer(newAccount, "finalizer.stable.example.com")
```
jharrington22: @fahlmant should we make this `finzalizer.aws.managed.openshift.io`

fahlmant: I'm +1 to that

PR #37, opened 2019-04-08
Copy secret referenced by Account to secret that UHC expects

fahlmant:
This grabs the secret holding the IAM info referenced by the account, and creates a secret with the same data with the name and namespace UHC expects via the accountclaim CR.

WIP because the code is untested

### DIFF
--- a/deploy/crds/aws_v1alpha1_account_crd.yaml
+++ b/deploy/crds/aws_v1alpha1_account_crd.yaml
@@ -42,20 +42,20 @@ spec:
           type: object
         spec:
           properties:
-            awsaccountid:
+            awsAccountID:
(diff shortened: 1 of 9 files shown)


nimrodshn:
Just so you it is clear the secret UHC is expecting looks as follows:
```
kind: "Secret"
apiVersion: "v1"
metadata:
  name: secretName
  namespace: secretNamespace
data:
  aws_access_key_id: foo
  aws_secret_access_key: bar
```

fahlmant:
@nimrodshn Is that secret structure documented somewhere?

nimrodshn:
@fahlmant Not really, this is what hive [expects](https://github.com/openshift/hive/blob/89e7c908c3457bc03992e4042da7cd1a78ee5fb9/config/templates/cluster-deployment-customimageset.yaml#L46).

PR #39, opened 2019-04-11
Add Dockerfile and Makefile

jharrington22:
This PR adds a new Dockerfile and a Makefile to support the deploy scripts


Makefile
```diff
@@ -0,0 +1,50 @@
...
+
+# Build the docker image
+.PHONY: docker-build
+docker-build:
+	$(BUILD_CMD) -t ${IMG} build/.
+
+# Push the docker image
+.PHONY: docker-push
+docker-push:
+	$(DOCKER_CMD) push ${IMG}
```
jmelis: maybe you want a default for `DOCKER_CMD` ?

jharrington22: @jmelis here? https://github.com/openshift/aws-account-operator/pull/39/files#diff-b67911656ef5d18c4ae36cb6741b7965R5

jmelis: but that's BUILD_CMD not DOCKER_CMD?

jharrington22: ha. My eyes need a rest you are a right sorry I'll fix!

jmelis: ah I see, it's a typo it should be s/DOCKER_CMD/BUILD_CMD/

jharrington22: @jmelis it should be different since the one is `docker build` and the other is `docker push` I just got mixed up. All fixed now.

jharrington22:
@jmelis I'll merge if this looks good?

PR #42, opened 2019-04-11
AccountClaim crd name is plural

jharrington22:
Fix missing `s` on the accountclaim CRD within the template.


rafael-azevedo:
LGTM

PR #45, opened 2019-04-15
Rename role_binding and role to cluster_role_binding and cluster_role

fahlmant:
Rename the role_binding and role to cluster_role_binding and cluster_role to reflect their actual types.


jharrington22:
@fahlmant this would need to change too https://github.com/openshift/aws-account-operator/blob/master/hack/generate-operator-bundle.py#L57

rafael-azevedo:
LGTM

PR #47, opened 2019-04-15
Implemented conditions for account controller

rafael-azevedo:
This PR creates the scaffolding for using conditions in the account controller.

### DIFF
--- a/pkg/apis/aws/v1alpha1/account_types.go
+++ b/pkg/apis/aws/v1alpha1/account_types.go
@@ -42,8 +42,9 @@ type AccountStatus struct {
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
-	Claimed bool   `json:"claimed"`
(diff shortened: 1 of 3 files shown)


pkg/apis/aws/v1alpha1/account_types.go
```diff
@@ -69,6 +70,13 @@ type AccountCondition struct {
 // AccountConditionType is a valid value for AccountCondition.Type
 type AccountConditionType string
 
+const (
```
jharrington22: @rafael-azevedo pending too?

pkg/controller/account/account_controller.go
```diff
@@ -41,6 +41,9 @@ const (
 	awsSecretName           = "aws-config"
 	awsAMI                  = "ami-000db10762d0c4c05"
 	awsInstanceType         = "t2.micro"
+	AccountCreating         = "Creating"
```
jharrington22: I think we'll need comments for all of these to keep linting happy?

pkg/apis/aws/v1alpha1/account_types.go
rafael-azevedo: added

pkg/controller/account/account_controller.go
rafael-azevedo: added comments

PR #50, opened 2019-04-15
Update conditions before updating status

rafael-azevedo:
This PR updates account conditions before updating the status in the account controller.


jharrington22:
@rafael-azevedo no need to merge here as it will happen automatically. Also the tests had not passed yet can we make sure we wait for them in the future please.

PR #51, opened 2019-04-16
SREP-1143: Prep for CI pipeline

lisa:
* Adds `config/config.go` (for openshift/osd-operators-registry)
* "Standardizes" `Makefile` (with helpers)
* "Standardizes" Dockerfile
  * Remove extraneous `build/bin` pieces

Signed-off-by: Lisa Seelye <lseelye@redhat.com>

@jharrington22 does this change interfere with app-sre bits?


lisa:
> @lisa re: breaking app-sre, looks like yes:
> 
> https://ci.ext.devshift.net/job/openshift-aws-account-operator-gh-pr-check/6/console
> 
> ```
> make: *** No rule to make target `docker-build'.  Stop.
> ```
> Recommend update Makefile to have a target for this.
> 
> ```
> .PHONY: docker-build
> docker-build: build
> ```
> ALSO need OWNERS file: https://github.com/openshift/dedicated-admin-operator/blob/master/OWNERS

Done...

OWNERS
```diff
@@ -0,0 +1,8 @@
+approvers:
+- aweiteka
+- jewzaam
+- mwoodson
+reviewers:
```
jharrington22: Are the approvers/reviewers just the leads or can I be included? I'm ok if that is not the process :)

PR #52, opened 2019-04-22
Use ENV for docker image name if it exists

jharrington22:
This PR fixes an issue with the app-sre deploy pipeline where it expects an image to be built with the tag "latest". This is set by using the ENV `$IMG` when the git webhook triggers the script `hack/app_sre_build_deploy.sh` , this change will use that ENV if it exists.


jewzaam:
/lgtm

I understand the need for this, but wanted to note this will override the versioned tag we set if `IMG` var is passed in.  Note there is a `latest` tag created as well, so problem is really that `IMG` isn't used and it's needed to put the image in the right place.

jharrington22:
@jewzaam we can clean this up, I'll create a card so that we align the app-sre build process with our conventions. Since those build scripts are called an the reside in the repo we can modify them however we like.

PR #53, opened 2019-04-23
Set pool and claim controllers to use specific namespace for accounts

jharrington22:
The PR sets the pool and claim controllers to use a set namespace for creating and claiming accounts.


jharrington22:
@fahlmant please test.

cc/ @jhernand @nimrodshn

jharrington22:
@fahlmant @rafael-azevedo please re-test.

PR #54, opened 2019-04-23
Bug Fix: accountpool status now correctly shows counts when doing oc get

fahlmant:
The case of the JSONPath must match the field name in the CRD


fahlmant:
@jharrington22 Please review

PR #55, opened 2019-04-24
Added failures anywhere we could return an error

rafael-azevedo:
This pr makes the account states failed when returning and error from the reconcile loop


jharrington22:
@rafael-azevedo what about a retry for each of these operations before returning failed?

PR #59, opened 2019-04-25
Added admin policy attachment to IAM user

rafael-azevedo:
This PR adds the attachment of admin policy to the IAM users that are created for an account.


jharrington22:
cc/ @jhernand

PR #64, opened 2019-05-17
Deploy needed objects for prometheus metrics when the operator comes up

fahlmant:
This PR ensures that when the operator is deployed, all the proper objects are placed so prometheus can scrape the custom metrics.

-The operator creates an aws-account-operator service to expose the metrics over on deploy
-The operator creates an aws-account-operator servicemonitor to tell prometheus how to digest the metrics
-Removes the operator-sdk metrics service for now
-Includes a prometheus-k8s role for the aws-account-operator namespace that has permissions to get, list and watch services, pods and endpoints.
-Includes a prometheus-k8s rolebinding to allow the prometheus-k8s serviceaccount from the openshift-monitoring namespace to use the aformentioned role.

When the operator and the role/rolebindings are deployed, the aws-account-operator namespace must be given the following label to get everything working: openshift.io/cluster-monitoring="true"

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -105,8 +105,15 @@ func main() {
 	}
 
 	// Create Service object to expose the metrics port.
-	s, _ := operatormetrics.GenerateService(8080, "metrics")
+	s, svcerr := operatormetrics.GenerateService(8080, "metrics")
+	if svcerr != nil {
+		log.Error(err, "Error generating metrics service object.")
+	} else {
+		log.Info("Generated metrics service object")
+	}
+
 	sm := operatormetrics.GenerateServiceMonitor(s)
+	log.Info("Generated metrics servicemonitor object")
 	err = mgr.GetClient().Create(context.TODO(), s)
 	if err != nil {
 		log.Error(err, "error creating metrics Service")
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -12,7 +12,8 @@ import (
 	operatormetrics "github.com/openshift/aws-account-operator/pkg/metrics"
 	"github.com/operator-framework/operator-sdk/pkg/leader"
 	"github.com/operator-framework/operator-sdk/pkg/log/zap"
-	"github.com/operator-framework/operator-sdk/pkg/metrics"
+
+	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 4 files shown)


pkg/metrics/service.go
```diff
@@ -0,0 +1,92 @@
...
+						Type:   intstr.Int,
+						IntVal: port,
+					},
+					Name: portName,
+				},
+			},
+			Selector: map[string]string{"name": operatorName},
+		},
+	}
+	return service, nil
```
jharrington22: We should return `err` here and just log it in `cmd/manager/main.go` otherwise the function return type or `error` unnecessary and probably confusing.

cmd/manager/main.go
```diff
@@ -92,16 +93,31 @@ func main() {
...
+
 	// Setup all Controllers
 	if err := controller.AddToManager(mgr); err != nil {
 		log.Error(err, "")
 		os.Exit(1)
 	}
 
 	// Create Service object to expose the metrics port.
-	_, err = metrics.ExposeMetricsPort(ctx, metricsPort)
+	s, _ := operatormetrics.GenerateService(8080, "metrics")
```
jharrington22: See comment RE the function

cmd/manager/main.go
```diff
@@ -92,16 +93,31 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err := monitoringv1.AddToScheme(mgr.GetScheme()); err != nil {
+		log.Error(err, "error registering prometheus monitoring objects")
+		os.Exit(1)
```
jharrington22: Do we not want to kill the operator?

fahlmant: I copied what was above in 93, which is generated by the sdk. Should we follow their pattern or remove?

cmd/manager/main.go
fahlmant: Will add an err check here

pkg/metrics/service.go
fahlmant: Not sure what you mean here. There are two checks for err above, at this point there is no err to return.

jharrington22: Ah, sorry I though they were doing nothing with the error. Ignore.

cmd/manager/main.go
jharrington22: @tparikh whats your opinion here? If we can't register the scheme for the monitoring objects we'll kill the operator.

fahlmant: Fwiw, this is in the approved certman-operator pr

jharrington22: Ok we'll leave this for another discussion.

jharrington22:
@jewzaam @mwoodson can we get a review here please?

PR #65, opened 2019-05-21
Add prometheus role and rolebinding to the generated CSV

fahlmant:
This PR adds the needed prometheus-k8s RBAC objects to the generation of the CSV

### DIFF
--- a/hack/generate-operator-bundle.py
+++ b/hack/generate-operator-bundle.py
@@ -62,6 +62,24 @@ with open('deploy/cluster_role.yaml', 'r') as stream:
             'serviceAccountName': 'aws-account-operator',
         })
 
+# Add prometheus-k8s role to the CSV:
+with open('deploy/prometheus-k8s-role.yaml', 'r') as stream:
+    prom_role = yaml.load(stream)
+    csv['spec']['install']['spec']['clusterPermissions'].append(
+        {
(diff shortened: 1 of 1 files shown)


lisa:
@jewzaam I'm not 100% sure about the saasherder directory. Maybe @jharrington22 or @jmelis ?

jharrington22:
@jewzaam I put a hold on this because of the role binding I knew it wasn't specified in CSV so I wanted to double check, thanks I'll remove it now.

@lisa @jewzaam we don't want this role bound to the service account?

jewzaam:
> @jewzaam I put a hold on this because of the role binding I knew it wasn't specified in CSV so I wanted to double check, thanks I'll remove it now.
> 
> @lisa @jewzaam we don't want this role bound to the service account?

We want this bound to the `prometheus-k8s` SA in `openshift-monitoring`. See: https://github.com/openshift/aws-account-operator/blob/master/deploy/prometheus-k8s-rolebinding.yaml#L9-L12

You need to drop the role and rolebinding yaml in the bundle.  This is done for crds here:
https://github.com/openshift/aws-account-operator/blob/master/hack/generate-operator-bundle.py#L45-L49

Same can be done for any ClusterRole, ClusterRoleBinding, Role, and RoleBinding.

PR #66, opened 2019-05-21
Added account limit checking and better aws error handling

rafael-azevedo:
This pr sets a constant of awslimit which is an int where we will no longer create new accounts, but we will exit without updating the account state to "Failed". This make will mean that this account will still be counted as an unclaimedAccountCount and the accountpool controller will not create a new account. The next reconcile loop will just try to process this account again. 

This pr also adds better error handling for the aws call to create an account.
I moved the code below till after attempting to create an account. That way if we fail the account state has no been modified and we can try again. 

```
// set state creating if the account was able to create
setAccountClaimStatus(reqLogger, currentAcctInstance, "Attempting to create account", awsv1alpha1.AccountCreating, "Creating")
err = r.Client.Status().Update(context.TODO(), currentAcctInstance)
if err != nil {
}
```

When creating an account If we get any of the error codes or failure reasons below. We do not modify the account and fail. This will let the account be retried in the future and will prevent the accountpool controller from creating a new account. If any other error codes appear we set the account to a failed stae. 

```
Error codes :
ErrCodeConstraintViolationException
ErrCodeServiceException
ErrCodeTooManyRequestsException

Failure Reasons:
ACCOUNT_LIMIT_EXCEEDED
INTERNAL_FAILURE
```

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -55,14 +55,14 @@ const (
 
 // Custom errors
 
-// ErrAwsAccountLimit indicates the orgnization account limit has been reached.
-var ErrAwsAccountLimit = errors.New("ACCOUNT_LIMIT_EXCEEDED")
+// ErrAwsAccountLimitExceeded indicates the orgnization account limit has been reached.
+var ErrAwsAccountLimitExceeded = errors.New("AccountLimitExceeded")
 
 // ErrAwsInternalFailure indicates that there was an internal failure on the aws api
-var ErrAwsInternalFailure = errors.New("INTERNAL_FAILURE")
+var ErrAwsInternalFailure = errors.New("InternalFailure")
 
 // ErrAwsFailedCreateAccount indicates that an account creation failed
-var ErrAwsFailedCreateAccount = errors.New("Failed to create account")
+var ErrAwsFailedCreateAccount = errors.New("FailedCreateAccount")
 
 // ErrAwsTooManyRequests indicates that to many requests were sent in a short period
 var ErrAwsTooManyRequests = errors.New("TooManyRequestsException")
@@ -143,21 +143,20 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 		awsRegion:  "us-east-1",
 	})
 	if err != nil {
-		reqLogger.Info("Failed to get AWS client")
+		reqLogger.Error(err, "Failed to get AWS client")
 		return reconcile.Result{}, err
 	}
 
 	// before doing anything make sure we are not over the limit if we are just error
 	accountTotal, err := TotalAwsAccounts(awsSetupClient)
 	if err != nil {
-		reqLogger.Info(err.Error())
+		reqLogger.Info("Failed to get AWS account total from AWS api", "Error", err.Error())
 		return reconcile.Result{}, err
 	}
 
 	if accountTotal >= awsLimit {
-		print(accountTotal)
-		reqLogger.Info("At aws account limit Total Account: %d")
-		return reconcile.Result{}, ErrAwsAccountLimit
+		reqLogger.Error(ErrAwsAccountLimitExceeded, "AWS Account limit reached", "Account Total", accountTotal)
+		return reconcile.Result{}, ErrAwsAccountLimitExceeded
 	}
 
 	// Fetch the Account instance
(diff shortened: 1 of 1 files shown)


pkg/controller/account/account_controller.go
```diff
@@ -120,9 +135,34 @@ type newAwsClientInput struct {
...
+	// before doing anything make sure we are not over the limit if we are just error
+	accountTotal, err := TotalAwsAccounts(awsSetupClient)
+	if err != nil {
+		reqLogger.Info(err.Error())
+		return reconcile.Result{}, err
+	}
+
+	if accountTotal >= awsLimit {
+		print(accountTotal)
+		reqLogger.Info("At aws account limit Total Account: %d")
```
jharrington22: nit picking, capitals for AWS, this should also be an error

pkg/controller/account/account_controller.go
```diff
@@ -120,9 +135,34 @@ type newAwsClientInput struct {
...
+		secretName: awsSecretName,
+		nameSpace:  awsv1alpha1.AccountCrNamespace,
+		awsRegion:  "us-east-1",
+	})
+	if err != nil {
+		reqLogger.Info("Failed to get AWS client")
+		return reconcile.Result{}, err
+	}
+
+	// before doing anything make sure we are not over the limit if we are just error
```
jewzaam: Where does the error bubble up? I.e. how does any SRE team know that this condition is true?

pkg/controller/account/account_controller.go
```diff
@@ -34,6 +34,7 @@ import (
 var log = logf.Log.WithName("controller_account")
 
 const (
+	awsLimit                = 4500
```
jewzaam: This requires all our orgs to support at least `4500` accounts.  What would it take to put this in config somewhere we can tweak in cluster, make this a default value?

pkg/controller/account/account_controller.go
```diff
@@ -52,6 +53,20 @@ const (
...
+var ErrAwsAccountLimit = errors.New("ACCOUNT_LIMIT_EXCEEDED")
+
+// ErrAwsInternalFailure indicates that there was an internal failure on the aws api
+var ErrAwsInternalFailure = errors.New("INTERNAL_FAILURE")
+
+// ErrAwsFailedCreateAccount indicates that an account creation failed
+var ErrAwsFailedCreateAccount = errors.New("Failed to create account")
+
+// ErrAwsTooManyRequests indicates that to many requests were sent in a short period
+var ErrAwsTooManyRequests = errors.New("TooManyRequestsException")
```
jewzaam: Why the 3 schemes for the arg to errors.New?  Can we be consistent and pick one or is there some functional reason?

pkg/controller/account/account_controller.go
jewzaam:
```suggestion
		reqLogger.Error("AWS Account limit reached: %d")
```

pkg/controller/account/account_controller.go
rafael-azevedo: var ErrAwsAccountLimit = errors.New("ACCOUNT_LIMIT_EXCEEDED")
var ErrAwsInternalFailure = errors.New("INTERNAL_FAILURE")
var ErrAwsTooManyRequests = errors.New("TooManyRequestsException")

The above I took the names from aws , but I will standardize

The below was in the code before. Once I make sure no one is checking these error strings I will make this change as well.

pkg/controller/account/account_controller.go
rafael-azevedo: This goes up to the reconcile loop and errors in console.  I am putting in a better logging message that is log.Error(). Is there anything else you'd like to see

pkg/controller/account/account_controller.go
jharrington22: Right now the quickest way to get this in is to hard code it, we can switch it to a config map and then adjust pipelines etc to support the limits late

pkg/controller/account/account_controller.go
jewzaam: Resolving in favor of suggestion in my review.  Thanks for pointing it out :)

pkg/controller/account/account_controller.go
jewzaam: Is there any alert via PD to someone to we know the limit has been reached?  When will SRE be paged is the question.  I don't see a metric updated here that could be used for a prometheus alert.

pkg/controller/account/account_controller.go
jewzaam: Hold on CM, we also need to look at support for multiple orgs.  So this would probably be tied to  that configuration somewhere (new CRD, part of account pool, something else? tbd)

pkg/controller/account/account_controller.go
rafael-azevedo: Understood , We need to look over this code base again for metrics. I think that will be best suited to another PR.

pkg/controller/account/account_controller.go
rafael-azevedo: I agree this should be a separate PR as well.

PR #70, opened 2019-05-24
Update CSV to add prom yamls to bundle output

rafael-azevedo:
This pr moves all Prometheus related yaml files  to its own directory. That way we can copy everything over to the bundle out put dir. 

This will also allow us to add prometheus rules into this directory so that they can be deployed to the cluster where this operator runs.


jewzaam:
Actually this is going to need to be part of app-interface or something else, the metrics for this are scraped from another cluster.

fahlmant:
@jewzaam Can you please take another look?

PR #85, opened 2019-06-21
[WIP] Add AWS account resuse helper scripts and modify account controller

jharrington22:
This PR adds support for reusing AWS accounts, if an account CR contains a populated `spec.awsAccountID` field it will assume the account is already created and move on. The PR also contains some helper scripts to scrub the AWS env of IAM users and policies as well as reset the local ENV.


iamkirkbater:
Do we want to close this PR since it's super old or is this still relevant?

rogbas:
@jharrington22 can we close that PR?

PR #87, opened 2019-07-15
SREP-1751 - Expose metric with total of accounts in PendingVerification

rogbas:
Adding `aws_account_operator_total_aws_account_pending_verification` metric to the operator

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -56,6 +56,10 @@ var (
 		Name: "aws_account_operator_pool_size_vs_unclaimed",
 		Help: "Report the difference between the pool size and the number of unclaimed account CRs",
 	}, []string{"name"})
+	metricTotalAccountPendingVerification = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "aws_account_operator_total_aws_account_pending_verification",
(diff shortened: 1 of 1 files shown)


pkg/metrics/metrics.go
```diff
@@ -56,6 +56,10 @@ var (
 		Name: "aws_account_operator_pool_size_vs_unclaimed",
 		Help: "Report the difference between the pool size and the number of unclaimed account CRs",
 	}, []string{"name"})
+	metricTotalAccountPendingVerification = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "aws_account_operator_total_aws_account_pending_verification",
+		Help: "Report how many accounts are waiting for limit increase in AWS",
```
jharrington22: Does it not wait for the enterprise support too?

rafael-azevedo: Nit pick but Maybe 

"Report the number of accounts are waiting for limit increase in AWS"

rogbas: Yes, both. Will update the metric description

rogbas: updated

rogbas: updated

PR #89, opened 2019-07-16
Move to using operator-custom-metrics library for custom metrics

fahlmant:
Gets rid of `pkg/metrics/service.go` and the logic for bringing up the prometheus metric server. Moves to using github.com/openshift/operator-custom-metrics pkg to register and expose metrics.


cmd/manager/main.go
```diff
@@ -107,8 +108,19 @@ func main() {
 		os.Exit(1)
 	}
 
+	//Create metrics endpoint and register metrics
```
rafael-azevedo: I think all this metric logic and calling of the upstream metric package should be contained inside of our local metrics package. This means we do not need to import 	`metrics "github.com/openshift/aws-account-operator/pkg/metrics"`
just to pass metrics into `metricsServer := custommetrics.NewBuilder().WithPort("9090").WithPath("/metrics").`


We can just define a function in `metrics "github.com/openshift/aws-account-operator/pkg/metrics"` and pass in port and path. This will allow all metrics related logic to be self contained.

cmd/manager/main.go
```diff
@@ -12,7 +12,8 @@ import (
 	"github.com/openshift/aws-account-operator/pkg/apis"
 	"github.com/openshift/aws-account-operator/pkg/controller"
 	"github.com/openshift/aws-account-operator/pkg/credentialwatcher"
-	operatormetrics "github.com/openshift/aws-account-operator/pkg/metrics"
+	metrics "github.com/openshift/aws-account-operator/pkg/metrics"
+	custommetrics "github.com/openshift/operator-custom-metrics/pkg/metrics"
```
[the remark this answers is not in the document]
fahlmant: :+1:

cmd/manager/main.go
fahlmant: I disagree. The operator-custom-metrics API holds all the logic for starting a prometheus endpoint to expose metrics and can exist and be used completely separately from our definitions of metrics. So combining the two confuses the functions of the two APIs.

cmd/manager/main.go
```diff
@@ -107,8 +108,19 @@ func main() {
 		os.Exit(1)
 	}
 
+	//Create metrics endpoint and register metrics
+	metricsServer := custommetrics.NewBuilder().WithPort("9090").WithPath("/metrics").
```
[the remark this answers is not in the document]
fahlmant: :+1:

cmd/manager/main.go
```diff
@@ -12,7 +12,8 @@ import (
 	"github.com/openshift/aws-account-operator/pkg/apis"
 	"github.com/openshift/aws-account-operator/pkg/controller"
 	"github.com/openshift/aws-account-operator/pkg/credentialwatcher"
-	operatormetrics "github.com/openshift/aws-account-operator/pkg/metrics"
+	metrics "github.com/openshift/aws-account-operator/pkg/metrics"
```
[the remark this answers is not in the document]
fahlmant: :+1:

cmd/manager/main.go
```diff
@@ -107,8 +108,19 @@ func main() {
 		os.Exit(1)
 	}
 
+	//Create metrics endpoint and register metrics
+	metricsServer := custommetrics.NewBuilder().WithPort("9090").WithPath("/metrics").
+		WithCollectors(metrics.MetricTotalAWSAccounts).
+		WithCollectors(metrics.MetricTotalAccountCRs).
+		WithCollectors(metrics.MetricTotalAccountCRsUnclaimed).
+		WithCollectors(metrics.MetricTotalAccountCRsClaimed).
+		WithCollectors(metrics.MetricTotalAccountCRsFailed).
```
rogbas: I didn't find the PendingVerificationCount metric in this list of collectors

pkg/metrics/metrics.go
```diff
@@ -16,85 +16,60 @@ package metrics
...
-	metricsList = []prometheus.Collector{
-		metricTotalAWSAccounts,
-		metricTotalAccountCRs,
-		metricTotalAccountCRsUnclaimed,
-		metricTotalAccountCRsClaimed,
-		metricTotalAccountCRsFailed,
-		metricTotalAccountClaimCRs,
-		metricPoolSizeVsUnclaimed,
-		metricTotalAccountPendingVerification,
+	MetricsList = []prometheus.Collector{
```
rogbas: This variable seems to be unused in the operator, is it something you need from the `operator-custom-metrics` pkg?

fahlmant: This will be used in a future change of the custom metrics package coming in the next week or so

cmd/manager/main.go
fahlmant: :+1: will add

cmd/manager/main.go
jharrington22: It makes sense to me that the localmetrics pkg configures and updates the mertics it knows about and registers those with the operator-custom-metrics pkg. This is how I'd expect it to be configured, if we write a function within the localmetrics package to do this for us its likely this implementation would differ from the next operators implementation which we should avoid. We should add this as an example to the operator-custom-metrics README to ensure consistency.

PR #91, opened 2019-07-17
added ready CRs metric

ninataneja:
added metric for reporting on the number of ready accounts CRs


fahlmant:
@ninataneja I think you forgot to save the file when committing. Update with your changes

PR #94, opened 2019-07-18
Fix for pending verification bug: SREP-1698

ALimaRedHat:
Fix for https://jira.coreos.com/browse/SREP-1698
Tested with: @rafael-azevedo

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -117,6 +117,9 @@ var ErrAwsSupportCaseIDNotFound = errors.New("SupportCaseIdNotfound")
 // ErrAwsFailedDescribeSupportCase indicates that the support case describe failed
 var ErrAwsFailedDescribeSupportCase = errors.New("FailedDescribeSupportCase")
 
+// ErrCreateEC2Instance indicates that the CreateEC2Instance function timed out
+var ErrCreateEC2Instance = errors.New("EC2CreationTimeout")
+
 /**
 * USER ACTION REQUIRED: This is a scaffold file intended for the user to modify with their own Controller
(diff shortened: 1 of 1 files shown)


pkg/controller/account/account_controller.go
```diff
@@ -849,18 +850,24 @@ func CreateEC2Instance(client awsclient.Client) (string, error) {
 			MinCount:     aws.Int64(1),
 			MaxCount:     aws.Int64(1),
 		})
-		if runErr == nil {
-			instanceID = *runResult.Instances[0].InstanceId
-			break
+
+		// Return on unexpected errors
+		if runErr != nil {
+			aerr, ok := runErr.(awserr.Error)
```
fahlmant: I'm reading through some other parts of the operator and its getting terse to read. Could you add comments into your PR, especially in this block?

fahlmant: I'm reading through some other parts of the operator and its getting terse to read. Could you add comments into your PR, especially in this block?

ALimaRedHat: @fahlmant This block in particular already has a comment that I think pretty well encapsulates its functionality. I asked for a second, outside opinion and it was agreed. I'd be happy to go through the rest of the code and see how I could better clarify, but I think that it would be appropriate to leave that for another ticket and get this fix pushed as I have already greatly increased the level of commenting. 

I did forget to remove the print blocks used for testing, so those will go away.

rogbas: I agree that at first glance it's not obvious why we do that on line #856

I'd suggest adding a comment addressing that line, explaining we do that so aws API will add more info on top of the error, like the Code we use right below

PR #97, opened 2019-07-19
Initialize all regions with a EC2 instance

jharrington22:
This PR uses a go routine to initialize all supported regions by creating and terminating an EC2 instance in each region. Because we are using a go routine for each creation and monitoring its termination (or error) using a channel the creation of these resources will only take as long and the longest create/terminate action.

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
-	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/aws/aws-sdk-go/service/organizations"
 	"github.com/aws/aws-sdk-go/service/sts"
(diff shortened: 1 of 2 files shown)


pkg/controller/account/ec2.go
```diff
@@ -0,0 +1,196 @@
...
+		go r.InitializeRegion(reqLogger, region, regions[region]["initializationAMI"], ec2Notifications, ec2Errors, creds)
+	}
+
+	// Wait for all the regions to complete
+	for i := 0; i < len(regions); i++ {
+		select {
+		case msg := <-ec2Notifications:
+			reqLogger.Info(msg)
+		case errMsg := <-ec2Errors:
+			reqLogger.Error(errors.New(errMsg), errMsg)
```
rogbas: This is the point where I mentioned a failed EC2 creation in a region wouldn't be noticed, because we don't return any error.

pkg/controller/account/account_controller.go
```diff
@@ -73,22 +72,52 @@ const (
...
-	"eu-west-3",
-	"ap-northeast-1",
-	"ap-northeast-2",
-	"ap-south-1",
-	"ap-southeast-1",
-	"ap-southeast-2",
-	"sa-east-1",
+var coveredRegions = map[string]map[string]string{
+	"us-east-1": {
+		"initializationAMI": "ami-000db10762d0c4c05",
```
fahlmant: These would be better suited for a config map. We have other hard coded values in the operator that could go in a configmap too, so if you want to commit this now and wait for a future refactor, I think that would be fine.

pkg/controller/account/ec2.go
fahlmant: I think this whole block starting at 31 could use some comments but especially this line.

pkg/controller/account/ec2.go
```diff
@@ -0,0 +1,196 @@
...
+	// Make sure we close our channels when we're done
+	defer close(ec2Notifications)
+	defer close(ec2Errors)
+
+	for region := range regions {
+		go r.InitializeRegion(reqLogger, region, regions[region]["initializationAMI"], ec2Notifications, ec2Errors, creds)
+	}
+
+	// Wait for all the regions to complete
+	for i := 0; i < len(regions); i++ {
```
fahlmant: Not sure of the best solution here, but would it make sense to parallelize this? Seems slow to go through each region and init one at a time.

pkg/controller/account/ec2.go
```diff
@@ -0,0 +1,196 @@
...
+		}
+	}
+
+	reqLogger.Info("Completed initializing all supported regions")
+
+	return nil
+}
+
+// InitializeRegion sets up a connection to the AWS `region` and then creates and terminates an EC2 instance
+func (r *ReconcileAccount) InitializeRegion(reqLogger logr.Logger, region string, ami string, ec2Notifications chan string, ec2Errors chan string, creds *sts.AssumeRoleOutput) error {
```
fahlmant: Would it make sense to wrap some of the items passed to this function into a struct? A lot of the same args are passed to `BuildandDestroyEC2Instances`

pkg/controller/account/ec2.go
jharrington22: @fahlmant So all regions are being initialized in parallel all this loop is doing is waiting to receive notifications from all regions that they are complete.

pkg/controller/account/account_controller.go
jharrington22: Agree, I'll create a card for this and we can refactor later.

pkg/controller/account/ec2.go
jharrington22: @rogbas we log the error here but we do not fail the account because of it, is that what you mean?

pkg/controller/account/ec2.go
jharrington22: Agree but I think we can refactor this after

jharrington22:
@fahlmant @rogbas changes made and rebased because of #94

pkg/controller/account/ec2.go
fahlmant: Ah I see now. Resolving

pkg/controller/account/ec2.go
fahlmant: :+1;

pkg/controller/account/account_controller.go
fahlmant: :+1:

pkg/controller/account/ec2.go
rogbas: Yes, exactly. If we wanted to error out (returning from the function), shoe could consider something like:
```go
		select {
		case msg := <-ec2Notifications:
			reqLogger.Info(msg)
		case errMsg := <-ec2Errors:
			err = errors.New(errMsg)
			reqLogger.Error(err, errMsg)
			return err
		}
```

jharrington22:
@rogbas @fahlmant could you take one last look please, we will always attempt to terminate any instance we create now.

PR #102, opened 2019-07-25
Better AWS error handling

ghost:
~~**This PR is a WIP of ticket:** https://jira.coreos.com/browse/SREP-1737~~
I made some changes to some AWS logs output. These were the basic and easy to spot changes.
I would like some feedback from you guys on other ways/locations I should be improving the error handling.

@rogbas @jharrington22 @rafael-azevedo

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -333,7 +333,7 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 
 	if (currentAcctInstance.Status.State == "") && (currentAcctInstance.Status.Claimed == false) {
 		// before doing anything make sure we are not over the limit if we are just error
-		accountTotal, err := TotalAwsAccounts(awsSetupClient)
+		accountTotal, err := TotalAwsAccounts(reqLogger, awsSetupClient)
 		if err != nil {
 			reqLogger.Info("Failed to get AWS account total from AWS api", "Error", err.Error())
 			return reconcile.Result{}, err
@@ -431,7 +431,12 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 		if err != nil {
 			var returnErr error
 			if aerr, ok := err.(awserr.Error); ok {
-				reqLogger.Error(returnErr, fmt.Sprintf("Unable to create AWS connection with SRE credentials, AWS Error Message: %s", aerr.Message()))
+				reqLogger.Error(returnErr,
+					fmt.Sprintf(`Unable to create AWS connection with SRE credentials, 
+						AWS Error Code: %s, 
+						AWS Error Message: %s`,
+						aerr.Code(),
+						aerr.Message()))
 			}
 			return reconcile.Result{}, err
 		}
@@ -476,7 +481,7 @@ func (r *ReconcileAccount) BuildAccount(reqLogger logr.Logger, awsClient awsclie
 	reqLogger.Info("Creating Account")
 
 	email := formatAccountEmail(account.Name)
-	orgOutput, orgErr := CreateAccount(awsClient, account.Name, email)
+	orgOutput, orgErr := CreateAccount(reqLogger, awsClient, account.Name, email)
 	// If it was an api or a limit issue don't modify account and exit if anything else set to failed
 	if orgErr != nil {
 		switch orgErr {
@@ -518,7 +523,7 @@ func (r *ReconcileAccount) BuildIAMUser(reqLogger logr.Logger, awsClient awsclie
 	reqLogger.Info("IAM User Created")
 
 	// Create IAM user
-	_, userErr := CreateIAMUser(awsClient, iamUserName)
+	_, userErr := CreateIAMUser(reqLogger, awsClient, iamUserName)
 	// TODO: better error handling but for now scrap account
 	if userErr != nil {
 		failedToCreateIAMUserMsg := fmt.Sprintf("Failed to create IAM user %s", iamUserName)
@@ -533,7 +538,7 @@ func (r *ReconcileAccount) BuildIAMUser(reqLogger logr.Logger, awsClient awsclie
 
 	reqLogger.Info(fmt.Sprintf("Attaching Admin Policy to IAM user %s", iamUserName))
 	// Setting user access policy
-	_, policyErr := AttachAdminUserPolicy(awsClient, iamUserName)
+	_, policyErr := AttachAdminUserPolicy(reqLogger, awsClient, iamUserName)
 	if policyErr != nil {
 		failedToAttachUserPolicyMsg := fmt.Sprintf("Failed to attach admin policy to IAM user %s", iamUserName)
 		SetAccountStatus(reqLogger, account, failedToAttachUserPolicyMsg, awsv1alpha1.AccountFailed, "Failed")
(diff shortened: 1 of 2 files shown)


pkg/controller/account/account_controller.go
```diff
@@ -373,7 +373,7 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 		if err != nil {
 			var returnErr error
 			if aerr, ok := err.(awserr.Error); ok {
-				reqLogger.Error(returnErr, fmt.Sprintf("Unable to create AWS connection with SRE credentials, AWS Error Message: %s", aerr.Message()))
+				reqLogger.Error(returnErr, fmt.Sprintf("Unable to create AWS connection with SRE credentials, AWS Error Code: %s, AWS Error Message: %s", aerr.Code(), aerr.Message()))
```
rogbas: You addressed the log msgs, by adding Code and additional info to them, but I understand another point of this card is to handle specific AWS Errors (like we do in checkcaseResolution function), instead of just checking if err != nil for success.

ALimaRedHat: @rogbas, I talked to @jharrington22 and @fahlmant and they indicated that it would be an ongoing process since we need to see possible errors in order to efficiently build cases for them. If you have insight on errors we know of I can definitely start adding those.

rogbas: ok

c-e-brumm: If the messages and codes are going to be followed up on later is it worth adding a keyword to make extracting them from the rest of the log easier? Maybe the message body already well known and unique enough. Maybe this operator log isn't very chatty? Just thinking aloud.

jharrington22: The main concern of this card is that we were not reporting the AWS error code and message when we get an error from the AWS API. This makes debugging and handing those errors in the future hard if we don't know what they are.

@c-e-brumm The errors will include the account name which is unique, so if we have an issue with a particular account we'll be able to extract the messages fairly easily. Good thought though.

jharrington22: Its hard to predict what errors we'll want to handle for each service so logging them will allow us to easily handle them in the future.

rogbas: Having clarified that it's going to be a gradual process of matching to specific AWS Errors, I see other places where the error msg could be improved to add aws Code. Everytime we call a function from the `awsclient.Client` interface, examples: AttachAdminUserPolicy, CreateUserAccessKey.

jharrington22: Hopefully @ALimaRedHat will catch them all, i've asked that we look at every AWS API call and ensure we're bubbling up the code and message. This is still a WIP but feel free to help point them out.

rogbas: @jharrington22 probably not part of that card, but we could start addressing some of the errors related to aws rate limiting us

jharrington22: @rogbas :100: - thats pretty much the intent of this card, lets see which are real errors and where we are being rate limited.

ALimaRedHat:
Updated the code to log errors received from calls to the `awsclient.Client` interface. @rogbas could you take a look?

pkg/controller/account/account_controller.go
```diff
@@ -607,14 +613,57 @@ func (r *ReconcileAccount) BuildIAMUser(reqLogger logr.Logger, awsClient awsclie
 	return userSecret.ObjectMeta.Name, nil
 }
 
+<<<<<<< HEAD
```
jharrington22: @ALimaRedHat you have conflict markers in here

pkg/controller/account/account_controller.go
```diff
@@ -607,14 +613,57 @@ func (r *ReconcileAccount) BuildIAMUser(reqLogger logr.Logger, awsClient awsclie
...
+		return err
+	}
+
+	reqLogger.Info("EC2 Instance Terminated")
+
+	return nil
+}
+
+=======
+>>>>>>> 374f9ecc032626647c9fb4c810866b599a2df6de
```
jharrington22: Here as well

pkg/controller/account/account_controller.go
```diff
@@ -830,8 +928,110 @@ func formatAccountEmail(name string) string {
 	return email
 }
 
+<<<<<<< HEAD
```
jharrington22: Here too, also this code gets moved

pkg/controller/account/account_controller.go
```diff
@@ -830,8 +928,110 @@ func formatAccountEmail(name string) string {
...
+					aerr.Message()))
+		}
+		return err
+	}
+
+	return nil
+}
+
+=======
+>>>>>>> 374f9ecc032626647c9fb4c810866b599a2df6de
```
jharrington22: Here too

pkg/controller/account/ec2.go
```diff
@@ -163,7 +163,7 @@ func CreateEC2Instance(reqLogger logr.Logger, client awsclient.Client, ami strin
 
 }
 
-// DescribeEC2Instances returns the InstanceState code
+//DescribeEC2Instances returns the InstanceState code
```
jharrington22: Nit, add a space here before the comment wording

pkg/controller/account/account_controller.go
```diff
@@ -325,7 +325,8 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 
 	if (currentAcctInstance.Status.State == "") && (currentAcctInstance.Status.Claimed == false) {
 		// before doing anything make sure we are not over the limit if we are just error
-		accountTotal, err := TotalAwsAccounts(awsSetupClient)
+		reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
```
jharrington22: Why did we add this again its already declared here: https://github.com/openshift/aws-account-operator/blob/041603a044a4238565ab79cc60ccb96082ae00ea/pkg/controller/account/account_controller.go#L236

sam-nguyen7: +1 this line should be removed. Probably a mistake.

sam-nguyen7:
@jharrington22 PR lgtm

pkg/controller/account/account_controller.go
```diff
@@ -660,35 +672,51 @@ func CreateAccount(client awsclient.Client, accountName, accountEmail string) (*
...
 	var createUserOutput *iam.CreateUserOutput
-	if ok && awserr.Code() == iam.ErrCodeNoSuchEntityException {
-		createResult, err := client.CreateUser(&iam.CreateUserInput{
-			UserName: aws.String(userName),
-		})
-		if err != nil {
-			return &iam.CreateUserOutput{}, err
+	// handle errors
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
```
rafael-azevedo: should we return an error if this line is !ok ?

jharrington22: Good catch, yes we should

pkg/controller/account/account_controller.go
```diff
@@ -660,35 +672,53 @@ func CreateAccount(client awsclient.Client, accountName, accountEmail string) (*
...
+				// Log new AWS error
+				reqLogger.Error(aerr,
+					fmt.Sprintf(`New AWS Error during creation of IAM user (not ErrCodeNoSuchEntityException), 
+						AWS Error Code: %s, 
+						AWS Error Message: %s`,
+						aerr.Code(),
+						aerr.Message()))
+				return &iam.CreateUserOutput{}, err
+			}
+		} else {
```
rafael-azevedo: you don't need the else it can be omitted and  the behavior would be the same

ArielLima: +1 fixed

pkg/controller/account/account_controller.go
```diff
@@ -813,7 +867,7 @@ func formatAccountEmail(name string) string {
 }
 
 // TotalAwsAccounts returns the total number of aws accounts in the aws org
-func TotalAwsAccounts(client awsclient.Client) (int, error) {
+func TotalAwsAccounts(reqLogger logr.Logger, client awsclient.Client) (int, error) {
```
sam-nguyen7: this is being used in localmetrics line 101 and in account_controller_test line 114, which has not been updated with the reqLogger. Those 2 instances needs to incorporate in the reqLogger as well

PR #104, opened 2019-07-30
Fix for pending verification bug: SREP-1797

ALimaRedHat:
Previous logic for the createEC2Instance function was flawed. This is a revision of that code.
@rafael-azevedo @rogbas @jharrington22

### DIFF
--- a/pkg/controller/account/ec2.go
+++ b/pkg/controller/account/ec2.go
@@ -86,6 +86,17 @@ func (r *ReconcileAccount) BuildandDestroyEC2Instances(reqLogger logr.Logger, aw
 
 	instanceID, err := CreateEC2Instance(reqLogger, awsClient, ami)
 	if err != nil {
+		// Terminate instance id if it exists
+		if instanceID != "" {
+			// Log instance id of instance that will be terminated
+			reqLogger.Error(err, fmt.Sprintf(`Early termination of instance with ID: %s`, instanceID))
+			termErr := TerminateEC2Instance(reqLogger, awsClient, instanceID)
+			if termErr != nil {
+				if aerr, ok := termErr.(awserr.Error); ok {
+					reqLogger.Error(aerr, fmt.Sprintf(`AWS error while attempting to terminate instace, AWS Error Code: %s, AWS Error Message: %s`, aerr.Code(), aerr.Message()))
+				}
+			}
+		}
 		return err
 	}
 
@@ -129,10 +140,11 @@ func (r *ReconcileAccount) BuildandDestroyEC2Instances(reqLogger logr.Logger, aw
 
 // CreateEC2Instance creates ec2 instance and returns its instance ID
 func CreateEC2Instance(reqLogger logr.Logger, client awsclient.Client, ami string) (string, error) {
-	// Create EC2 service client
 
-	var instanceID string
-	var runErr error
+	// Retain instance id
+	var timeoutInstanceID string
+
+	// Loop until an EC2 instance is created or timeout.
 	attempt := 1
 	for i := 0; i < 300; i++ {
 		time.Sleep(time.Duration(attempt*5) * time.Second)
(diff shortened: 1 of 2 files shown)


pkg/controller/account/ec2.go
```diff
@@ -127,12 +127,10 @@ func (r *ReconcileAccount) BuildandDestroyEC2Instances(reqLogger logr.Logger, aw
 	return nil
 }
 
-// CreateEC2Instance creates ec2 instance and returns its instance ID
+//CreateEC2Instance creates ec2 instance and returns its instance ID.
```
jharrington22: Nit, can we add a space before comment starts and remove the period at the end

pkg/controller/account/ec2.go
```diff
@@ -142,25 +140,37 @@ func CreateEC2Instance(reqLogger logr.Logger, client awsclient.Client, ami strin
 		}
 		// Specify the details of the instance that you want to create.
 		runResult, runErr := client.RunInstances(&ec2.RunInstancesInput{
-			// An Amazon Linux AMI ID for t2.micro instances in the us-west-2 region
+			// An Amazon Linux AMI ID for t2.micro instances in a given region:
```
jharrington22: I think we can remove this comment since these are defined elsewhere

pkg/controller/account/ec2.go
```diff
@@ -142,25 +140,37 @@ func CreateEC2Instance(reqLogger logr.Logger, client awsclient.Client, ami strin
...
+							aerr.Code(),
+							aerr.Message()))
+					return "", runErr
+				}
+			}
+		}
 
-	if runErr != nil {
-		return "", runErr
+		// No error was found, save instance id and return
```
jharrington22: I think this maybe makes more sense? "No error was found, instance is running return instance Id"

pkg/controller/account/ec2.go
```diff
@@ -142,25 +140,37 @@ func CreateEC2Instance(reqLogger logr.Logger, client awsclient.Client, ami strin
...
+		return *runResult.Instances[0].InstanceId, nil
 	}
 
-	return instanceID, nil
-
+	// Timeout occurred
+	return "", ErrCreateEC2Instance
```
jharrington22: If we have an instance Id we should return it so we can clean it up

jharrington22:
@ALimaRedHat Could you modify the logic so that if we return a `ErrCreateEC2Instance` error from `CreateEC2Instance` we don't exit but continue provided we have a instance Id. This will make sure we terminate the instance if there is a timeout.

pkg/controller/account/ec2.go
```diff
@@ -142,25 +143,39 @@ func CreateEC2Instance(reqLogger logr.Logger, client awsclient.Client, ami strin
...
 			MaxCount:     aws.Int64(1),
 		})
-		if runErr == nil {
-			instanceID = *runResult.Instances[0].InstanceId
-			break
-		}
 
-	}
+		// Assign instance id
+		instanceID = *runResult.Instances[0].InstanceId
```
rafael-azevedo: This needs to be set after we check for errors.

pkg/controller/account/ec2.go
```diff
@@ -142,25 +143,39 @@ func CreateEC2Instance(reqLogger logr.Logger, client awsclient.Client, ami strin
...
+							aerr.Message()))
+					return "", runErr
+				}
+			}
+		}
 
-	if runErr != nil {
-		return "", runErr
+		// No error was found, instance is running, return instance id
+		return *runResult.Instances[0].InstanceId, nil
```
rafael-azevedo: we should return the instanceID variable we created instead of the pointer

pkg/controller/account/ec2.go
```diff
@@ -129,10 +129,11 @@ func (r *ReconcileAccount) BuildandDestroyEC2Instances(reqLogger logr.Logger, aw
 
 // CreateEC2Instance creates ec2 instance and returns its instance ID
 func CreateEC2Instance(reqLogger logr.Logger, client awsclient.Client, ami string) (string, error) {
-	// Create EC2 service client
 
+	// Retain instance id
 	var instanceID string
```
rafael-azevedo: This should be 

`instanceID  = ""`

that way if we never action it and return it it will not return a nil variable , the other option is to return an empty string instead of the the variable when we return an error.

pkg/controller/account/ec2.go
```diff
@@ -142,25 +143,39 @@ func CreateEC2Instance(reqLogger logr.Logger, client awsclient.Client, ami strin
...
-			instanceID = *runResult.Instances[0].InstanceId
-			break
-		}
 
-	}
+		// Assign instance id
+		instanceID = *runResult.Instances[0].InstanceId
+
+		// Return on unexpected errors:
+		if runErr != nil {
```
rafael-azevedo: I think somehow this error should be tracked and returned after the loop or we should have a counter that says the number of errors outside of the loop and if it is greater then 1 return the error.

pkg/controller/account/ec2.go
ALimaRedHat: The reason we do it here is so that in the case that we receive a timeout we have the instance ID to look at. If I put it after the tests we will never have it.

pkg/controller/account/ec2.go
ALimaRedHat: Like I said above the instance id variable is used for timeout errors

pkg/controller/account/ec2.go
ALimaRedHat: We already log all errors that are not expected:
```
reqLogger.Error(runErr,
  fmt.Sprintf(`Failed while trying to create EC2 instance, 
  AWS Error Code: %s, 
  AWS Error Message: %s`,
  aerr.Code(),
  aerr.Message()))
```

jharrington22:
@ALimaRedHat @rafael-azevedo  Ok changes are good however we do nothing with that `timeoutInstanceID` we need to not `err` here if an instance ID exists https://github.com/openshift/aws-account-operator/blob/56e2fd44787913bcbfc7269916867d1f4681ac88/pkg/controller/account/ec2.go#L88

rafael-azevedo:
> @ALimaRedHat @rafael-azevedo Ok changes are good however we do nothing with that `timeoutInstanceID` we need to not `err` here if an instance ID exists
> 
> https://github.com/openshift/aws-account-operator/blob/56e2fd44787913bcbfc7269916867d1f4681ac88/pkg/controller/account/ec2.go#L88


`timeoutInstanceID` is used line 178 ` return timeoutInstanceID, ErrCreateEC2Instance`

If we do not check if instance ID exists at line 88 line 119 will fail `err = TerminateEC2Instance(reqLogger, awsClient, instanceID)`

jharrington22:
@rafael-azevedo 

> > @ALimaRedHat @rafael-azevedo Ok changes are good however we do nothing with that `timeoutInstanceID` we need to not `err` here if an instance ID exists
> > https://github.com/openshift/aws-account-operator/blob/56e2fd44787913bcbfc7269916867d1f4681ac88/pkg/controller/account/ec2.go#L88
> 
> `timeoutInstanceID` is used line 178 ` return timeoutInstanceID, ErrCreateEC2Instance`
> 
> If we do not check if instance ID exists at line 88 line 119 will fail `err = TerminateEC2Instance(reqLogger, awsClient, instanceID)`

You are returning an error along with the instance ID so the instance ID will never get used because you err out.

PR #105, opened 2019-07-30
[WIP] Set policy for Federation Token request

fahlmant:
The `GetFederationTokenInput` struct looks like this:
```
type GetFederationTokenInput struct {
    _    struct{}    `type:"structure"`
    DurationSeconds    *int64    `min:"900" type:"integer"`
    Name    *string    `min:"2" type:"string" required:"true"`
    Policy    *string    `min:"1" type:"string"`
    PolicyArns    []*PolicyDescriptorType    `type:"list"`
}
```
This PR adds that policy to the struct during creation.

### DIFF
--- a/pkg/controller/account/credentials_rotator.go
+++ b/pkg/controller/account/credentials_rotator.go
@@ -51,7 +51,7 @@ func (r *ReconcileAccount) RotateCredentials(reqLogger logr.Logger, awsSetupClie
 
 	// Set IAM policy for Web Console login, this policy cannot grant more permissions than the IAM user has which creates it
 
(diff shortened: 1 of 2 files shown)


fahlmant:
@rafael-azevedo 
We are already logging the AWS Error message there:
`reqLogger.Error(err, fmt.Sprintf("Error: %s, %s", awsErr.Code(), awsErr.Message()))`
Also, that `return` is not releated to this PR.
:-1:

rafael-azevedo:
> @rafael-azevedo
> We are already logging the AWS Error message there:
> `reqLogger.Error(err, fmt.Sprintf("Error: %s, %s", awsErr.Code(), awsErr.Message()))`
> Also, that `return` is not releated to this PR.

That error message to me is to generic its just reprinting the error twice. A little message about where that is occurring will make troubleshooting easier.

fahlmant:
Added text.
In this particular case though, this wouldn't have helped as the segfaults happen before that if block is evven hit.

PR #107, opened 2019-08-02
SREP-1792: Error if sts token output is nil

fahlmant:
I have been unable to replicate this part segfaulting. Adding in some protection to make sure that a nil point is not accessed.

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -159,6 +159,9 @@ var ErrAwsSupportCaseIDNotFound = errors.New("SupportCaseIdNotfound")
 // ErrAwsFailedDescribeSupportCase indicates that the support case describe failed
 var ErrAwsFailedDescribeSupportCase = errors.New("FailedDescribeSupportCase")
 
+// ErrFederationTokenOutputNil indicates that getting a federation token from AWS failed
+var ErrFederationTokenOutputNil = errors.New("FederationTokenOutputNil")
+
 /**
 * USER ACTION REQUIRED: This is a scaffold file intended for the user to modify with their own Controller
 * business logic.  Delete these comments after modifying this file.*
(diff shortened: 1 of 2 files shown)


jewzaam:
Is this an intermittent issue or something that's failing in a specific scenario?  It sounds like this will keep the operator from crashing.  But it may mask issues with processing the STS credentials.  Do we have alerts on these tokens being stale or not updated?

fahlmant:
@jewzaam I haven't been able to recreate the issue, seem intermittent on stage. 

We have a credentials rotator to make sure these are fresh. I'm not sure an alert is the right approach for that, but that's out of scope of what this PR is trying to accomplish.

rafael-azevedo:
> Is this an intermittent issue or something that's failing in a specific scenario? It sounds like this will keep the operator from crashing. But it may mask issues with processing the STS credentials. Do we have alerts on these tokens being stale or not updated?

The issue processing the STS token is most likely api flakiness. By breaking here in the future refactor we should be able to start from this step again should it break here instead of fail the account.

pkg/controller/account/account_controller.go
```diff
@@ -129,6 +129,9 @@ var ErrAwsSupportCaseIDNotFound = errors.New("SupportCaseIdNotfound")
 // ErrAwsFailedDescribeSupportCase indicates that the support case describe failed
 var ErrAwsFailedDescribeSupportCase = errors.New("FailedDescribeSupportCase")
 
+//ErrFederationTokenOutputNil indicates that getting a federation token from AWS failed
```
jharrington22: Nit pick but could you add a space before the text?

jharrington22:
> > Is this an intermittent issue or something that's failing in a specific scenario? It sounds like this will keep the operator from crashing. But it may mask issues with processing the STS credentials. Do we have alerts on these tokens being stale or not updated?
> 
> The issue processing the STS token is most likely api flakiness. By breaking here in the future refactor we should be able to start from this step again should it break here instead of fail the account.

Since we are returning an error now the operator will retry this step since its before we set `status.rotateCredentials` to `false`

PR #114, opened 2019-08-08
Reduce STS credential duration until we update roles

jharrington22:
We are unable to increase the `STSCredentialDuration` to over an hour until we update the roles maximum session duration. This change will require a larger amount of work, this gets the credentials working.


jharrington22:
@mwoodson @jewzaam CR please?

PR #115, opened 2019-08-12
SREP-1883 aws-account-operator increase support case check interval

ALimaRedHat:
- Added a log before requeuing a support case check. 
- Increase interval between each check on a support case with the "PendingVerification" status to 10 minutes.
https://jira.coreos.com/browse/SREP-1883

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -54,7 +54,7 @@ const (
 	caseDesiredInstanceLimit      = 25
 	caseStatusResolved            = "resolved"
 	intervalAfterCaseCreationSecs = 30
-	intervalBetweenChecksSecs     = 30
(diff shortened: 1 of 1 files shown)


pkg/controller/account/account_controller.go
```diff
@@ -54,7 +54,7 @@ const (
 	caseDesiredInstanceLimit      = 25
 	caseStatusResolved            = "resolved"
 	intervalAfterCaseCreationSecs = 30
-	intervalBetweenChecksSecs     = 30
+	intervalBetweenChecksSecs     = 10
```
rogbas: The variable name has Secs, yet we use it as minutes. Can you rename it please?

ALimaRedHat: good catch

PR #117, opened 2019-08-13
SREP-1918

ALimaRedHat:
remove kubeclient dependency when running NewClient


ALimaRedHat:
`grep -rl "NewClient" .` output:
```
./vendor/...
./pkg/controller/account/account_controller.go
./pkg/awsclient/client.go
```

ALimaRedHat:
tested and working @jharrington22

PR #119, opened 2019-08-16
Modified Reuse metric to include ID

ninataneja:
Added legalID to as a label for the metric for available reused accounts. Instead of the total number of reused accounts available, the metric will now group the available accounts by their ID.

### DIFF
--- a/pkg/localmetrics/localmetrics.go
+++ b/pkg/localmetrics/localmetrics.go
@@ -97,7 +97,15 @@ func UpdateAccountCRMetrics(accountList *awsv1alpha1.AccountList) {
 	reuseAccountFailedCount := 0
 	pendingVerificationAccountCount := 0
 	readyAccountCount := 0
+	idMap := make(map[string]int)
 	for _, account := range accountList.Items {
+		if idMap[account.Spec.LegalEntity.ID] == 0 {
+			idMap[account.Spec.LegalEntity.ID] = 1
+		} else {
+			iDCount := idMap[account.Spec.LegalEntity.ID]
+			iDCount++
+		}
+
 		if account.Status.Claimed == false {
 			// Ignore unclaimed accounts in Failed status
 			if account.Status.State != "Failed" {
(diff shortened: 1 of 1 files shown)


pkg/localmetrics/localmetrics.go
```diff
@@ -97,7 +97,15 @@ func UpdateAccountCRMetrics(accountList *awsv1alpha1.AccountList) {
 	reuseAccountFailedCount := 0
 	pendingVerificationAccountCount := 0
 	readyAccountCount := 0
+	idMap := make(map[string]int)
 	for _, account := range accountList.Items {
+		if idMap[account.Spec.LegalEntity.ID] == 0 {
```
fahlmant: Not sure how Go does this, but I think you need to check if the key doesn't exist and create it.

pkg/localmetrics/localmetrics.go
```diff
@@ -97,7 +97,15 @@ func UpdateAccountCRMetrics(accountList *awsv1alpha1.AccountList) {
...
+			idMap[account.Spec.LegalEntity.ID] = 1
+		} else {
+			iDCount := idMap[account.Spec.LegalEntity.ID]
+			iDCount++
```
fahlmant: Could this be done in one line?

pkg/localmetrics/localmetrics.go
ninataneja: I think that the value is equal to 0 by default if the key doesn't exist so that is why i did if the val == 0, and then i think you can just set the key = to the value like in this example? https://blog.golang.org/go-maps-in-action

fahlmant: Thanks for the docs, this looks right then.

pkg/localmetrics/localmetrics.go
```diff
@@ -136,8 +143,11 @@ func UpdateAccountCRMetrics(accountList *awsv1alpha1.AccountList) {
 	MetricTotalAccountPendingVerification.With(prometheus.Labels{"name": "aws-account-operator"}).Set(float64(pendingVerificationAccountCount))
 	MetricTotalAccountCRsFailed.With(prometheus.Labels{"name": "aws-account-operator"}).Set(float64(failedAccountCount))
 	MetricTotalAccountCRsReady.With(prometheus.Labels{"name": "aws-account-operator"}).Set(float64(readyAccountCount))
-	MetricTotalAccountReusedAvailable.With(prometheus.Labels{"name": "aws-account-operator"}).Set(float64(reusedAccountAvailableCount))
 	MetricTotalAccountReuseFailed.With(prometheus.Labels{"name": "aws-account-operator"}).Set(float64(reuseAccountFailedCount))
+	for _, account := range accountList.Items {
```
rogbas: Shouldn't we loop thru the idMap slice here? What we're trying to accomplish is having a metric with labels like that below, isn't it?
```json
{
'LegalID-12334546': 12,
'LegalID-XYZZZWX': 7,
}

fahlmant: @ninataneja I missed this, but yes this should loop through the dictionary you made.

ninataneja:
@fahlmant

PR #122, opened 2019-08-19
Revert "Modified Reuse metric to include ID"

jharrington22:
Reverts openshift/aws-account-operator#119 - This PR is causing the operator to crash, reverting until we have tested fix.


fahlmant:
@jharrington22 Please close this in favor of #124

PR #124, opened 2019-08-19
Fixed label issue for Reuse metric 

ninataneja:
Put correct label for LegalID in metric definition and moved name label to a constant label


ninataneja:
@fahlmant @jharrington22 would it be better to roll forward with this instead of merging the revert #122 ? i think this is the fix for the label

fahlmant:
Closing in favor of #125

PR #125, opened 2019-08-19
New fix for adding LegalID to Reuse Metric

ninataneja:
Added legalID to reuse metric so that the number of available reused accounts are grouped by ID. Also fixed label issues that were causing problems in other PR. There is now a label for the LegalID in the metric itself, and the name label is just a const label in the metric definition.


jharrington22:
@fahlmant please review

PR #128, opened 2019-08-20
Added mocks and sample test

rafael-azevedo:
This PR adds mocks for the awsClient using mockgen from the gomock package. It also used these mocks to test the TotalAwsAccounts() function as an example of how to use this mock structure to test. 

We will need to vendor the following 
```
github.com/golang/mock/
github.com/stretchr/testify/assert
github.com/evanphx/json-patch
```

For testing locally we need to install the mockgen library
`go install github.com/golang/mock/mockgen`

The mock file will need to be updated every time we modify the interfaces used to generate it.
For example if we modify awsclient.go we will need to run the following 
`mockgen --source=pkg/awsclient/client.go  --destination=pkg/awsclient/mock/mock_client.go --package=mock`


mwoodson:
@rafael-azevedo I like this!  Great start, and clean implementation here.

I'd like to get other's opinions on this as well, as I feel like this is a good way to implement unit tests.

pkg/controller/account/account_controller_test.go
```diff
@@ -0,0 +1,141 @@
...
+			},
+			validateTotal: func(t *testing.T, total int) {
+				assert.Equal(t, 2, total)
+			}},
+		{
+			name:          "AccountList with NextToken, return 4 accounts",
+			errorExpected: false,
+			setupAWSMock: func(r *mockAWS.MockClientMockRecorder) {
+				gomock.InOrder(
+					r.ListAccounts(gomock.Any()).Return(
```
fahlmant: Why have two account lists here?

fahlmant: Nvm, you're listing AWS accounts, not CRs

PR #129, opened 2019-08-20
Small capitalization fix

fahlmant:
Not sure if we need to or should regen CRDs, maybe in another PR.


rogbas:
Just checked the auto generated functions, they're good, don't need to be updated.

PR #130, opened 2019-08-20
Regenerate CRDs to match current definitions

fahlmant:
Using operator-sdk version v0.8.2

Regenerating openapi and k8s via the cli to match what's currently in the definitions.


jharrington22:
@rogbas @rafael-azevedo can you think of any reasons the new required fields could cause issues? I don't think so since OCM will pass these along. What about existing CR's?

PR #137, opened 2019-08-22
Created Go Routine to update total AWS Accounts every N hours

ninataneja:
Added Go routine in main.go to update the total number of AWS account every hour. Moved logic of updating this number to local metrics file so that total number is calculated in local metrics. The API call still needs to be tested however

### DIFF
--- a/pkg/localmetrics/localmetrics.go
+++ b/pkg/localmetrics/localmetrics.go
@@ -16,9 +16,13 @@ package localmetrics
 
 import (
 	"math"
+	"time"
 
 	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
+	"github.com/openshift/aws-account-operator/pkg/awsclient"
+	"github.com/openshift/aws-account-operator/pkg/controller/account"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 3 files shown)


pkg/localmetrics/localmetrics.go
```diff
@@ -83,9 +87,19 @@ var (
...
+func UpdateAWSMetrics(kubeClient kubeclientpkg.Client, hour int) {
+	awsClient, _ := awsclient.GetAWSClient(kubeClient, awsclient.NewAwsClientInput{
+		SecretName: account.AwsSecretName,
+		NameSpace:  awsv1alpha1.AccountCrNamespace,
+		AwsRegion:  "us-east-1",
+	})
+
+	d := time.Duration(hour) * time.Minute
+	for range time.Tick(d) {
+		accountTotal, _ := account.TotalAwsAccounts(awsClient)
```
fahlmant: The err should be caught and logged if not nil. This means the metric will only be updated if there is no err.

jharrington22:
@fahlmant @rogbas @rafael-azevedo I have created a bug for this and since @ninataneja is unlikely to update this branch at the moment (all good Nina :slightly_smiling_face:) lets merge and I'll have the bug fixed asap.

jharrington22:
Bug created to fix this issue SREP-2085

PR #138, opened 2019-08-29
SREP-1979 Implement unit tests for accountpool controller

fahlmant:
Create tests for helper functions and reconcile loop in the accountpool controller.

### DIFF
--- a/pkg/controller/accountpool/accountpool_controller_test.go
+++ b/pkg/controller/accountpool/accountpool_controller_test.go
@@ -1,0 +1,351 @@
+package accountpool
+
+import (
+	"github.com/golang/mock/gomock"
+	awsaccountapis "github.com/openshift/aws-account-operator/pkg/apis"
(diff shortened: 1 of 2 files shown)


pkg/controller/accountpool/accountpool_controller_test.go
```diff
@@ -0,0 +1,307 @@
...
+	return mocks
+}
+
+func TestReconcileAccountPool(t *testing.T) {
+	awsaccountapis.AddToScheme(scheme.Scheme)
+	tests := []struct {
+		name                  string
+		localObjects          []runtime.Object
+		expectedAccountPool   awsv1alpha1.AccountPool
+		verifyAccountFunction func(client.Client, *awsv1alpha1.AccountPool) bool
```
rafael-azevedo: you don't need to define this here unless you are defining a different function in line for each test. Since you are using the same function for all the tests and its defined outside of this block you can just call it like you are below.

fahlmant: I did define a different function for each test

PR #140, opened 2019-09-03
SREP-2029 Increase account limit to 4550

rogbas:


pkg/controller/account/account_controller.go
```diff
@@ -36,7 +36,7 @@ import (
 var log = logf.Log.WithName("controller_account")
 
 const (
-	awsLimit                = 4450
+	awsLimit                = 4550
```
lisa: How was 4550 reached? (How do we know to increase it?)

PR #141, opened 2019-09-03
Refactor unclaimed account metric

fahlmant:
Create a separate function for the unclaimed metric to ensure accurate reporting.

### DIFF
--- a/pkg/localmetrics/localmetrics.go
+++ b/pkg/localmetrics/localmetrics.go
@@ -17,6 +17,7 @@ package localmetrics
 import (
 	"fmt"
 	"math"
+	"sync"
 	"time"
 
 	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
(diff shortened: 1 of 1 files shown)


pkg/localmetrics/localmetrics.go
```diff
@@ -88,9 +88,28 @@ func UpdateAWSMetrics(totalAccounts int) {
...
 }
 
+// UpdateAccountCRUnclaimedMetric updates the unclaimed account metric
+func UpdateAccountCRUnclaimedMetric(accountList awsv1alpha1.AccountList) {
+
+	unclaimedAccountCount := 0
+
+	for _, account := range accountList.Items {
+		if account.Status.Claimed == false && account.Status.Reused == false {
+			if account.Status.State == "Ready" || account.Status.State == "PendingVerifiation" {
```
rogbas: Typo here in PendingVerifiation

pkg/localmetrics/localmetrics.go
```diff
@@ -88,9 +88,28 @@ func UpdateAWSMetrics(totalAccounts int) {
...
+
+	MetricTotalAccountCRsUnclaimed.With(prometheus.Labels{"name": "aws-account-operator"}).Set(float64(unclaimedAccountCount))
+
+}
+
 // UpdateAccountCRMetrics updates all metrics related to Account CRs
 func UpdateAccountCRMetrics(accountList *awsv1alpha1.AccountList) {
-	unclaimedAccountCount := 0
+
+	go UpdateAccountCRUnclaimedMetric(*accountList)
```
rogbas: Since this is a go routine, do we need to think on using any kind of sync/WaitGroup approach?

pkg/localmetrics/localmetrics.go
jharrington22: Don't we have a constant for that?

pkg/localmetrics/localmetrics.go
jharrington22: I believe we'll need to sync because is it possible these could run simultaneously? We would get an API error if it did

pkg/localmetrics/localmetrics.go
fahlmant: Ugh and I thought double checked it :(

pkg/localmetrics/localmetrics.go
fahlmant: Since this is just taking a copy of a struct, there's no danger in leaving it as is

rafael-azevedo: a waitgroup might be nice to limit the number of concurrent threads

rogbas: Is it possible to have multiple calls of that go routine running at the same time? What if we call `MetricTotalAccountCRsUnclaimed.With(prometheus.Labels{"name": "aws-account-operator"}).Set(float64(unclaimedAccountCount))` simultaneously ?

jharrington22: Just a thought here, if we are not waiting for the the go routine to finish what is the benefit of having it in a go routine? @rogbas @fahlmant

fahlmant: We are waiting for it to return.

PR #142, opened 2019-09-03
Added logic to delete bucket objects

rafael-azevedo:
This adds functionality to delete objects in a bucket if the bucket is not empty. Still requires testing.

### DIFF
--- a/pkg/controller/accountclaim/reuse.go
+++ b/pkg/controller/accountclaim/reuse.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/go-logr/logr"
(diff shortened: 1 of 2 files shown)


pkg/controller/accountclaim/reuse.go
```diff
@@ -223,12 +224,37 @@ func (r *ReconcileAccountClaim) cleanUpAwsAccountS3(reqLogger logr.Logger, awsCl
...
 		if err != nil {
-			delError := fmt.Sprintf("Failed deleting S3 bucket: %s", *bucket.Name)
-			reqLogger.Error(err, delError)
-			awsErrors <- delError
-			return err
+			if aerr, ok := err.(awserr.Error); ok {
+				DelObjErr := fmt.Sprintf("Failed deleting items in S3 bucket: %s", *bucket.Name)
+				switch aerr.Code() {
+				// if bucket is not empty empty the bucket and attempt to delete again
+				case "BucketNotEmpty":
```
jharrington22: Do you think this logic should be part of the error? I think the correct logic would be to check if the bucket is empty or not and then perform the appropriate action.

rafael-azevedo: I had that thought @jharrington22  but so far we only hit this issue with the velero buckets and then we delete it. We can check if the bucket is empty first , but I see more logic needed and logic that may not necessarily need to be run every time, but I am happy to change it if you guys want, but it would probably have to be tomorrow at that point.

rogbas:
PR tested locally
/lgtm

PR #149, opened 2019-09-09
Explicitly support * on accountpools

fahlmant:
I don't think this is required, but being explicit with RBAC is probably a safer bet.

### DIFF
--- a/deploy/cluster_role.yaml
+++ b/deploy/cluster_role.yaml
@@ -44,6 +44,7 @@ rules:
   - '*'
   - accountclaims
   - accounts
+  - accountpools
   verbs:
   - '*'
 - apiGroups:


jharrington22:
@fahlmant Is there a reason you added this? Its not required right now because the operator doesn't need to modify account pool CR's and unless we do its probably a safer security wise to leave it out, would you agree?

fahlmant:
@jharrington22 I think this line already gives that access:
https://github.com/openshift/aws-account-operator/pull/149/files#diff-991bb55ef23f18e7ffa537df2221d7bcR44

I've also been running into a strange issue with listing accountpools while testing the operator. I thought this was the issue, but now I'm sure its not. But still I thought it might be to explicit call out the perms we're granting. If line 44 doesn't allow for accountpools then sure, close the PR

jharrington22:
Ah I see ok, i'll merge since its allowed anyway.

PR #152, opened 2019-09-10
Add permissive option to image registry initializer

jharrington22:
This PR adds the `--permissive` flag to the image registry docker build to ignore errors when loading and initializing the OLM bundles restoring the behavior the image registry initializer had before this PR https://github.com/operator-framework/operator-registry/pull/74

This option should be removed once https://github.com/operator-framework/operator-registry/pull/76 is merged which fixes the initial issue where the initalizer attempts to parse any file as JSON or YAML.

This should be considered temporary to fix our Jenkins builds until the above is complete.


rafael-azevedo:
should we merge this as well https://github.com/openshift/aws-account-operator/pull/133

PR #155, opened 2019-09-11
SREP-1964 Check status error when updating account status

sam-nguyen7:
Found one instance in setStatusFailed where status error was not being checked when updating account status


rafael-azevedo:
This PR seems to cover all the situations where we were not checking the errors when updating status. We should be good to merge

PR #157, opened 2019-09-14
Wait for all routines to finish, don't log errors twice

jharrington22:
This PR fixes an issue were we exit a function closing message channels while go routines using those channels are still running. I've also removed unnecessary logging from those functions since we'll print any errors that arrive on the channel.

### DIFF
--- a/pkg/controller/accountclaim/reuse.go
+++ b/pkg/controller/accountclaim/reuse.go
@@ -82,6 +82,9 @@ func (r *ReconcileAccountClaim) resetAccountSpecStatus(reqLogger logr.Logger, re
 }
 
 func (r *ReconcileAccountClaim) cleanUpAwsAccount(reqLogger logr.Logger, accountClaim *awsv1alpha1.AccountClaim, account *awsv1alpha1.Account) error {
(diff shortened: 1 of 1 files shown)


pkg/controller/accountclaim/reuse.go
```diff
@@ -125,7 +125,6 @@ func (r *ReconcileAccountClaim) cleanUpAwsAccount(reqLogger logr.Logger, account
 		case errMsg := <-awsErrors:
 			err = errors.New(errMsg)
 			reqLogger.Error(err, errMsg)
-			return err
```
rogbas: I don't think we can simply remove this return. We still need to inform the caller function that we had an error during the cleanup. That error will trigger the account being flagged as Failed Reuse.

Maybe we could store the error messages and a flag error = true, and after all the 3 functions finish we return it?

jharrington22: Ah I see yes we can just keep a bool indicating there was an error and return once the routines are complete.

rogbas: /lgtm

PR #158, opened 2019-09-16
Handle and log errors in localmetrics

liamwazherealso:
The aws-account-operator total number of AWS account metric is showing 0 occasionally, I suspect this is due to the API returning the "TooManyRequestsException" error (which we've experienced before) and then returning 0

Maybe we want to parse the returning API error?

https://github.com/openshift/aws-account-operator/blob/master/pkg/controller/account/account_controller.go#L820

This only merged in staging, we want to resolve this before our next prod push

https://jira.coreos.com/browse/SREP-2108

Along with this, I updated the other error handing case in metrics.

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -817,7 +817,11 @@ func TotalAwsAccounts(client awsclient.Client) (int, error) {
 	for {
 		awsAccountList, err := client.ListAccounts(&organizations.ListAccountsInput{NextToken: nextToken})
 		if err != nil {
-			return 0, errors.New("Error getting a list of accounts")
+			errMsg := "Error getting a list of accounts"
+			if aerr, ok := err.(awserr.Error); ok {
+				errMsg = fmt.Sprintf("Failed to get account list with error code %s", aerr.Message())
+			}
+			return 0, errors.New(errMsg)
 		}
 		awsAccounts = append(awsAccounts, awsAccountList.Accounts...)
 		if awsAccountList.NextToken != nil {
--- a/pkg/localmetrics/localmetrics.go
+++ b/pkg/localmetrics/localmetrics.go
@@ -15,6 +15,7 @@
 package localmetrics
 
 import (
+	"fmt"
 	"math"
 	"time"
 
@@ -23,7 +24,10 @@ import (
 	"github.com/openshift/aws-account-operator/pkg/controller/account"
 	"github.com/prometheus/client_golang/prometheus"
 	kubeclientpkg "sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 )
+
+var log = logf.Log.WithName("localmetrics")
 
 var (
 	MetricTotalAWSAccounts = prometheus.NewGaugeVec(prometheus.GaugeOpts{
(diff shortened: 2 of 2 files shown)


pkg/localmetrics/localmetrics.go
```diff
@@ -89,16 +92,29 @@ var (
 
 // UpdateAWSMetrics updates the total AWS Accounts metric every N hours
 func UpdateAWSMetrics(kubeClient kubeclientpkg.Client, hour int) {
-	awsClient, _ := awsclient.GetAWSClient(kubeClient, awsclient.NewAwsClientInput{
+	metricLogger := log.WithValues("Namespace", "aws-account-operator-operator")
+	metricLogger.Info("Metrics for AWS Operator")
```
fahlmant: This is only a single metric being updated, not all metrics for AWS Operator. So I'd recommend moving this somewhere else

jharrington22: I don't think we need to log at all every time we update metrics, would you agree @fahlmant ?

pkg/localmetrics/localmetrics.go
```diff
@@ -89,16 +92,29 @@ var (
...
+	}
+
 	d := time.Duration(hour) * time.Minute
 	for range time.Tick(d) {
-		accountTotal, _ := account.TotalAwsAccounts(awsClient)
-		MetricTotalAWSAccounts.With(prometheus.Labels{"name": "aws-account-operator"}).Set(float64(accountTotal))
+		accountTotal, err := account.TotalAwsAccounts(awsClient)
+
+		if err != nil {
+			metricLogger.Error(err, "Failed to get account count, not updating metric.")
```
jharrington22: What would be better here is if `TotalAwsAccounts() returned the AWS error code and we printed it here eg `fmt.Sprintf(err, "Failed to get total number of AWS accounts: %s", err)`

pkg/localmetrics/localmetrics.go
```diff
@@ -89,16 +93,28 @@ var (
...
+	}
+
 	d := time.Duration(hour) * time.Minute
 	for range time.Tick(d) {
-		accountTotal, _ := account.TotalAwsAccounts(awsClient)
-		MetricTotalAWSAccounts.With(prometheus.Labels{"name": "aws-account-operator"}).Set(float64(accountTotal))
+		accountTotal, err := account.TotalAwsAccounts(awsClient)
+
+		if err != nil {
+			metricLogger.Error(err, fmt.Sprintf("AAO Failed to get total number of AWS accounts: %s", err))
```
jharrington22: Remove "AAO" we don't use that and since this error would only been seen in the aws-account-operator error logs its obvious where its coming from :slightly_smiling_face:

pkg/controller/account/account_controller.go
```diff
@@ -817,7 +817,11 @@ func TotalAwsAccounts(client awsclient.Client) (int, error) {
 	for {
 		awsAccountList, err := client.ListAccounts(&organizations.ListAccountsInput{NextToken: nextToken})
 		if err != nil {
-			return 0, errors.New("Error getting a list of accounts")
+			errMsg := "Error getting a list of accounts"
+			if aerr, ok := err.(awserr.Error); ok {
+				errMsg = fmt.Sprintf("Failed to get account list with error code %i", aerr.Code())
```
jharrington22: `%i` is not a verb `fmt` supports. Could you change this to `%s` and `aerr.Message()`. Apologies I did say code but I meant message.

pkg/localmetrics/localmetrics.go
fahlmant: Yes, I really think this should be removed overall

pkg/controller/account/account_controller.go
liamwazherealso: ack

pkg/localmetrics/localmetrics.go
liamwazherealso: ack

pkg/localmetrics/localmetrics.go
```diff
@@ -89,16 +93,28 @@ var (
...
 		NameSpace:  awsv1alpha1.AccountCrNamespace,
 		AwsRegion:  "us-east-1",
 	})
 
+	if err != nil {
+		metricLogger.Error(err, "Failed to get awsClient")
+		return
+	}
+
 	d := time.Duration(hour) * time.Minute
```
fahlmant: Please change this to the correct time unit.

pkg/localmetrics/localmetrics.go
```diff
@@ -89,16 +93,28 @@ var (
...
 	for range time.Tick(d) {
-		accountTotal, _ := account.TotalAwsAccounts(awsClient)
-		MetricTotalAWSAccounts.With(prometheus.Labels{"name": "aws-account-operator"}).Set(float64(accountTotal))
+		accountTotal, err := account.TotalAwsAccounts(awsClient)
+
+		if err != nil {
+			metricLogger.Error(err, fmt.Sprintf("Failed to get total number of AWS accounts: %s", err))
```
fahlmant: Logging here is good, but one has to actually look at the logs to know this is failing. Maybe we set the metric here to `-1` and alert on it if its `-1` for some number of hours. Thoughts?

jharrington22: Thats the problem we have right now though right it gets set to 0 and our graphs get screwed?

fahlmant: Yeah I guess this would screw up history. Ignore this then

pkg/localmetrics/localmetrics.go
liamwazherealso: Done

pkg/localmetrics/localmetrics.go
liamwazherealso: Done

liamwazherealso:
Tested

PR #159, opened 2019-09-17
Explicitly call out namespace in clusterrolebinding

fahlmant:
Since the operator and therefore the service account is always in the `aws-account-operator` namespace by convention, I want to make this explicit as its the only bit that needs replacing when deploying the operator.


jharrington22:
@fahlmant makes sense to me.

PR #164, opened 2019-09-23
add permissions to uhc-cluster-role

igoihman:
As part of OCM migration to `controller-runtime`, we need to have permissions to create events and create and get configmaps. This is done only in a dedicated namespaces called `uhc-leadership`.

### DIFF
--- a/deploy/uhc_cluster_role.yaml
+++ b/deploy/uhc_cluster_role.yaml
@@ -13,3 +13,16 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - events
+  verbs:
+  - create
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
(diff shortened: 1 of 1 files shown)


igoihman:
@jhernand @jharrington22 PTAL

jhernand:
Looks good to me.

jharrington22:
@jhernand @igoihman I'm not sure why we need to update the role in the aws-account-operator? The service account that this role is bound too was specifically for OCM to be able to list accountclaims. Should you be using a different service account to access these resources? Or maybe i'm misunderstanding what is happening here :slightly_smiling_face:

igoihman:
@jharrington22 OCM is using this service account in order to register to accountclaim events.
As we decided to migrate to `controller-runtime` we noticed it has a mechanism of creating a configmap and events in a specific namespace that we define. If we won't have the permissions I added we won't be able to get accountclaim events.

PR #167, opened 2019-09-26
added update configmaps permission to UHC cluster role

igoihman:
added missing RBAC permission


igoihman:
@jhernand @jharrington22 PTAL

PR #171, opened 2019-09-27
Move check for awslimit outside of reconcile.

liamwazherealso:
https://jira.coreos.com/browse/SREP-2163
Currently the aws-account-operator will check the total number of created accounts when an account CR has `.status.state` as "" and `.status.claimed` as false.

We should do this independently of the reconcile loop and run it on a timer since we consistently hit API errors from AWS blocking the controller as well as blocking the refreshing of already created accounts. We would do this similar to the way we're updating the localmetics in main.go.

The problem is that we need to check this against the awsLimit var so that we don't create too many accounts. Should we create an internal counter too that is refreshed on every update from AWS?

Done criteria:
- Checking of total number of AWS accounts is done outside the loop
- We need to observe awsLimit until we implement some rate limiting when creating accounts

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -13,12 +13,14 @@ import (
 	"github.com/openshift/aws-account-operator/pkg/controller"
 	"github.com/openshift/aws-account-operator/pkg/credentialwatcher"
 	"github.com/openshift/aws-account-operator/pkg/localmetrics"
+	"github.com/openshift/aws-account-operator/pkg/totalaccountwatcher"
 	"github.com/openshift/operator-custom-metrics/pkg/metrics"
 	"github.com/operator-framework/operator-sdk/pkg/leader"
 	"github.com/operator-framework/operator-sdk/pkg/log/zap"
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
 	"github.com/spf13/pflag"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
@@ -31,6 +33,7 @@ var (
 	metricsPath                   = "/metrics"
 	secretWatcherScanInterval     = time.Duration(10) * time.Minute
 	hours                     int = 1
+	totalWatcherInterval          = time.Duration(15) * time.Minute
 )
 
 var log = logf.Log.WithName("cmd")
@@ -126,13 +129,20 @@ func main() {
 	// work
 	stopCh := signals.SetupSignalHandler()
 
+	accountWatcherClient, err := client.New(cfg, client.Options{})
+	if err != nil {
+		log.Error(err, "")
+	}
+
 	// Initialize the SecretWatcher
 	credentialwatcher.Initialize(mgr.GetClient(), secretWatcherScanInterval)
+
+	totalaccountwatcher.Initialize(accountWatcherClient, totalWatcherInterval)
 
 	// Start the secret watcher
 	go credentialwatcher.SecretWatcher.Start(log, stopCh)
 
-	go localmetrics.UpdateAWSMetrics(mgr.GetClient(), hours)
+	go totalaccountwatcher.TotalAccountWatcher.Start(log, stopCh)
 
 	log.Info("Starting the Cmd.")
 
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -18,6 +18,8 @@ import (
 	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
 	"github.com/openshift/aws-account-operator/pkg/awsclient"
 	controllerutils "github.com/openshift/aws-account-operator/pkg/controller/utils"
+	totalaccountwatcher "github.com/openshift/aws-account-operator/pkg/totalaccountwatcher"
+
 	corev1 "k8s.io/api/core/v1"
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -35,13 +37,12 @@ import (
 var log = logf.Log.WithName("controller_account")
 
 const (
-	awsLimit                = 4700
+	AwsLimit                = 4700
 	awsCredsUserName        = "aws_user_name"
 	awsCredsSecretIDKey     = "aws_access_key_id"
 	awsCredsSecretAccessKey = "aws_secret_access_key"
 	iamUserNameUHC          = "osdManagedAdmin"
 	iamUserNameSRE          = "osdManagedAdminSRE"
-	AwsSecretName           = "aws-account-operator-credentials"
 	awsAMI                  = "ami-000db10762d0c4c05"
 	awsInstanceType         = "t2.micro"
 	createPendTime          = 10 * time.Minute
@@ -243,7 +244,7 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 
 	// We expect this secret to exist in the same namespace Account CR's are created
 	awsSetupClient, err := awsclient.GetAWSClient(r.Client, awsclient.NewAwsClientInput{
-		SecretName: AwsSecretName,
+		SecretName: controllerutils.AwsSecretName,
 		NameSpace:  awsv1alpha1.AccountCrNamespace,
 		AwsRegion:  "us-east-1",
 	})
@@ -333,27 +334,23 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 	}
 
 	if (currentAcctInstance.Status.State == "") && (currentAcctInstance.Status.Claimed == false) {
-		// before doing anything make sure we are not over the limit if we are just error
-		accountTotal, err := TotalAwsAccounts(awsSetupClient)
-		if err != nil {
-			reqLogger.Info("Failed to get AWS account total from AWS api", "Error", err.Error())
-			return reconcile.Result{}, err
-		}
-
-		if accountTotal >= awsLimit {
-			reqLogger.Error(ErrAwsAccountLimitExceeded, "AWS Account limit reached", "Account Total", accountTotal)
-			return reconcile.Result{}, ErrAwsAccountLimitExceeded
-		}
 
 		var awsAccountID string
 
 		if currentAcctInstance.Spec.AwsAccountID == "" {
+			// before doing anything make sure we are not over the limit if we are just error
+			if totalaccountwatcher.TotalAccountWatcher.Total >= AwsLimit {
+				reqLogger.Error(ErrAwsAccountLimitExceeded, "AWS Account limit reached", "Account Total", totalaccountwatcher.TotalAccountWatcher.Total)
+				return reconcile.Result{}, ErrAwsAccountLimitExceeded
+			}
 
 			// Build Aws Account
 			awsAccountID, err = r.BuildAccount(reqLogger, awsSetupClient, currentAcctInstance)
 			if err != nil {
 				return reconcile.Result{}, err
 			}
+
+			totalaccountwatcher.TotalAccountWatcher.Total += 1
 
 			// set state creating if the account was able to create
 			SetAccountStatus(reqLogger, currentAcctInstance, "Attempting to create account", awsv1alpha1.AccountCreating, "Creating")
@@ -373,7 +370,7 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 
 			awsAccountID = currentAcctInstance.Spec.AwsAccountID
 
-			// set state creating if the account was alredy created
+			// set state creating if the account was already created
 			SetAccountStatus(reqLogger, currentAcctInstance, "AWS account already created", awsv1alpha1.AccountCreating, "Creating")
 			err = r.Client.Status().Update(context.TODO(), currentAcctInstance)
 			if err != nil {
@@ -811,33 +808,6 @@ func formatAccountEmail(name string) string {
 
 	email := prefix + "+" + splitString[len(splitString)-1] + "@redhat.com"
 	return email
-}
-
-// TotalAwsAccounts returns the total number of aws accounts in the aws org
-func TotalAwsAccounts(client awsclient.Client) (int, error) {
-	var awsAccounts []*organizations.Account
-
-	var nextToken *string
-
-	// Ensure we paginate through the account list
-	for {
-		awsAccountList, err := client.ListAccounts(&organizations.ListAccountsInput{NextToken: nextToken})
-		if err != nil {
-			errMsg := "Error getting a list of accounts"
-			if aerr, ok := err.(awserr.Error); ok {
-				errMsg = fmt.Sprintf("Failed to get account list with error code %s", aerr.Message())
-			}
-			return 0, errors.New(errMsg)
-		}
-		awsAccounts = append(awsAccounts, awsAccountList.Accounts...)
-		if awsAccountList.NextToken != nil {
-			nextToken = awsAccountList.NextToken
-		} else {
-			break
-		}
-	}
-
-	return len(awsAccounts), nil
 }
 
 // SetAccountStatus sets the status of an account
--- a/pkg/localmetrics/localmetrics.go
+++ b/pkg/localmetrics/localmetrics.go
@@ -15,16 +15,11 @@
 package localmetrics
 
 import (
-	"fmt"
 	"math"
 	"sync"
-	"time"
 
 	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
-	"github.com/openshift/aws-account-operator/pkg/awsclient"
-	"github.com/openshift/aws-account-operator/pkg/controller/account"
 	"github.com/prometheus/client_golang/prometheus"
-	kubeclientpkg "sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 )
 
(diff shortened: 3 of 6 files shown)


liamwazherealso:
The reason I separated the code from localmetrics was that I thought that we may want to run this in a higher frequency than the localmetrics update. Was still figuring that out. Seems not to be the case so I will consolidate.

pkg/localmetrics/localmetrics.go
```diff
@@ -113,7 +114,7 @@ func UpdateAWSMetrics(kubeClient kubeclientpkg.Client, hour int) {
 
 	d := time.Duration(hour) * time.Hour
 	for range time.Tick(d) {
-		accountTotal, err := account.TotalAwsAccounts(awsClient)
+		accountTotal := totalawsaccountwatcher.TotalAccountWatcher.awsLimit
```
fahlmant: I think at this point, since we already have this functionality in the totalaccountwatcher, this function can be removed, and we can put this line:
`MetricTotalAWSAccounts.With(prometheus.Labels{"name": "aws-account-operator"}).Set(float64(accountTotal))`

on line 96 of the `pkg/totalawsaccountwatcher/totalawsaccountwatcher.go` file

liamwazherealso: Will do

cmd/manager/main.go
```diff
@@ -119,7 +119,7 @@ func main() {
 	// Configure metrics if it errors log the error but continue
 	if err := metrics.ConfigureMetrics(context.TODO(), *metricsServer); err != nil {
 		log.Error(err, "Failed to configure Metrics")
-		os.Exit(1)
+		// os.Exit(1)
```
fahlmant: Please remove the comment

liamwazherealso: ack

cmd/manager/main.go
```diff
@@ -126,13 +128,20 @@ func main() {
 	// work
 	stopCh := signals.SetupSignalHandler()
 
+	notClient, err := client.New(cfg, client.Options{})
```
jharrington22: @liamwazherealso Do you think we want to keep this var name as `notClient` :slightly_smiling_face:

pkg/totalaccountwatcher/totalaccountwatcher.go
```diff
@@ -0,0 +1,155 @@
...
+	accountTotal, err := TotalAwsAccounts()
+	if err != nil {
+		log.Error(err, "Failed to get AWS account total from AWS api")
+	}
+
+	TotalAccountWatcher.Total = accountTotal
+	log.Info(fmt.Sprintf("Total AWS accounts: %d", TotalAccountWatcher.Total))
+
+	gciOut, err := TotalAccountWatcher.AwsClient.GetCallerIdentity(&sts.GetCallerIdentityInput{})
+	log.Info(fmt.Sprintf("Caller identity: %+v", gciOut))
```
jharrington22: What is this for? I think this might be an unnecessary copy-paste?

pkg/totalaccountwatcher/totalaccountwatcher.go
```diff
@@ -0,0 +1,155 @@
...
+func (s *totalAccountWatcher) UpdateTotalAccounts(log logr.Logger) error {
+
+	// Reset cache
+	secretList := &corev1.SecretList{}
+	secret := &corev1.Secret{}
+
+	listOps := &client1.ListOptions{Namespace: awsv1alpha1.AccountCrNamespace}
+	if err := TotalAccountWatcher.client.List(context.TODO(), listOps, secretList); err != nil {
+		log.Error(err, fmt.Sprintf("Unable to list secrets in namespace %s", awsv1alpha1.AccountCrNamespace))
+	}
```
jharrington22: This is unnecessary now?

pkg/totalaccountwatcher/totalaccountwatcher.go
```diff
@@ -0,0 +1,155 @@
...
+	log.Info(fmt.Sprintf("Total AWS accounts: %d", TotalAccountWatcher.Total))
+
+	gciOut, err := TotalAccountWatcher.AwsClient.GetCallerIdentity(&sts.GetCallerIdentityInput{})
+	log.Info(fmt.Sprintf("Caller identity: %+v", gciOut))
+
+	if err != nil {
+		log.Error(err, fmt.Sprintf("Failed to get total number of AWS accounts: %s", err))
+	} else {
+		localmetrics.MetricTotalAWSAccounts.With(prometheus.Labels{"name": "aws-account-operator"}).Set(float64(accountTotal))
+	}
```
jharrington22: Should this be moved up now?

pkg/totalaccountwatcher/totalaccountwatcher.go
```diff
@@ -0,0 +1,155 @@
...
+	// Ensure we paginate through the account list
+	for {
+		awsAccountList, err := TotalAccountWatcher.AwsClient.ListAccounts(&organizations.ListAccountsInput{NextToken: nextToken})
+		if err != nil {
+			errMsg := "Error getting a list of accounts"
+			if aerr, ok := err.(awserr.Error); ok {
+				errMsg = fmt.Sprintf("Failed to get account list with error code %s", aerr.Message())
+			}
+			return 0, errors.New(errMsg)
+		}
```
jharrington22: This probably needs a refactor now since we don't ever want to return `0` if we get an error we just want to return the existing total correct? We may want log an error to say that the API called failed, @fahlmant whats your thoughts on logging this error since we wont do anything with it? I think it would be good to know when the call fails eg the total account number has been consistent and we want to see if the call is failing?

pkg/totalaccountwatcher/totalaccountwatcher.go
```diff
@@ -0,0 +1,155 @@
...
+	secret := &corev1.Secret{}
+
+	listOps := &client1.ListOptions{Namespace: awsv1alpha1.AccountCrNamespace}
+	if err := TotalAccountWatcher.client.List(context.TODO(), listOps, secretList); err != nil {
+		log.Error(err, fmt.Sprintf("Unable to list secrets in namespace %s", awsv1alpha1.AccountCrNamespace))
+	}
+
+	if err := TotalAccountWatcher.client.Get(context.TODO(), types.NamespacedName{Name: AwsSecretName, Namespace: "aws-account-operator"}, secret); err != nil {
+		log.Error(err, fmt.Sprintf("Unable to get secrets in namespace %s", awsv1alpha1.AccountCrNamespace))
+	}
```
jharrington22: Do we need this?

jharrington22:
@liamwazherealso this job is also failing CI

cmd/manager/main.go
liamwazherealso: I like the ring to it. However, I will refactor.

pkg/totalaccountwatcher/totalaccountwatcher.go
liamwazherealso: This was a debug log that I kept in. Can be removed.

pkg/totalaccountwatcher/totalaccountwatcher.go
liamwazherealso: Yes, moving up.

pkg/totalaccountwatcher/totalaccountwatcher.go
liamwazherealso: This case is taken care of in ```	if err != nil {
		log.Error(err, fmt.Sprintf("Failed to get total number of AWS accounts: %s", err))
	} else {
		localmetrics.MetricTotalAWSAccounts.With(prometheus.Labels{"name": "aws-account-operator"}).Set(float64(accountTotal))
	}
```

File: cmd/manager/main.go [hunk 9]
```diff
@@ -126,13 +128,20 @@ func main() {
 	// work
 	stopCh := signals.SetupSignalHandler()
 
+	client1, err := client.New(cfg, client.Options{})
```
jharrington22: I think we can do better than `client1` `accountWatcherClient` maybe?

pkg/controller/account/account_controller.go
```diff
@@ -333,21 +335,18 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
 		var awsAccountID string
 
 		if currentAcctInstance.Spec.AwsAccountID == "" {
+			// before doing anything make sure we are not over the limit if we are just error
+			accountTotal := totalaccountwatcher.TotalAccountWatcher.Total
+			if accountTotal >= AwsLimit {
+				reqLogger.Error(ErrAwsAccountLimitExceeded, "AWS Account limit reached", "Account Total", accountTotal)
+				return reconcile.Result{}, ErrAwsAccountLimitExceeded
+			} else {
+				reqLogger.Info("Limit not reached")
```
jharrington22: We don't need to log that the limit isn't reached

pkg/totalaccountwatcher/totalaccountwatcher.go
```diff
@@ -0,0 +1,144 @@
...
+
+	accountTotal, err := TotalAwsAccounts()
+	if err != nil {
+		log.Error(err, "Failed to get AWS account total from AWS api")
+	}
+
+	TotalAccountWatcher.Total = accountTotal
+	log.Info(fmt.Sprintf("Total AWS accounts: %d", TotalAccountWatcher.Total))
+
+	if err != nil {
```
jharrington22: Which error is this referring to?

cmd/manager/main.go
liamwazherealso: Yup makes sense to me.

pkg/controller/account/account_controller.go
liamwazherealso: Ack

pkg/totalaccountwatcher/totalaccountwatcher.go
liamwazherealso: The error that was caught in the previous error clause. Removing.

pkg/controller/account/account_controller.go
```diff
@@ -344,10 +344,10 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 			if accountTotal >= AwsLimit {
 				reqLogger.Error(ErrAwsAccountLimitExceeded, "AWS Account limit reached", "Account Total", accountTotal)
 				return reconcile.Result{}, ErrAwsAccountLimitExceeded
-			} else {
-				reqLogger.Info("Limit not reached")
 			}
 
+			totalaccountwatcher.TotalAccountWatcher.Total += 1
```
fahlmant: should we not get this value from the api?

cmd/manager/main.go
```diff
@@ -126,13 +128,20 @@ func main() {
 	// work
 	stopCh := signals.SetupSignalHandler()
 
+	accountWatcherClient, err := client.New(cfg, client.Options{})
```
fahlmant: What are the implications of having two clients?

pkg/controller/account/account_controller.go
```diff
@@ -333,28 +335,25 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
-			reqLogger.Error(ErrAwsAccountLimitExceeded, "AWS Account limit reached", "Account Total", accountTotal)
-			return reconcile.Result{}, ErrAwsAccountLimitExceeded
-		}
 
 		var awsAccountID string
 
 		if currentAcctInstance.Spec.AwsAccountID == "" {
+			// before doing anything make sure we are not over the limit if we are just error
+			accountTotal := totalaccountwatcher.TotalAccountWatcher.Total
+			if accountTotal >= AwsLimit {
```
fahlmant: Any reason not to make this
```		
if totalaccountwatcher.TotalAccountWatcher.Total >= AwsLimit {
```

liamwazherealso: No problem. Chose it for style.

pkg/totalaccountwatcher/totalaccountwatcher.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// TotalAccountWatcher will trigger AwsLimitUpdate every `scanInternal` and only stop if the operator is killed or a
+// message is sent on the stopCh
+func (s *totalAccountWatcher) Start(log logr.Logger, stopCh <-chan struct{}) {
+	log.Info("Starting the totalAccountWatcher")
+	for {
+		select {
+		case <-time.After(s.watchInterval):
+			log.Info("totalAccountWatcher: updating a")
```
jharrington22: What is this log message?

pkg/totalaccountwatcher/totalaccountwatcher.go
```diff
@@ -0,0 +1,141 @@
...
+
+	listOps := &client.ListOptions{Namespace: awsv1alpha1.AccountCrNamespace}
+	if err := TotalAccountWatcher.client.List(context.TODO(), listOps, secretList); err != nil {
+		log.Error(err, fmt.Sprintf("Unable to list secrets in namespace %s", awsv1alpha1.AccountCrNamespace))
+	}
+
+	accountTotal, err := TotalAwsAccounts()
+	if err != nil {
+		log.Error(err, fmt.Sprintf("Failed to get total number of AWS accounts: %s", err))
+	} else {
```
jharrington22: Did we talk about returning after this error?

pkg/totalaccountwatcher/totalaccountwatcher.go
```diff
@@ -0,0 +1,141 @@
+package totalaccountwatcher
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/aws/aws-sdk-go/service/organizations"
```
fahlmant: Can you sort this imports to group together 
1) Local packages
2) k8s packges
3) other

pkg/totalaccountwatcher/totalaccountwatcher.go
```diff
@@ -0,0 +1,141 @@
...
+	client        client.Client
+	Total         int
+}
+
+// Initialize creates a global instance of the TotalAccountWatcher
+func Initialize(client client.Client, watchInterval time.Duration) {
+	log.Info("Initializing the totalAccountWatcher")
+
+	AwsClient, err := awsclient.GetAWSClient(client, awsclient.NewAwsClientInput{
+		SecretName: AwsSecretName,
```
fahlmant: I feel like this is already defined somewhere and can be used just like namespace, as an exported value

pkg/totalaccountwatcher/totalaccountwatcher.go
```diff
@@ -0,0 +1,141 @@
...
+// UpdateTotalAccounts will update the TotalAccountWatcher's total field
+func (s *totalAccountWatcher) UpdateTotalAccounts(log logr.Logger) error {
+
+	// Reset cache
+	secretList := &corev1.SecretList{}
+
+	listOps := &client.ListOptions{Namespace: awsv1alpha1.AccountCrNamespace}
+	if err := TotalAccountWatcher.client.List(context.TODO(), listOps, secretList); err != nil {
+		log.Error(err, fmt.Sprintf("Unable to list secrets in namespace %s", awsv1alpha1.AccountCrNamespace))
+	}
```
jharrington22: @liamwazherealso what is happening here? I don't think this is required.

pkg/totalaccountwatcher/totalaccountwatcher.go
fahlmant: I don't think it is. You're setting accountTotal to the first returned value, which is `0` in this case.

pkg/totalaccountwatcher/totalaccountwatcher.go
```diff
@@ -0,0 +1,141 @@
...
+		log.Error(err, fmt.Sprintf("Failed to get total number of AWS accounts: %s", err))
+	} else {
+		localmetrics.MetricTotalAWSAccounts.With(prometheus.Labels{"name": "aws-account-operator"}).Set(float64(accountTotal))
+	}
+
+	if accountTotal != TotalAccountWatcher.Total {
+		log.Info("Updating total from %d to %d", TotalAccountWatcher.Total, accountTotal)
+	}
+
+	TotalAccountWatcher.Total = accountTotal
```
fahlmant: I think this should go in the `if` block right above. No point in updating if they are the same.

cmd/manager/main.go
jharrington22: @fahlmant since we are using this only to list secrets once every 15 minutes I don't expect their to be a big issue.

pkg/totalaccountwatcher/totalaccountwatcher.go
liamwazherealso: Removed

pkg/totalaccountwatcher/totalaccountwatcher.go
liamwazherealso: Done.

pkg/totalaccountwatcher/totalaccountwatcher.go
liamwazherealso: Done

pkg/totalaccountwatcher/totalaccountwatcher.go
liamwazherealso: This is the code that reloaded the cache. Removing.

pkg/totalaccountwatcher/totalaccountwatcher.go
liamwazherealso: It's exported in the account-controller. But importing that would create an import cycle.

pkg/totalaccountwatcher/totalaccountwatcher.go
liamwazherealso: Good :eye:

pkg/controller/account/account_controller.go
liamwazherealso: The value is used in the log message as well.

pkg/totalaccountwatcher/totalaccountwatcher.go
fahlmant: Can we move those values to utils or something?

pkg/totalaccountwatcher/totalaccountwatcher.go
jharrington22: @fahlmant only if there is no error, hes returning an error with the zero

pkg/controller/account/account_controller.go
liamwazherealso: Nope implementing

pkg/totalaccountwatcher/totalaccountwatcher.go
liamwazherealso: Done

pkg/controller/account/account_controller.go
```diff
@@ -333,28 +334,24 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
+				return reconcile.Result{}, ErrAwsAccountLimitExceeded
+			}
 
 			// Build Aws Account
 			awsAccountID, err = r.BuildAccount(reqLogger, awsSetupClient, currentAcctInstance)
 			if err != nil {
 				return reconcile.Result{}, err
 			}
 
+			totalaccountwatcher.TotalAccountWatcher.Total += 1
```
fahlmant: Just a quick though, want to make sure I understand. So since this var is from an instance of the totalAccountWatcher object that lives in the main, this value gets persisted through each reconcile right? Have you tested creating an account and then reconciling again, and the outcome? I assume this works but just want to make sure this actually writes to memory outside the scope of the reconcile function.

liamwazherealso:
@jharrington22  Please approve

PR #172, opened 2019-09-30
MVP for AWSFederatedAccountAccess controller 

rafael-azevedo:
This PR adds basic MVP logic for the AWS Federeated Access self service.

### DIFF
--- a/deploy/crds/aws_v1alpha1_awsfederatedaccountaccess_crd.yaml
+++ b/deploy/crds/aws_v1alpha1_awsfederatedaccountaccess_crd.yaml
@@ -3,6 +3,13 @@ kind: CustomResourceDefinition
 metadata:
   name: awsfederatedaccountaccesses.aws.managed.openshift.io
 spec:
+  additionalPrinterColumns:
+  - JSONPath: .status.state
+    name: State
+    type: string
+  - JSONPath: .metadata.creationTimestamp
+    name: Age
+    type: date
   group: aws.managed.openshift.io
   names:
     kind: AWSFederatedAccountAccess
--- a/pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
+++ b/pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
@@ -1,0 +1,358 @@
+package awsfederatedaccountaccess
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/aws/aws-sdk-go/service/sts"
+	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
+	"github.com/openshift/aws-account-operator/pkg/awsclient"
+	controllerutils "github.com/openshift/aws-account-operator/pkg/controller/utils"
+
+	corev1 "k8s.io/api/core/v1"
+	k8serr "k8s.io/apimachinery/pkg/api/errors"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+// Custom errors
+
+// ErrFederatedAccessRoleNotFound indicates the role requested by AWSFederatedAccountAccess Cr was not found as a AWSFederatedRole Cr
+var ErrFederatedAccessRoleNotFound = errors.New("FederatedAccessRoleNotFound")
+
+// ErrFederatedAccessRoleFailedCreate indicates that the AWSFederatedRole requested failed to be created in the account requested by the AWSFederatedAccountAccess CR
+var ErrFederatedAccessRoleFailedCreate = errors.New("FederatedAccessRoleFailedCreate")
+
+var log = logf.Log.WithName("controller_awsfederatedaccountaccess")
+
+// Add creates a new AWSFederatedAccountAccess Controller and adds it to the Manager. The Manager will set fields on the Controller
+// and Start it when the Manager is Started.
+func Add(mgr manager.Manager) error {
+	return add(mgr, newReconciler(mgr))
+}
+
+// newReconciler returns a new reconcile.Reconciler
+func newReconciler(mgr manager.Manager) reconcile.Reconciler {
+	return &ReconcileAWSFederatedAccountAccess{
+		client: mgr.GetClient(),
+		scheme: mgr.GetScheme(),
+	}
+}
+
+// add adds a new Controller to mgr with r as the reconcile.Reconciler
+func add(mgr manager.Manager, r reconcile.Reconciler) error {
+	// Create a new controller
+	c, err := controller.New("awsfederatedaccountaccess-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to primary resource AWSFederatedAccountAccess
+	err = c.Watch(&source.Kind{Type: &awsv1alpha1.AWSFederatedAccountAccess{}}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// blank assignment to verify that ReconcileAWSFederatedAccountAccess implements reconcile.Reconciler
+var _ reconcile.Reconciler = &ReconcileAWSFederatedAccountAccess{}
+
+// ReconcileAWSFederatedAccountAccess reconciles a AWSFederatedAccountAccess object
+type ReconcileAWSFederatedAccountAccess struct {
+	// This client, initialized using mgr.Client() above, is a split client
+	// that reads objects from the cache and writes to the apiserver
+	client client.Client
+	scheme *runtime.Scheme
+}
+
+// Reconcile reads that state of the cluster for a AWSFederatedAccountAccess object and makes changes based on the state read
+// and what is in the AWSFederatedAccountAccess.Spec
+func (r *ReconcileAWSFederatedAccountAccess) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	reqLogger.Info("Reconciling AWSFederatedAccountAccess")
+
+	// Fetch the AWSFederatedAccountAccess instance
+	currentFAA := &awsv1alpha1.AWSFederatedAccountAccess{}
+	err := r.client.Get(context.TODO(), request.NamespacedName, currentFAA)
+	if err != nil {
+		if k8serr.IsNotFound(err) {
+			// Request object not found, could have been deleted after reconcile request.
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
+
+	// If the state is ready don't do anything
+	if currentFAA.Status.State == awsv1alpha1.AWSFederatedAccountStateReady || currentFAA.Status.State == awsv1alpha1.AWSFederatedAccountStateFailed {
+		return reconcile.Result{}, nil
+	}
+
+	// Get a list of all available roles
+	federatedRoleList := &awsv1alpha1.AWSFederatedRoleList{}
+	if r.client.List(context.TODO(), &client.ListOptions{}, federatedRoleList); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// Create map to make finding role easier
+	roleMap := make(map[string]awsv1alpha1.AWSFederatedRole)
+	for _, role := range federatedRoleList.Items {
+		roleMap[role.Name] = role
+	}
+
+	// See if requested role exists
+	if _, ok := roleMap[currentFAA.Spec.AWSFederatedRoleName]; !ok {
+		SetStatuswithCondition(currentFAA, "Requested role does not exist", awsv1alpha1.AWSFederatedAccountFailed, awsv1alpha1.AWSFederatedAccountStateFailed)
+		reqLogger.Error(ErrFederatedAccessRoleNotFound, fmt.Sprintf("Resquested role '%s' not found", currentFAA.Spec.AWSFederatedRoleName))
+
+		err := r.client.Status().Update(context.TODO(), currentFAA)
+		if err != nil {
+			reqLogger.Error(err, fmt.Sprintf("Status update for %s failed", currentFAA.Name))
+			return reconcile.Result{}, err
+		}
+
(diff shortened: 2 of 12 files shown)


deploy/crds/aws_v1alpha1_awsfederatedaccountaccess_crd.yaml
```diff
@@ -3,6 +3,13 @@ kind: CustomResourceDefinition
 metadata:
   name: awsfederatedaccountaccesses.aws.managed.openshift.io
 spec:
+  additionalPrinterColumns:
```
fahlmant: Did you test these?

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -0,0 +1,368 @@
...
+	// Watch for changes to primary resource AWSFederatedAccountAccess
+	err = c.Watch(&source.Kind{Type: &awsv1alpha1.AWSFederatedAccountAccess{}}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return err
+	}
+
+	// watch for changes to secondary resource AWSFederatedRole
+	err = c.Watch(&source.Kind{Type: &awsv1alpha1.AWSFederatedRole{}}, &handler.EnqueueRequestForOwner{
+		IsController: true,
+		OwnerType:    &awsv1alpha1.AccountPool{},
```
fahlmant: Are AWSFederatedRoles owned by an accountpool?

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -0,0 +1,368 @@
...
+			// Request object not found, could have been deleted after reconcile request.
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
+
+	// If the state is applied don't do anything
```
fahlmant: What is 'applied'? This comment doesn't seem to match what you're checking, which is "Ready" or "Failed"

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -0,0 +1,368 @@
...
+	}
+
+	// Get a list of all available roles
+	federatedRoleList := &awsv1alpha1.AWSFederatedRoleList{}
+	if r.client.List(context.TODO(), &client.ListOptions{}, federatedRoleList); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// Create map to make finding role easier
+	roleMap := make(map[string]awsv1alpha1.AWSFederatedRole)
```
fahlmant: This could probably be used elsewhere. Maybe could be moved to util function.

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -0,0 +1,368 @@
...
+		awsClientErr := fmt.Sprintf("Unable to create aws client for region ")
+		reqLogger.Error(err, awsClientErr)
+		return reconcile.Result{}, err
+	}
+
+	// here create the custom policy in the cluster account
+	err = r.createIAMPolicy(awsClient, roleMap[currentFAA.Spec.AWSFederatedRoleName], *currentFAA)
+	if err != nil {
+		// if we were unable to create the policy fail this CR.
+		SetStatuswithCondition(currentFAA, "Failed to create custom policy", awsv1alpha1.AWSFederatedAccountFailed, awsv1alpha1.AWSFederatedAccountStateFailed)
```
fahlmant: Should this be set to failed? What if there's an API issue and this needs to just be retried for it work? As of now, this would permanently make this CR dead, i.e. never reconciled again without human intervention due to line 111.

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -0,0 +1,368 @@
...
+		}
+
+		return reconcile.Result{}, nil
+
+	}
+
+	// create role role and apply custom policys and awsmanagedpolicies
+	err = r.createIAMRole(awsClient, roleMap[currentFAA.Spec.AWSFederatedRoleName], *currentFAA)
+	if err != nil {
+		SetStatuswithCondition(currentFAA, "Failed to create role", awsv1alpha1.AWSFederatedAccountFailed, awsv1alpha1.AWSFederatedAccountStateFailed)
```
fahlmant: Again does API failure mean the CR should go into a failed state?

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -0,0 +1,368 @@
...
+		if err != nil {
+			reqLogger.Error(err, fmt.Sprintf("Status update for %s failed", currentFAA.Name))
+			return reconcile.Result{}, err
+		}
+
+		return reconcile.Result{}, nil
+	}
+
+	// Get account number of cluster account
+	gciOut, err := awsClient.GetCallerIdentity(&sts.GetCallerIdentityInput{})
```
fahlmant: Same question of API failure

deploy/crds/aws_v1alpha1_awsfederatedaccountaccess_crd.yaml
rafael-azevedo: yes

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
rafael-azevedo: @fahlmant Actually this should have no secondary resource right now

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
rafael-azevedo: One of the cards coming out of this is to move the shared functionally to a util package there are more functions that should be moved as well.

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
rafael-azevedo: We can create a retry in the future , this is just the MVP. We also before this goes into production have cleaning steps whenever this is failed or else it will not work on next reconcile if the objects attempted to be created have the same name.

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
jharrington22: You should be able to use this when you call `r.client.List()` to return the single role. That will eliminate some the code below. 

`&client.ListOptions{FieldSelector: fields.Set{"metadata.name": currentFAA.Spec.AWSFederatedRoleName}.AsSelector().String()}`

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -0,0 +1,360 @@
...
+	}
+
+	// Get aws client
+	awsClient, err := awsclient.GetAWSClient(r.client, awsclient.NewAwsClientInput{
+		SecretName: currentFAA.Spec.AwsCredentialSecret.Name,
+		NameSpace:  currentFAA.Spec.AwsCredentialSecret.Namespace,
+		AwsRegion:  "us-east-1",
+	})
+	if err != nil {
+		awsClientErr := fmt.Sprintf("Unable to create aws client for region ")
```
jharrington22: Missing format parameters in `fmt.Sprintf()`

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
jharrington22: Happy to leave these as is provided we track it after MVP.

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
jharrington22: Also this var could be renamed `awsClientErrMsg`

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
jharrington22: We have the account CR where we could retrieve this from instead of calling AWS on every reconcile loop. OCM would have to provide the Account CR or Claim name.

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -0,0 +1,360 @@
...
+			return reconcile.Result{}, err
+		}
+
+		return reconcile.Result{}, nil
+	}
+
+	accountID := *gciOut.Account
+	// Add requested aws managed policies to the role
+	awsManagedPolicyNames := []string{}
+	// add all aws managed policy names to a array
```
jharrington22: Capital

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -0,0 +1,360 @@
...
+		NameSpace:  currentFAA.Spec.AwsCredentialSecret.Namespace,
+		AwsRegion:  "us-east-1",
+	})
+	if err != nil {
+		awsClientErr := fmt.Sprintf("Unable to create aws client for region ")
+		reqLogger.Error(err, awsClientErr)
+		return reconcile.Result{}, err
+	}
+
+	// here create the custom policy in the cluster account
```
jharrington22: Capital

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -0,0 +1,360 @@
...
+		Statement: statements,
+	}
+
+	// marshal policydoc to json
+	jsonPolicyDoc, err := json.Marshal(&policyDoc)
+	if err != nil {
+		return errors.New("Error marshalling jsonPolicy doc")
+	}
+
+	fmt.Printf("\n\n\n STRUCT:::   \n %+v", afr.Spec.AWSCustomPolicy.Statements)
```
jharrington22: Debug line?

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -0,0 +1,360 @@
...
+	fmt.Println(string(jsonPolicyDoc))
```
jharrington22: Debug here too?

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -0,0 +1,360 @@
...
+		Statement []awsStatement
+	}{
+		Version:   "2012-10-17",
+		Statement: statements,
+	}
+
+	// marshal policydoc to json
+	jsonPolicyDoc, err := json.Marshal(&policyDoc)
+	if err != nil {
+		return errors.New("Error marshalling jsonPolicy doc")
```
jharrington22: Would we not return a `fmt.ErrorF()` here instead with the actual `err`?

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
jharrington22: If we did that we would also not need the secret name.

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -0,0 +1,360 @@
...
+	awsPolicyArnPrefix := ""
+
+	if awsManaged == true {
+		awsPolicyArnPrefix = "arn:aws:iam::" + "aws" + ":policy/"
+	} else {
+		awsPolicyArnPrefix = "arn:aws:iam::" + accountID + ":policy/"
+	}
+	policyArns := []string{}
+	for _, policy := range policyNames {
+		policyArns = append(policyArns, awsPolicyArnPrefix+policy)
```
jharrington22: `fmt.Sprintf()` for all of these would be more consistent

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
fahlmant: :+1:

rafael-azevedo:
When attempting to use the fieldselector .String() method I get the following error.

`pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go:111:55: cannot use fields.Set literal.AsSelector().String() (type string) as type fields.Selector in field va
lue:
string does not implement fields.Selector (missing DeepCopySelector method)
make: *** [standard.mk:62: gocheck] Error 2`

@jharrington22

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -0,0 +1,358 @@
...
+	if currentFAA.Status.State == awsv1alpha1.AWSFederatedAccountStateReady || currentFAA.Status.State == awsv1alpha1.AWSFederatedAccountStateFailed {
+		return reconcile.Result{}, nil
+	}
+
+	// Get a list of all available roles
+	federatedRoleList := &awsv1alpha1.AWSFederatedRoleList{}
+
+	if r.client.List(context.TODO(), &client.ListOptions{FieldSelector: fields.Set{"metadata.name": currentFAA.Spec.AWSFederatedRoleName}.AsSelector()}, federatedRoleList); err != nil {
+		return reconcile.Result{}, err
+	}
```
jharrington22:
```suggestion

	listOptions := &client.ListOptions{FieldSelector: fields.Set{"metadata.name": currentFAA.Spec.AWSFederatedRoleName}.AsSelector()}
	err = r.client.List(context.TODO(), listOptions, federatedRoleList)

	if err != nil {
		return reconcile.Result{}, err
	}
```

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
jharrington22: @rafael-azevedo doesn't look like you updated this?

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -0,0 +1,358 @@
...
+		Resource  []string               `json:"Resource,omitempty"`
+		Principal *awsv1alpha1.Principal `json:"Principal,omitempty"`
+	}
+
+	statements := []awsStatement{}
+
+	for _, sm := range afr.Spec.AWSCustomPolicy.Statements {
+		var a awsStatement = awsStatement(sm)
+		statements = append(statements, a)
+	}
```
fahlmant:
```suggestion
	}	statements := []awsStatement{}

	for _, statement := range role.Spec.AWSCustomPolicy.Statements {
		statements = append(statements, awsStatement(statement))
	}
```

fahlmant: idk what happened with the formatting, but this is obviously for lines 246-249, not starting at 249

PR #173, opened 2019-09-30
Expand aws api errors

liamwazherealso:
Re-pull request from https://github.com/openshift/aws-account-operator/pull/102


pkg/controller/account/account_controller.go
```diff
@@ -35,7 +35,7 @@ import (
 var log = logf.Log.WithName("controller_account")
 
 const (
-	awsLimit                = 4700
+	awsLimit                = 4670
```
fahlmant: This can be removed from this commit.

PR #176, opened 2019-10-01
Correct service name to match whats on cluster and in monitoring

fahlmant:
Before, the service name defaulted to the name of the operator, which the route creation used for the route name. However, other services may already be created with the name name, so a default name was built in to the library. This PR explicitly uses the name of the operator for the service to match the service/route names our current monitoring set up expects. 

tl;dr
We use this route:
aws-account-operator-aws-account-operator.example.com

After #175, the operator creates this route:
operator-metrics-aws-account-operator.example.com

This PR ensures that the first one is created/updated, not the second one.


fahlmant:
Closed due to incorrect branch name

PR #181, opened 2019-10-09
Change secret reference to account name reference

fahlmant:
This changes the AWSFederatedAccountAccess to hold a reference to an Account CR rather than a Secret. 

This does 2 things:
1. The code no longer has to make an external STS call for the customer AWS Account ID
2. The code can build the secret name based on the account name


pkg/apis/aws/v1alpha1/awsfederatedaccountaccess_types.go
```diff
@@ -23,10 +23,10 @@ const (
 // AWSFederatedAccountAccessSpec defines the desired state of AWSFederatedAccountAccess
 // +k8s:openapi-gen=true
 type AWSFederatedAccountAccessSpec struct {
-	// CustomerAccountID holds the account number associated with the AWS account that will be used to access the associated role
+	// CustomerAccountID holds the external AWS account ID
 	CustomerAccountID string `json:"customerAccountID"`
```
jharrington22: ExternalCustomerAwsAccountID? I still think that CustomerAccountID might be confusing?

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -131,10 +131,13 @@ func (r *ReconcileAWSFederatedAccountAccess) Reconcile(request reconcile.Request
...
 
+	// Use the account name reference to get the correct secret name
+	secretName := currentFAA.Spec.AccountReference + "-secret"
+
 	// Get aws client
 	awsClient, err := awsclient.GetAWSClient(r.client, awsclient.NewAwsClientInput{
-		SecretName: currentFAA.Spec.AwsCredentialSecret.Name,
-		NameSpace:  currentFAA.Spec.AwsCredentialSecret.Namespace,
+		SecretName: secretName,
+		NameSpace:  "aws-account-operator",
```
jharrington22: We can use `awsv1alpha1.AccountCrNamespace` here

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -175,22 +178,13 @@ func (r *ReconcileAWSFederatedAccountAccess) Reconcile(request reconcile.Request
 		return reconcile.Result{}, nil
 	}
 
-	// Get account number of cluster account
-	gciOut, err := awsClient.GetCallerIdentity(&sts.GetCallerIdentityInput{})
+	accountCR := &awsv1alpha1.Account{}
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: currentFAA.Spec.AccountReference, Namespace: "aws-account-operator"}, accountCR)
```
jharrington22: Same constant can be used here for namespace

pkg/apis/aws/v1alpha1/awsfederatedaccountaccess_types.go
fahlmant: :+1:

PR #185, opened 2019-10-15
Add BYOC fields to AccountClaim Spec

fahlmant:
Adding a BYOC fields to support byoc accountclaim controller logic.


pkg/apis/aws/v1alpha1/accountclaim_types.go
```diff
@@ -99,9 +102,8 @@ type LegalEntity struct {
 	ID   string `json:"id"`
 }
 
-// AwsCredentialSecret contains the name of the secret and name of the namespace
-// where UHC would like the AWS credentials secret to be placed
-type AwsCredentialSecret struct {
+// SecretRef contains the name of a secret and its namespace
+type SecretRef struct {
```
jharrington22: `type BYOCSecretRef struct {` ?

fahlmant: So the secret is just a name and namespace combination, so I moved the AWS Credentials Secret Ref struct to be a generic SecretRef struct so it can be used multiple places

PR #187, opened 2019-10-23
Add claimLinkNamespace field to Account CR

fahlmant:
Currently, an account CR only has a reference to the name of an accountClaim. However, these names are not unique. In order to ensure the Account knows the accountClaim its linked to, the namespace of the accountClaim should also be referenced in its spec.

### DIFF
--- a/deploy/crds/aws_v1alpha1_account_crd.yaml
+++ b/deploy/crds/aws_v1alpha1_account_crd.yaml
@@ -52,6 +52,8 @@ spec:
               type: boolean
             claimLink:
               type: string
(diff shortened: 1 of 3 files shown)


deploy/crds/aws_v1alpha1_account_crd.yaml
```diff
@@ -52,6 +52,8 @@ spec:
               type: boolean
             claimLink:
               type: string
+            claimLinkNamespace:
```
lisa: What is it for? (Does it make sense to add a `description` for it?)

PR #188, opened 2019-10-25
Implement BYOC Account controller logic

fahlmant:
Account Controller BYOC Logic

1. Skip creating the AWS account
2. Create new IAM access key for root credentials
3. Delete console login (??)
4. Delete original root IAM access keys that were provided
5. Assume user in this account using new root credentials (I think this and 6 are the same)
6. Following existing account setup logic
7. Skip creating support tickets for limit and enterprise support logic
8. Update status and conditions

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -22,6 +22,7 @@ import (
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	kubeclientpkg "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -65,6 +66,9 @@ const (
 	AccountReady = "Ready"
 	// AccountPendingVerification indicates verification (of AWS limits and Enterprise Support) is pending
 	AccountPendingVerification = "PendingVerification"
+	byocRole                   = "BYOCAdminAccess"
+
+	adminAccessArn = "arn:aws:iam::aws:policy/AdministratorAccess"
 )
 
 var awsAccountID string
@@ -243,20 +247,9 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.Info("Reconciling Account")
 
-	// We expect this secret to exist in the same namespace Account CR's are created
-	awsSetupClient, err := awsclient.GetAWSClient(r.Client, awsclient.NewAwsClientInput{
-		SecretName: controllerutils.AwsSecretName,
-		NameSpace:  awsv1alpha1.AccountCrNamespace,
-		AwsRegion:  "us-east-1",
-	})
-	if err != nil {
-		reqLogger.Error(err, "Failed to get AWS client")
-		return reconcile.Result{}, err
-	}
-
 	// Fetch the Account instance
 	currentAcctInstance := &awsv1alpha1.Account{}
-	err = r.Client.Get(context.TODO(), request.NamespacedName, currentAcctInstance)
+	err := r.Client.Get(context.TODO(), request.NamespacedName, currentAcctInstance)
 	if err != nil {
 		if k8serr.IsNotFound(err) {
 			return reconcile.Result{}, nil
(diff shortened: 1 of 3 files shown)


pkg/controller/account/account_controller.go
```diff
@@ -242,143 +243,231 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
 
-		var awsAccountID string
-
-		if currentAcctInstance.Spec.AwsAccountID == "" {
-			// before doing anything make sure we are not over the limit if we are just error
-			if totalaccountwatcher.TotalAccountWatcher.Total >= AwsLimit {
-				reqLogger.Error(ErrAwsAccountLimitExceeded, "AWS Account limit reached", "Account Total", totalaccountwatcher.TotalAccountWatcher.Total)
-				return reconcile.Result{}, ErrAwsAccountLimitExceeded
+			// Create new root access keys
+			userSecretInfo, err := CreateUserAccessKey(awsSetupClient, userName)
```
jharrington22: This will error if two AccessKey's already exist, we should list them first?

pkg/controller/account/account_controller.go
```diff
@@ -242,143 +243,243 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
+		})
+		if err != nil {
+			reqLogger.Error(err, "Failed to get AWS client")
+			return reconcile.Result{}, err
+		}
 
-			awsAccountID = currentAcctInstance.Spec.AwsAccountID
+		// Test PendingVerification state creating support case and checking for case status
+		if currentAcctInstance.Status.State == AccountPendingVerification {
+			// reqLogger.Info("Account in PendingVerification state", "AccountID", currentAcctInstance.Spec.AwsAccountID)
```
liamwazherealso: Uncomment this?

fahlmant: I was moving this existing code around, so I'm not sure who commented it out or for what reason. I think it would be better to just remove

pkg/controller/account/account_controller.go
```diff
@@ -242,143 +243,243 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
+			reqLogger.Info(fmt.Sprintf(`Case %s not resolved, 
+			trying again in %d minutes`,
+				currentAcctInstance.Status.SupportCaseID,
+				intervalBetweenChecksMinutes))
+			return reconcile.Result{RequeueAfter: intervalBetweenChecksMinutes * time.Minute}, nil
+		}
+
+		// Update account Status.Claimed to true if the account is ready and the claim link is not empty
+		if currentAcctInstance.Status.State == AccountReady && currentAcctInstance.Spec.ClaimLink != "" {
+			if currentAcctInstance.Status.Claimed != true {
```
liamwazherealso: Why not consolidate 420 + 421 into one line.

pkg/controller/account/account_controller.go
```diff
@@ -242,143 +243,243 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
+			if k8serr.IsNotFound(err) {
+				return reconcile.Result{}, nil
+			}
 			return reconcile.Result{}, err
 		}
 
-		// Case Resolved, account is Ready
-		if resolved {
-			reqLogger.Info(fmt.Sprintf("Case %s resolved", currentAcctInstance.Status.SupportCaseID))
+		//Get credentials
```
jharrington22: Space after // :)

pkg/controller/account/account_controller.go
```diff
@@ -242,143 +243,243 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
-	}
+			err = r.Client.Get(context.TODO(),
+				types.NamespacedName{Name: accountClaim.Spec.BYOCSecretRef.Name, Namespace: accountClaim.Spec.BYOCSecretRef.Namespace},
+				accountClaimSecret)
 
-	if (currentAcctInstance.Status.State == "") && (currentAcctInstance.Status.Claimed == false) {
+			accessKeyID := string(accountClaimSecret.Data["aws_access_key_id"])
 
-		var awsAccountID string
+			accessKeyList, err := awsSetupClient.ListAccessKeys(&iam.ListAccessKeysInput{UserName: &userName})
```
jharrington22: You can use `aws.String("root")` here to return a pointer

pkg/controller/account/iam.go
```diff
@@ -235,3 +244,42 @@ func getStsCredentials(reqLogger logr.Logger, client awsclient.Client, iamRoleNa
...
+	iamAdministratorPolicy := "arn:aws:iam::aws:policy/AdministratorAccess"
+
+	assumeRolePolicyDoc := struct {
+		Version   string
+		Statement []awsStatement
+	}{
+		Version: "2012-10-17",
+		Statement: []awsStatement{{
+			Effect: "Allow",
+			Action: []string{"sts:AssumeRole"},
```
jharrington22: The IAM user you are creating doesn't need `sts:AssumeRole` correct?

pkg/controller/account/account_controller.go
```diff
@@ -254,134 +268,261 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
+
+		// Get associated AccountClaim
+		accountClaim := &awsv1alpha1.AccountClaim{}
+
+		err := r.Client.Get(context.TODO(),
+			types.NamespacedName{Name: currentAcctInstance.Spec.ClaimLink, Namespace: currentAcctInstance.Spec.ClaimLinkNamespace},
+			accountClaim)
+		if err != nil {
+			if k8serr.IsNotFound(err) {
+				return reconcile.Result{}, nil
```
jharrington22: If the claim is not found we just return and do nothing?

pkg/controller/account/account_controller.go
```diff
@@ -254,134 +268,261 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
+			SecretName: accountClaim.Spec.BYOCSecretRef.Name,
+			NameSpace:  accountClaim.Spec.BYOCSecretRef.Namespace,
+			AwsRegion:  "us-east-1",
+		})
+
+		reqLogger.Info("Marking BYOC account claimed")
+		// Ensure the account is marked as claimed
+		if currentAcctInstance.Status.Claimed != true {
+			currentAcctInstance.Status.Claimed = true
+			//return reconcile.Result{}, r.statusUpdate(reqLogger, currentAcctInstance)
```
jharrington22: Should we be returning here? What happens if the controller is interrupted here and we recover?

pkg/controller/account/account_controller.go
fahlmant: Yes because then we cannot get the secret holding the BYOC IAM credentials.

pkg/controller/account/account_controller.go
fahlmant: Ah yes, I was having problems with the status update here. Should be uncommented.

pkg/controller/account/account_controller.go
```diff
@@ -269,124 +262,272 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
+			reqLogger.Info("Updating BYOC to creating")
+			currentAcctInstance.Status.State = AccountCreating
+			SetAccountStatus(reqLogger, currentAcctInstance, "BYOC Account Creating", awsv1alpha1.AccountCreating, "Creating")
+			err = r.Client.Status().Update(context.TODO(), currentAcctInstance)
 			if err != nil {
 				return reconcile.Result{}, err
 			}
+			if err != nil {
+				return reconcile.Result{}, err
+			}
```
jharrington22: Is this a double up?

PR #189, opened 2019-11-01
Add check for AWSCustomPolicy and AWSManagedPolicies

sam-nguyen7:
At least one of the fields need to NOT be empty for CR to reconcile.
If both are empty then exit with nil.

### DIFF
--- a/pkg/controller/awsfederatedrole/awsfederatedrole_controller.go
+++ b/pkg/controller/awsfederatedrole/awsfederatedrole_controller.go
@@ -115,6 +115,11 @@ func (r *ReconcileAWSFederatedRole) Reconcile(request reconcile.Request) (reconc
 	// Build custom policy in AWS-valid JSON and converts to string
 	jsonPolicy, err := utils.MarshalIAMPolicy(*instance)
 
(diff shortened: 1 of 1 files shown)


pkg/controller/awsfederatedrole/awsfederatedrole_controller.go
```diff
@@ -115,6 +115,11 @@ func (r *ReconcileAWSFederatedRole) Reconcile(request reconcile.Request) (reconc
 	// Build custom policy in AWS-valid JSON and converts to string
 	jsonPolicy, err := utils.MarshalIAMPolicy(*instance)
 
+	// If AWSCustomPolicy and AWSManagedPolicies don't exist, exit reconcile
+	if len(instance.Spec.AWSManagedPolicies) == 0 && instance.Spec.AWSCustomPolicy.Name == "" {
+		return reconcile.Result{}, nil
```
jharrington22: Should we error instead here or at least log a message? cc/ @fahlmant

sam-nguyen7: update condition and log err

PR #190, opened 2019-11-01
CreateOrUpdate logic for Self-service IAM Roles and Policies

fahlmant:
Right now, the controller blindly tries to `Create` both IAM Roles and Policies. However, if they already exist in the AWS account, and the CR has been updated with new permissions, they will not be applied as the `Create` will fail with `EntityAlreadyExists`. This PR adds `CreateOrUpdate` logic to delete and recreate Roles and Policies when they already exist in an AWS account to ensure they are always up to date.

### DIFF
--- a/pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
+++ b/pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/iam"
 	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
 	"github.com/openshift/aws-account-operator/pkg/awsclient"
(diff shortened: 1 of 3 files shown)


pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -254,19 +254,19 @@ func (r *ReconcileAWSFederatedAccountAccess) createIAMPolicy(awsClient awsclient
 	// Marshal policydoc to json
 	jsonPolicyDoc, err := json.Marshal(&policyDoc)
 	if err != nil {
-		return fmt.Errorf("Error marshalling jsonPolicy doc : Error %s", err.Error())
+		return nil, fmt.Errorf("Error marshalling jsonPolicy doc : Error %s", err.Error())
```
rafael-azevedo: the return here should be &*iam.Policy{} instead of nil

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -309,6 +309,41 @@ func (r *ReconcileAWSFederatedAccountAccess) createIAMRole(awsClient awsclient.C
 	return nil
 }
 
+func (r *ReconcileAWSFederatedAccountAccess) createOrUpdateIAMPolicy(awsClient awsclient.Client, afr awsv1alpha1.AWSFederatedRole, afaa awsv1alpha1.AWSFederatedAccountAccess) error {
```
rafael-azevedo: does this function compile. Since you do not use the policy variable you can make it a '_'

fahlmant:
@rafael-azevedo @jharrington22 Updated

PR #194, opened 2019-11-06
Refactor aws errors

liamwazherealso:
Create LogAwsError function in utils to consolidate code.

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -427,14 +427,7 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 
 		if err != nil {
 			var returnErr error
-			if aerr, ok := err.(awserr.Error); ok {
-				reqLogger.Error(returnErr,
-					fmt.Sprintf(`Unable to create AWS connection with SRE credentials, 
-						AWS Error Code: %s, 
-						AWS Error Message: %s`,
-						aerr.Code(),
-						aerr.Message()))
-			}
+			controllerutils.LogAwsError(reqLogger, "Unable to create AWS connection with SRE credentials", returnErr, err)
 			return reconcile.Result{}, err
 		}
 
@@ -618,13 +611,7 @@ func CreateAccount(reqLogger logr.Logger, client awsclient.Client, accountName, 
 				returnErr = ErrAwsTooManyRequests
 			default:
 				returnErr = ErrAwsFailedCreateAccount
-				// Log AWS error
-				reqLogger.Error(aerr,
-					fmt.Sprintf(`New AWS Error during account creation, 
-						AWS Error Code: %s, 
-						AWS Error Message: %s`,
-						aerr.Code(),
-						aerr.Message()))
+				controllerutils.LogAwsError(reqLogger, "New AWS Error during account creation", returnErr, err)
 			}
 
 		}
@@ -686,24 +673,12 @@ func CreateIAMUser(reqLogger logr.Logger, client awsclient.Client, userName stri
 					UserName: aws.String(userName),
 				})
 				if err != nil {
-					// Log new AWS error
-					reqLogger.Error(aerr,
-						fmt.Sprintf(`New AWS Error during creation of IAM user (ErrCodeNoSuchEntityException), 
-							AWS Error Code: %s, 
-							AWS Error Message: %s`,
-							aerr.Code(),
-							aerr.Message()))
+					controllerutils.LogAwsError(reqLogger, "New AWS Error during creation of IAM user (ErrCodeNoSuchEntityException)", nil, err)
 					return &iam.CreateUserOutput{}, err
 				}
 				createUserOutput = createResult
 			default:
-				// Log new AWS error
-				reqLogger.Error(aerr,
-					fmt.Sprintf(`New AWS Error during creation of IAM user (not ErrCodeNoSuchEntityException), 
-						AWS Error Code: %s, 
-						AWS Error Message: %s`,
-						aerr.Code(),
-						aerr.Message()))
+				controllerutils.LogAwsError(reqLogger, "New AWS Error during creation of IAM user (not ErrCodeNoSuchEntityException)", nil, err)
 				return &iam.CreateUserOutput{}, err
 			}
 		} else {
@@ -730,14 +705,7 @@ func AttachAdminUserPolicy(reqLogger logr.Logger, client awsclient.Client, userN
 		}
 	}
 	if err != nil {
-		if aerr, ok := err.(awserr.Error); ok {
-			reqLogger.Error(aerr,
-				fmt.Sprintf(`New AWS Error while attaching admin user policy, 
-					AWS Error Code: %s, 
-					AWS Error Message: %s`,
-					aerr.Code(),
-					aerr.Message()))
-		}
+		controllerutils.LogAwsError(reqLogger, "New AWS Error while attaching admin user policy", nil, err)
 		return &iam.AttachUserPolicyOutput{}, err
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/controller/utils/utils.go
```diff
@@ -71,3 +75,22 @@ func AddFinalizer(object metav1.Object, finalizer string) {
 	finalizers.Insert(finalizer)
 	object.SetFinalizers(finalizers.List())
 }
+
+// LogAwsError formats and logs aws error and returns if err was an awserr
+func LogAwsError(logger logr.Logger, errMsg string, returnError error, err error) bool {
```
fahlmant: Does this need to return a bool?

pkg/controller/utils/utils.go
```diff
@@ -71,3 +75,22 @@ func AddFinalizer(object metav1.Object, finalizer string) {
...
+	if aerr, ok := err.(awserr.Error); ok {
+		if returnError == nil {
```
fahlmant: Why is `returnError` passed in? It's being set to `aerr` which could be directly used in `logger.Error` and is not passed in via pointer so it won't propagate up when `LogAwsError` returns. Can you clarify what this is used for?

pkg/controller/account/account_controller.go
```diff
@@ -428,14 +428,8 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
-			if aerr, ok := err.(awserr.Error); ok {
-				reqLogger.Error(returnErr,
-					fmt.Sprintf(`Unable to create AWS connection with SRE credentials, 
-						AWS Error Code: %s, 
-						AWS Error Message: %s`,
-						aerr.Code(),
-						aerr.Message()))
-			}
+			errMsg := `Unable to create AWS connection with SRE credentials`
+			controllerutils.LogAwsError(reqLogger, errMsg, returnErr, err)
```
fahlmant: I think this would be better to collapse into
controllerutils.LogAwsError(reqLogger, "Unable to create AWS connection with SRE credentials", returnErr, err)

pkg/controller/account/account_controller.go
```diff
@@ -428,14 +428,8 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
 			var returnErr error
-			if aerr, ok := err.(awserr.Error); ok {
-				reqLogger.Error(returnErr,
-					fmt.Sprintf(`Unable to create AWS connection with SRE credentials, 
-						AWS Error Code: %s, 
-						AWS Error Message: %s`,
-						aerr.Code(),
-						aerr.Message()))
-			}
+			errMsg := `Unable to create AWS connection with SRE credentials`
```
fahlmant: Should be `"` Any particular reason to use backticks?

pkg/controller/account/account_controller.go
```diff
@@ -619,13 +613,8 @@ func CreateAccount(reqLogger logr.Logger, client awsclient.Client, accountName,
...
 				returnErr = ErrAwsFailedCreateAccount
-				// Log AWS error
-				reqLogger.Error(aerr,
-					fmt.Sprintf(`New AWS Error during account creation, 
-						AWS Error Code: %s, 
-						AWS Error Message: %s`,
-						aerr.Code(),
-						aerr.Message()))
+				errMsg := `New AWS Error during account creation`
+				controllerutils.LogAwsError(reqLogger, errMsg, returnErr, err)
```
fahlmant: Same comment for backticks vs `"` and please collapse.

pkg/controller/account/account_controller.go
```diff
@@ -687,24 +676,14 @@ func CreateIAMUser(reqLogger logr.Logger, client awsclient.Client, userName stri
...
 				})
 				if err != nil {
-					// Log new AWS error
-					reqLogger.Error(aerr,
-						fmt.Sprintf(`New AWS Error during creation of IAM user (ErrCodeNoSuchEntityException), 
-							AWS Error Code: %s, 
-							AWS Error Message: %s`,
-							aerr.Code(),
-							aerr.Message()))
+					errMsg := `New AWS Error during creation of IAM user (ErrCodeNoSuchEntityException)`
```
fahlmant: Same as above.

pkg/controller/account/account_controller.go
```diff
@@ -687,24 +676,14 @@ func CreateIAMUser(reqLogger logr.Logger, client awsclient.Client, userName stri
...
 				createUserOutput = createResult
 			default:
-				// Log new AWS error
-				reqLogger.Error(aerr,
-					fmt.Sprintf(`New AWS Error during creation of IAM user (not ErrCodeNoSuchEntityException), 
-						AWS Error Code: %s, 
-						AWS Error Message: %s`,
-						aerr.Code(),
-						aerr.Message()))
+				errMsg := `New AWS Error during creation of IAM user (not ErrCodeNoSuchEntityException)`
```
fahlmant: Same as above.

pkg/controller/account/account_controller.go
```diff
@@ -731,14 +710,8 @@ func AttachAdminUserPolicy(reqLogger logr.Logger, client awsclient.Client, userN
...
 	if err != nil {
-		if aerr, ok := err.(awserr.Error); ok {
-			reqLogger.Error(aerr,
-				fmt.Sprintf(`New AWS Error while attaching admin user policy, 
-					AWS Error Code: %s, 
-					AWS Error Message: %s`,
-					aerr.Code(),
-					aerr.Message()))
-		}
+		errMsg := `New AWS Error while attaching admin user policy`
```
fahlmant: Same as above.

pkg/controller/account/account_controller.go
```diff
@@ -753,16 +726,9 @@ func CreateUserAccessKey(reqLogger logr.Logger, client awsclient.Client, userNam
...
 	if err != nil {
-		if aerr, ok := err.(awserr.Error); ok {
-			reqLogger.Error(aerr,
-				fmt.Sprintf(`New AWS Error while creating user access key, 
-					AWS Error Code: %s, 
-					AWS Error Message: %s`,
-					aerr.Code(),
-					aerr.Message()))
-		}
+		errMsg := `New AWS Error while creating user access key`
```
fahlmant: Same as above.

pkg/controller/utils/utils.go
liamwazherealso: This statement is so that if one would like to return an error that is neither the `err` or `aerr` in the call to `logger.Error` they may. For example if you would like change the value based on a switch in the proceeding code. 

Places where this occurs
- [L160](https://github.com/openshift/aws-account-operator/pull/194/files#diff-f899a0fa089ce8f0b7744fc82ee53558R160)
-[L927](https://github.com/openshift/aws-account-operator/pull/194/files#diff-
32f12988d73ded1726ea7c67c8cb1f90R927)
-[L859[](https://github.com/openshift/aws-account-operator/pull/194/files#diff-32f12988d73ded1726ea7c67c8cb1f90R895)

pkg/controller/utils/utils.go
liamwazherealso: For use cases like [this](https://github.com/openshift/aws-account-operator/pull/194/files#diff-26f5cba4465c7a85a44a0dcf6ac15793R183) it's helpful.

fahlmant: But this logic is already handled in this `LogAwsError` function. The example you gave was written because not error checking is really done, but in line 81, you do this exact thing.

pkg/controller/account/account_controller.go
liamwazherealso: Nope will change

pkg/controller/utils/utils.go
liamwazherealso: If this function doesn't return a bool. Then this specific case can't be covered.

Because the other solutions here are to change [this one call](https://github.com/openshift/aws-account-operator/pull/194/files#diff-26f5cba4465c7a85a44a0dcf6ac15793R183) to not handle the case where `err.(awserr.Error)` fails. Or call the function the clause after it passes. Which is a redundant check.

pkg/controller/account/account_controller.go
liamwazherealso: Done

pkg/controller/utils/utils.go
fahlmant: All you would need to do for that case is fix this:
`if utils.LogAwsError(log, "", err, err) != true {`
to
`utils.LogAwsError(log, "Error listing managed AWS policies", err, err)`

liamwazherealso: Ack

liamwazherealso:
> Got to repetitive to comment on all the instances. But please use `"` instead of backticks, and collapse the errMessage var into the function call.

https://github.com/openshift/aws-account-operator/pull/194/commits/e641a99e45df05cd4719fa8596fc78f8d420d780 fixed in this commit

liamwazherealso:
Done

> You missed collapsing the error msg in `ec2.go`

Done

PR #195, opened 2019-11-06
Initialization of region verification checks unique instance

liamwazherealso:
https://jira.coreos.com/browse/SREP-2154

Currently initialization of regions may fail due to multiple instance since the specific instanceId is not checked.


liamwazherealso:
Tested two ways.

PR #199, opened 2019-11-08
[WIP] Remove initialize totalaccountwatcher's update call

liamwazherealso:
https://jira.coreos.com/browse/SREP-2369

Currently the initialization of totalaccountwatcher calls the update total function. This function often fails (as the aws call it wraps often does). Which will result in unnecessary failed starts. So the update call within initialize is being removed. The update total function is still called within a goroutine.


jharrington22:
Closing see #232

jharrington22:
Closing see #232

PR #203, opened 2019-11-13
Add additional printer column annotations

jharrington22:
This PR adds annotations so that the additional printer columns are generated with openapi


deploy/crds/aws_v1alpha1_account_crd.yaml
```diff
@@ -57,7 +61,6 @@ spec:
             iamUserSecret:
               type: string
             legalEntity:
-              description: 'Legal entity info populated at claim time, carried over from account claim'
```
fahlmant: Do you want these removed?

PR #209, opened 2019-11-20
Set ClaimLinkNamespace when Creating an Account CR

fahlmant:
`ClaimLinkNamespace` should be set to blank when a new account CR is generated by the AccountPool.


fahlmant:
@jharrington22 It already does
https://github.com/openshift/aws-account-operator/blob/master/pkg/controller/accountclaim/accountclaim_controller.go#L430

PR #211, opened 2019-11-22
Automatically generate CRDs in CSV

jharrington22:
Currently CRDs are not added into the CSV, this automates that process.

### DIFF
--- a/deploy/cluster_role.yaml
+++ b/deploy/cluster_role.yaml
@@ -45,6 +45,8 @@ rules:
   - accountclaims
   - accounts
   - accountpools
+  - awsfederatedaccountaccesses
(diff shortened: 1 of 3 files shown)


deploy/cluster_role.yaml
```diff
@@ -45,6 +45,8 @@ rules:
   - accountclaims
   - accounts
   - accountpools
+  - awsfederatedaccountaccesses
```
fahlmant: Do we want to keep the * on Line 44?

jharrington22: I think we chose to keep it last time but be also be specific, I'm happy to remove it but in a separate PR in case we break something.

PR #218, opened 2019-11-27
Handle error for BYOC aws client setup

fahlmant:
This PR ensures that if the BYOC AWS client does not get initialized correctly, the error is returned.


jharrington22:
cc/ @jmelis @meowfaceman @rogbas fixes for hive-int and staging

PR #223, opened 2019-12-05
Refactor BuildIAMUser and related functions

fahlmant:
BuildIAMUser in the account controller fails if there are 2 access keys for a given user, or a secret for that user/key already exists. This PR refactors BuildIAMUser to do the following:
1) If user already exists, do not error
2) If the secret does not exist, deletes all current access keys, creates a new one, and creates a new secret with the access key.
3) If the secret does exist, assumes the key in it is usable

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -16,6 +16,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/aws/aws-sdk-go/service/support"
 	"github.com/go-logr/logr"
+	"github.com/openshift/aws-account-operator/pkg/controller/utils"
 	corev1 "k8s.io/api/core/v1"
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -689,14 +690,32 @@ func (r *ReconcileAccount) BuildIAMUser(reqLogger logr.Logger, awsClient awsclie
 	_, userErr := CreateIAMUser(reqLogger, awsClient, iamUserName)
 	// TODO: better error handling but for now scrap account
 	if userErr != nil {
-		failedToCreateIAMUserMsg := fmt.Sprintf("Failed to create IAM user %s", iamUserName)
-		SetAccountStatus(reqLogger, account, failedToCreateIAMUserMsg, awsv1alpha1.AccountFailed, "Failed")
-		err := r.Client.Status().Update(context.TODO(), account)
-		if err != nil {
-			return "", err
+		//If the user already exists, don't erro
+		if aerr, ok := userErr.(awserr.Error); ok {
+			switch aerr.Code() {
+			case "EntityAlreadyExists":
+				break
+			default:
+				failedToCreateIAMUserMsg := fmt.Sprintf("Failed to create IAM user %s due to AWS Error: %s", iamUserName, aerr.Message())
+				utils.LogAwsError(reqLogger, "", nil, userErr)
+				SetAccountStatus(reqLogger, account, failedToCreateIAMUserMsg, awsv1alpha1.AccountFailed, "Failed")
+				err := r.Client.Status().Update(context.TODO(), account)
+				if err != nil {
+					return "", err
+				}
+				reqLogger.Info(failedToCreateIAMUserMsg)
+				return "", userErr
+			}
+		} else {
+			failedToCreateIAMUserMsg := fmt.Sprintf("Failed to create IAM user %s due to non-AWS Error", iamUserName)
+			SetAccountStatus(reqLogger, account, failedToCreateIAMUserMsg, awsv1alpha1.AccountFailed, "Failed")
+			err := r.Client.Status().Update(context.TODO(), account)
+			if err != nil {
+				return "", err
+			}
+			reqLogger.Info(failedToCreateIAMUserMsg)
+			return "", userErr
 		}
-		reqLogger.Info(failedToCreateIAMUserMsg)
-		return "", userErr
 	}
 
 	reqLogger.Info(fmt.Sprintf("Attaching Admin Policy to IAM user %s", iamUserName))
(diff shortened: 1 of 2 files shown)


pkg/controller/account/iam.go
```diff
@@ -355,3 +357,73 @@ func getSigninToken(reqLogger logr.Logger, federationEndpointURL string, federat
...
+				}
+				deleteAccessKey(reqLogger, client, userName, *accessKeyList.AccessKeyMetadata[0].AccessKeyId)
+			}
+		}
+	}
+
+	return output, nil
+}
+
+func deleteAccessKey(reqLogger logr.Logger, client awsclient.Client, userName string, accessKeyID string) error {
```
fahlmant: Maybe "smartDeleteAccessKeys" to differentiate from a direct delete?

pkg/controller/account/iam.go
```diff
@@ -355,3 +357,73 @@ func getSigninToken(reqLogger logr.Logger, federationEndpointURL string, federat
...
+			return &iam.CreateAccessKeyOutput{}, err
+		}
+	} else {
+		iamSecret := &corev1.Secret{}
+		err := r.Client.Get(context.TODO(), types.NamespacedName{Name: account.Spec.IAMUserSecret, Namespace: awsv1alpha1.AccountCrNamespace}, iamSecret)
+		if err != nil {
+			return &iam.CreateAccessKeyOutput{}, err
+		}
+
+		accessKeyID := string(iamSecret.Data["aws_access_key_id"])
```
rafael-azevedo: You should make sure this value exists so you do not get a panic 

```

if val, ok := dict["foo"]; ok {
    //do something here
}

```

fahlmant:
/hold
To avoid accidental merges

pkg/controller/account/iam.go
```diff
@@ -358,3 +361,79 @@ func getSigninToken(reqLogger logr.Logger, federationEndpointURL string, federat
 	return signinResponse, nil
 
 }
+
+func (r *ReconcileAccount) smartCreateAccessKey(reqLogger logr.Logger, client awsclient.Client, userName string, account *awsv1alpha1.Account) (*iam.CreateAccessKeyOutput, error) {
```
jharrington22: Could we rename this function `upsertAccessKey` I feel that makes sense here since its more descriptive of whats going to happen, update or insert (create)

pkg/controller/account/iam.go
```diff
@@ -358,3 +361,79 @@ func getSigninToken(reqLogger logr.Logger, federationEndpointURL string, federat
...
+				}
+				smartDeleteAccessKey(reqLogger, client, userName, *accessKeyList.AccessKeyMetadata[0].AccessKeyId)
+			}
+		}
+	}
+
+	return output, nil
+}
+
+func smartDeleteAccessKey(reqLogger logr.Logger, client awsclient.Client, userName string, accessKeyID string) error {
```
jharrington22: This function should be just `deleteAccessKeys`

fahlmant: It deletes the other access key than given. Don't remember why I exactly did that though.

pkg/controller/account/iam.go
```diff
@@ -358,3 +361,79 @@ func getSigninToken(reqLogger logr.Logger, federationEndpointURL string, federat
...
+
+	accessKeyList, err := client.ListAccessKeys(&iam.ListAccessKeysInput{})
+	if err != nil {
+		return err
+	}
+
+	for _, accessKey := range accessKeyList.AccessKeyMetadata {
+		if *accessKey.AccessKeyId != accessKeyID {
+			_, err = client.DeleteAccessKey(&iam.DeleteAccessKeyInput{AccessKeyId: accessKey.AccessKeyId})
```
jharrington22: Can we case the AWS error here too

pkg/controller/account/iam.go
```diff
@@ -358,3 +361,29 @@ func getSigninToken(reqLogger logr.Logger, federationEndpointURL string, federat
...
+				_, err = client.DeleteAccessKey(&iam.DeleteAccessKeyInput{AccessKeyId: accessKeyList.AccessKeyMetadata[0].AccessKeyId})
+				if err != nil {
+					return nil, err
+				}
+				output, err = CreateUserAccessKey(reqLogger, client, userName)
+				if err != nil {
+					return nil, err
+				}
+			}
+		}
```
ArielLima: If it is not an AWS error shouldn't we return it?

pkg/controller/account/account_controller.go
```diff
@@ -607,14 +609,29 @@ func (r *ReconcileAccount) BuildIAMUser(reqLogger logr.Logger, awsClient awsclie
...
 	// TODO: better error handling but for now scrap account
 	if userErr != nil {
-		failedToCreateIAMUserMsg := fmt.Sprintf("Failed to create IAM user %s", iamUserName)
-		SetAccountStatus(reqLogger, account, failedToCreateIAMUserMsg, awsv1alpha1.AccountFailed, "Failed")
-		err := r.Client.Status().Update(context.TODO(), account)
-		if err != nil {
-			return "", err
+		if aerr, ok := userErr.(awserr.Error); ok {
+			//If the user already exists, don't error
+			if aerr.Code() != "EntityAlreadyExists" {
```
jharrington22: Other places we are using a case statement here where the below logic would go in `default:` could you update to match?

fahlmant: :+1:

pkg/controller/account/account_controller.go
```diff
@@ -730,26 +737,66 @@ func (r *ReconcileAccount) BuildIAMUser(reqLogger logr.Logger, awsClient awsclie
...
-		SetAccountStatus(reqLogger, account, failedToCreateUserSecretMsg, awsv1alpha1.AccountFailed, "Failed")
-		err := r.Client.Status().Update(context.TODO(), account)
-		if err != nil {
+	// create user secrets if needed
+	userExistingSecret := &corev1.Secret{}
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: fmt.Sprintf("%s-secret", secretName), Namespace: nameSpace}, userExistingSecret)
+	if err != nil {
+		if !k8serr.IsNotFound(err) {
+			getSecretErr := fmt.Sprintf("Error getting secret for user %s, err: %+v", iamUserName, err)
+			reqLogger.Info(getSecretErr)
```
jharrington22: This should be an error?

PR #224, opened 2019-12-06
FAR better error handling

liamwazherealso:
https://jira.coreos.com/browse/SREP-2501

The FederatedAccountRole controller currently doesn't handle an "EntityAlreadyExists" aws error in the proper way and will err and reconcile when handling this err. This PR changes the FAR controller to handle this error in the same way as the FAA controller.

### DIFF
--- a/pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
+++ b/pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
@@ -147,8 +147,18 @@ func (r *ReconcileAWSFederatedAccountAccess) Reconcile(request reconcile.Request
 		return reconcile.Result{}, err
 	}
 
+	//TODO() will be refactored in Rafs pr to use secret
+	// create policy arn
+	accountCR := &awsv1alpha1.Account{}
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: currentFAA.Spec.AccountReference, Namespace: awsv1alpha1.AccountCrNamespace}, accountCR)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	accountID := accountCR.Spec.AwsAccountID
+	policyARN := fmt.Sprintf("arn:aws:iam::%s:policy/%s", accountID, roleMap[currentFAA.Spec.AWSFederatedRoleName].Spec.AWSCustomPolicy.Name)
+
 	// Here create the custom policy in the cluster account
-	err = r.createOrUpdateIAMPolicy(awsClient, roleMap[currentFAA.Spec.AWSFederatedRoleName], *currentFAA)
+	err = CreateOrUpdateIAMPolicy(awsClient, roleMap[currentFAA.Spec.AWSFederatedRoleName], policyARN)
 	if err != nil {
 		// if we were unable to create the policy fail this CR.
 		SetStatuswithCondition(currentFAA, "Failed to create custom policy", awsv1alpha1.AWSFederatedAccountFailed, awsv1alpha1.AWSFederatedAccountStateFailed)
@@ -178,13 +188,6 @@ func (r *ReconcileAWSFederatedAccountAccess) Reconcile(request reconcile.Request
 		return reconcile.Result{}, nil
 	}
 
-	accountCR := &awsv1alpha1.Account{}
-	err = r.client.Get(context.TODO(), types.NamespacedName{Name: currentFAA.Spec.AccountReference, Namespace: awsv1alpha1.AccountCrNamespace}, accountCR)
-	if err != nil {
-		return reconcile.Result{}, err
-	}
-
-	accountID := accountCR.Spec.AwsAccountID
 	// Add requested aws managed policies to the role
 	awsManagedPolicyNames := []string{}
 	// Add all aws managed policy names to a array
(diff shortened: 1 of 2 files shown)


pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -309,23 +309,29 @@ func (r *ReconcileAWSFederatedAccountAccess) createIAMRole(awsClient awsclient.C
...
+
+	policy, err := createIAMPolicy(awsClient, afr)
 	if err != nil {
 		if aerr, ok := err.(awserr.Error); ok {
 			if aerr.Code() == "EntityAlreadyExists" {
 
 				// If the Role already exists, delete it and recreate it
 				_, err = awsClient.DeletePolicy(&iam.DeletePolicyInput{PolicyArn: policy.Arn})
-				_, err = r.createIAMPolicy(awsClient, afr, afaa)
+				_, err = createIAMPolicy(awsClient, afr)
```
rafael-azevedo: These errors need to be handled properly or passed forward properly to the next func.

pkg/controller/awsfederatedrole/awsfederatedrole_controller.go
```diff
@@ -141,6 +141,9 @@ func (r *ReconcileAWSFederatedRole) Reconcile(request reconcile.Request) (reconc
 		PolicyName:     &instance.Spec.AWSCustomPolicy.Name,
 		PolicyDocument: &jsonPolicy,
 	})
+
+	err = awsfaa.CreateOrUpdateIAMPolicy(awsClient, *instance)
+
 	if err != nil {
 		if aerr, ok := err.(awserr.Error); ok {
 			if aerr.Code() == "MalformedPolicyDocument" {
```
rafael-azevedo: These functions errors here need to be refactored as well. since you are now not returning an aws error possibly from CreateOrUpdateIAMPolicy we need better handling here for instance 164 still does not update the status of the FAR

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -309,25 +309,37 @@ func (r *ReconcileAWSFederatedAccountAccess) createIAMRole(awsClient awsclient.C
...
 				// If the Role already exists, delete it and recreate it
 				_, err = awsClient.DeletePolicy(&iam.DeletePolicyInput{PolicyArn: policy.Arn})
-				_, err = r.createIAMPolicy(awsClient, afr, afaa)
+				if err != nil {
+					return err
+				}
+				_, err = createIAMPolicy(awsClient, afr)
+				if err != nil {
+					return err
+				}
```
rafael-azevedo: after this error check we need a return nil since it was able to do both the delete and create

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -147,8 +147,18 @@ func (r *ReconcileAWSFederatedAccountAccess) Reconcile(request reconcile.Request
 		return reconcile.Result{}, err
 	}
 
+	//TODO() will be refactored in Rafs pr to use secret
+	// create policy arn
+	accountCR := &awsv1alpha1.Account{}
```
fahlmant: What are these changes for? I don't see anything in the PR description about changing the account access controller

pkg/controller/awsfederatedrole/awsfederatedrole_controller.go
```diff
@@ -135,45 +136,45 @@ func (r *ReconcileAWSFederatedRole) Reconcile(request reconcile.Request) (reconc
...
 	}
 
+	// policy arn template
+	policyARN := fmt.Sprintf("arn:aws:iam::%s:policy/%s", *gciOut.Account, instance.Spec.AWSCustomPolicy.Name)
+
+	// Validates Custom IAM Policy
+	log.Info("Validating Custom Policies")
+	err = awsfaa.CreateOrUpdateIAMPolicy(awsClient, *instance, policyARN)
+	// Invalidates instance if unable not create policy
+	if err != nil {
```
fahlmant: Why do we no longer handle the AWS error? We need to unwrap it and log it somewhere like every other AWS error in the operator

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -309,25 +312,34 @@ func (r *ReconcileAWSFederatedAccountAccess) createIAMRole(awsClient awsclient.C
 	return nil
 }
 
-func (r *ReconcileAWSFederatedAccountAccess) createOrUpdateIAMPolicy(awsClient awsclient.Client, afr awsv1alpha1.AWSFederatedRole, afaa awsv1alpha1.AWSFederatedAccountAccess) error {
-
-	policy, err := r.createIAMPolicy(awsClient, afr, afaa)
+// TODO() this will need to be moved to a different package because it's shared across controller
+func CreateOrUpdateIAMPolicy(awsClient awsclient.Client, afr awsv1alpha1.AWSFederatedRole, policyARN string) error {
```
fahlmant: This should be moved to utils to avoid importing this whole controller to the federated role controller

PR #226, opened 2019-12-10
Add clusterrolebinding to oc template

jharrington22:
This PR adds the clusterrolebinding for the aws-acconut-operator-client ServiceAccount to the oc template that is applied on deployment.

Currently that template is not generated automatically as part of the deployment and probably should be if we are going to include things such as this clusterrolebinding. At the moment I don't see an easier way to apply the additional subject to the clusterrolebinding which is managed by OLM and at this time OLM does not support modifying or customizing the ciusterrolebinding it creates for service accounts listed in the CSV.

This clusterrolebinding is not going to have the same name as the one created by OLM but that shouldn't be an issue for the moment.

cc/ @maorfr @jewzaam @rogbas

### DIFF
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -39,3 +39,48 @@ objects:
     name: aws-account-operator
     source: aws-account-operator-catalog
     sourceNamespace: aws-account-operator
+
+- apiVersion: rbac.authorization.k8s.io/v1
+  kind: ClusterRoleBinding
+  metadata:
+    name: aws-account-operator-client
+  subjects:
+  - kind: ServiceAccount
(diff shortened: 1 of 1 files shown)


hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -38,4 +38,19 @@ objects:
...
+    name: aws-account-operator-client
+  subjects:
+  - kind: ServiceAccount
+    namespace: aws-account-operator-client
+    name: aws-account-operator-client
+  - kind: Group
+    name: aws-account-operator-client
+  roleRef:
+    kind: ClusterRole
+    name: aws-account-operator-client
```
rogbas: Where is this ClusterRole being created? it doesn't exist in hive prod currently

PR #227, opened 2019-12-10
Reorder clusterrole and clusterrolebinding

jharrington22:
The template will create resources in order so the role needs to be created before the binding can use it


jharrington22:
/hold

Putting a hold as the user that applies this role doesn't have the correct permissions

PR #229, opened 2019-12-11
Ensure we update the status on the accountclaim

jharrington22:
This PR fixes a few bugs with BYOC and also attempts to handle inconsistencies of the AWS API when working with STS tokens

There are two main problems its fixing:

- The AccountClaims status was never updated to Ready (For a few reasons)
- The AWS API seems to be returning "InvalidClientTokenId" or "AccessDenied" inconsistently with the STS tokens, I've implemented some retries.

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -745,38 +745,98 @@ func CreateAccount(reqLogger logr.Logger, client awsclient.Client, accountName, 
 	return accountStatus, nil
 }
 
+func checkIAMUserExists(reqLogger logr.Logger, client awsclient.Client, userName string) (bool, error) {
+	// Retry when getting IAM user information
+	// Sometimes we see a delay before credentials are ready to be user resulting in the AWS API returning 404's
+	attempt := 1
+	for i := 0; i < 10; i++ {
+		// check if username exists for this account
+		_, err := client.GetUser(&iam.GetUserInput{
+			UserName: aws.String(userName),
+		})
+
+		attempt++
+		// handle errors
+		if err != nil {
+			if aerr, ok := err.(awserr.Error); ok {
+				switch aerr.Code() {
+				case iam.ErrCodeNoSuchEntityException:
+					return false, nil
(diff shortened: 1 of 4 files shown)


pkg/controller/account/account_controller.go
```diff
@@ -745,38 +745,96 @@ func CreateAccount(reqLogger logr.Logger, client awsclient.Client, accountName,
...
+					}
+				default:
+					controllerutils.LogAwsError(reqLogger, "Unexpected AWS Error when checking IAM user exists", nil, err)
+					return false, err
+				}
+			} else {
+				return false, fmt.Errorf("Unable to check if user %s exists error: %s", userName, err)
+			}
+		}
+		time.Sleep(time.Duration(time.Duration(attempt*5) * time.Second))
```
rogbas: Should this sleep be inside the previous block? The way it is it will sleep even when `err == nil`

pkg/controller/account/account_controller.go
```diff
@@ -745,38 +745,96 @@ func CreateAccount(reqLogger logr.Logger, client awsclient.Client, accountName,
...
-
-			default:
-				controllerutils.LogAwsError(reqLogger, "New AWS Error during creation of IAM user (not ErrCodeNoSuchEntityException)", nil, err)
-				return &iam.CreateUserOutput{}, err
 			}
 		} else {
 			return &iam.CreateUserOutput{}, err
 		}
+
+		time.Sleep(time.Duration(time.Duration(attempt*5) * time.Second))
```
rogbas: Same thing from above, shouldn't this sleep be inside the `err != nill` block?

pkg/controller/account/account_controller.go
wshearn: Also err is never updated, so it will run the switch case with the original value failing every time.

pkg/controller/account/account_controller.go
wshearn: err is also not being updated here, so it will still keep the same value

PR #232, opened 2019-12-18
Removed call to aws from totalaccountwatcher initilize

rafael-azevedo:
On operator start we check that total number of accounts created in the AWS organization. This process holds the operator from starting until this call completes, this API call to AWS often takes 30 seconds to > 1 minute we should move it to a go routine.

This PR removes update total call from initialize function for totalaccountwatcher


jharrington22:
We are going to close this PR after running through the logic we've realised that if we don't initalize the the number of accounts on initialization there would be a period where the account operator could create accounts unbound until its checked. Closing.

PR #233, opened 2019-12-18
Added env variable check when developing locally

rafael-azevedo:
This PR is a suggestion to have a env variable checked when running that forces the operator to skip metrics configuration. This allows the user to set `export FORCE_DEV_MODE=local` and it skips metrics configuration. This allows the user to test development without having to comment out code which is less error prone while testing.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -29,12 +29,15 @@ import (
 
 // Change below variables to serve metrics on different host or port.
 var (
-	metricsPort                   = "8080"
-	metricsPath                   = "/metrics"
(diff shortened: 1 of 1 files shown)


iamkirkbater:
@clcollins and I were just talking about doing something like this yesterday.

We should also consider adding the same for the other places where we need to update the code in order to run this locally, using the same env var.

rogbas:
@iamkirkbater can you please close that PR after you merge/squash your identical PR?

iamkirkbater:
Closed in favor of the #284 which built off of this.

PR #236, opened 2019-12-19
Initialize account when creating, use spec AWS account ID

jharrington22:
This PR fixes an issue where the account CR enters the "Creating" status and we never attempt to initialize the account where eventually it would set to "Ready". Since we don't ever initialize the account we assume this account is stuck "Creating" and fail it causing an AWS account leak. No account could ever become "Ready" in this state.

I also remove references to `awsAccountID`  where not required and only initialize it inside the condition where it is used and not at the pkg scope.


jharrington22:
Tested
- BYOC
- Account reuse
- Account creation

PR #238, opened 2019-12-19
Only build BYOC aws client if needed

fahlmant:
Currently, every time a BYOC account is reconciled, an aws client is build based on the credentials in the secret referenced in the CR. This client is only used when first initializing the BYOC account and is never used again. This PR ensures that we only build the client when needed.

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -285,57 +285,57 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 	// If the account is BYOC, needs some different set up
 	if currentAcctInstance.Spec.BYOC {
 		reqLogger.Info("BYOC account")
-
-		// Create client for BYOC account
-		byocAWSClient, accountClaim, err := r.getBYOCClient(currentAcctInstance)
-		if err != nil {
-			if accountClaim != nil {
-				r.accountClaimBYOCError(reqLogger, accountClaim, err)
-			}
-			return reconcile.Result{}, err
-		}
-
-		err = validateBYOCClaim(accountClaim)
-		if err != nil {
-			r.accountClaimBYOCError(reqLogger, accountClaim, err)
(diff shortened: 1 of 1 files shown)


pkg/controller/account/account_controller.go
```diff
@@ -283,7 +283,7 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 	}
 
 	// If the account is BYOC, needs some different set up
-	if currentAcctInstance.Spec.BYOC {
+	if currentAcctInstance.Spec.BYOC && (currentAcctInstance.Status.State == "" || currentAcctInstance.Status.Claimed != true) {
```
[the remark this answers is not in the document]
fahlmant: Fixing

pkg/controller/account/account_controller.go
```diff
@@ -285,57 +285,57 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
-			// Rotate access keys
-			err = r.byocRotateAccessKeys(reqLogger, byocAWSClient, accountClaim)
+		if currentAcctInstance.Status.State == "" || currentAcctInstance.Status.Claimed != true {
+			// Create client for BYOC account
+			byocAWSClient, accountClaim, err := r.getBYOCClient(currentAcctInstance)
 			if err != nil {
-				reqLogger.Info("Failed to rotate BYOC access keys")
-				r.accountClaimBYOCError(reqLogger, accountClaim, err)
+				if accountClaim != nil {
+					r.accountClaimBYOCError(reqLogger, accountClaim, err)
```
jharrington22: I don't think we want to error here right? If a reconcile for the AccountClaim gets triggered we may just want to ignore it. What is the reason for the error is it an issue?

fahlmant: This already exists in master today, line 292 of the account_controller. The only change in this PR is that its within a new if statement above.

PR #243, opened 2020-01-08
Add roleDisplayName for OCM user interface

jharrington22:
OCM requires a way to dynamically display a user friendly role name, along with the description we've added a display name to the spec.


jharrington22:
cc @zgalor

PR #246, opened 2020-01-13
Added credential refresh to reuse

rafael-azevedo:
This PR adds Boolean rotateConsoleCredentials and RotateCredentials to true after account clean up so that credentials are refreshed when doing an account reuse.

### DIFF
--- a/pkg/controller/accountclaim/reuse.go
+++ b/pkg/controller/accountclaim/reuse.go
@@ -44,6 +44,22 @@ func (r *ReconcileAccountClaim) finalizeAccountClaim(reqLogger logr.Logger, acco
 		return err
 	}
 
+	// Get latest version of the account
+	reusedAccount, err = r.getClaimedAccount(accountClaim.Spec.AccountLink, awsv1alpha1.AccountCrNamespace)
+	if err != nil {
+		reqLogger.Error(err, "Failed to get claimed account")
+		return err
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/controller/accountclaim/reuse.go
```diff
@@ -44,6 +44,22 @@ func (r *ReconcileAccountClaim) finalizeAccountClaim(reqLogger logr.Logger, acco
 		return err
 	}
 
+	// Get latest version of the account
+	refreshedAccount, err := r.getClaimedAccount(accountClaim.Spec.AccountLink, awsv1alpha1.AccountCrNamespace)
```
jharrington22: Do we get an error if we don't refresh the account given the above two functions are updating the account CR? Also I'm not sure we need a new var here for the Account CR.

pkg/controller/accountclaim/reuse.go
```diff
@@ -44,6 +44,22 @@ func (r *ReconcileAccountClaim) finalizeAccountClaim(reqLogger logr.Logger, acco
...
+	if err != nil {
+		reqLogger.Error(err, "Failed to get claimed account")
+		return err
+	}
+
+	reqLogger.Info("Rotating Credentials")
```
jharrington22: Can we be more descriptive here and in the error? Reading the logs it would be confusing to know where the rotation is happening and also where the status update failed.

PR #247, opened 2020-01-14
Allow OCM access to federated user resources

jharrington22:
This PR allows the OCM service account and SDA access to federated user resources


jharrington22:
cc/ @zgalor

PR #252, opened 2020-01-23
Update the aws federated account access example CR

fahlmant:
Currently, the AWS Federated Account Access example CR does not reflect the actual struct in the CRD. This PR updates the example to match the current CRD


fahlmant:
Closing in favor of #241

PR #255, opened 2020-01-24
Changes random strings to use constants

iamkirkbater:
Also edits the constants to use CONSTANT case for clarity.

Still needs to be tested locally, I'm working to set up my environment now.

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -334,7 +334,7 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 				// Update the account CR to creating
 				reqLogger.Info("Updating BYOC to creating")
 				currentAcctInstance.Status.State = AccountCreating
-				SetAccountStatus(reqLogger, currentAcctInstance, "BYOC Account Creating", awsv1alpha1.AccountCreating, "Creating")
+				SetAccountStatus(reqLogger, currentAcctInstance, "BYOC Account Creating", awsv1alpha1.AccountCreating, AccountCreating)
 				err = r.Client.Status().Update(context.TODO(), currentAcctInstance)
 				if err != nil {
 					return reconcile.Result{}, err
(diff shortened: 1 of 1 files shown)


iamkirkbater:
This has been tested locally and is now ready for review.

clcollins:
Looks good to me, but I think it suffers from the same bug @fahlmant was fixing in #259, so might need to rebase on that when it's merged?

iamkirkbater:
I originally did that in my first attempt, but ended up trying to generalize too far.  Actually, though, I have an idea that I'll run with and get back to you on this.

iamkirkbater:
@fahlmant I originally tried putting those consts in the `utils` package, but that ended up creating a circular dependency within the `v1alpha1` package when I tried using them there as well.  So I just put them in the `v1alpha1` package and then used them from there.

Let me know if you have any concerns with that, otherwise this should be good if you think it is.

jharrington22:
@iamkirkbater general feedback here, with the naming of these variables its important to remember that the style of your variables is also related to syntax, eg the first letter of a variable denotes whether or not its exported from the package. As for style Go is always camel case besides that first letter of course.

Sharing the same variable across types is probably not a good idea as we want to keep types defined and contained to themselves ie. changing v1alpha1.READY would effect more than one types eg `v1alpha1.AccountConditionType.AccountReady` and `v1alpha1.ClaimStatus.ClaimStatusReady` this wouldn't be expected. When we define a type we expect that type to be the source of truth.

pkg/controller/account/account_controller.go
```diff
@@ -462,7 +462,7 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 	}
 
 	// Account init for both BYOC and Non-BYOC
-	if (currentAcctInstance.Spec.BYOC && currentAcctInstance.Status.State != "Ready") || ((currentAcctInstance.Status.State == string(awsv1alpha1.AccountCreating)) && (currentAcctInstance.Status.Claimed == false)) {
+	if (currentAcctInstance.Spec.BYOC && currentAcctInstance.Status.State != AccountReady) || ((currentAcctInstance.Status.State == string(awsv1alpha1.AccountCreating)) && (currentAcctInstance.Status.Claimed == false)) {
```
fahlmant: I think `string(awsv1alpha1.AccountCreating)` can also be changed to `AccountCreating`

PR #259, opened 2020-01-27
Fix for create user loop

fahlmant:
If the err is nil during CreateUser, it returns an empty struct due to a misplaced else. This PR fixes the aligntment of the else to match up with handing the AWS err as originally intended. It also consolidates the attempt counter and the control counter into one variable.


fahlmant:
Closing in favor of a PR @jharrington22 will open

PR #260, opened 2020-01-28
Wait for secret to be ready for AWSClient

jharrington22:
Once the osdManagedAdminSRE user is created we immediately attempt to create and then use a secret containing the IAM credentials to connect to AWS. This secret is not available immediately, we need to wait. 

The correct way to implement this would be with a watch, I'll create a card to come back and refactor this as a watch.

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -20,6 +20,7 @@ import (
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


pkg/controller/account/account_controller.go
```diff
@@ -530,6 +531,22 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
+		// Intermittently our secret wont be ready before the next call, lets ensure it exists
+		for i := 0; i < 10; i++ {
+			secret := &corev1.Secret{}
+			err := r.Client.Get(context.TODO(), types.NamespacedName{Name: SREIAMUserSecret, Namespace: request.Namespace}, secret)
+			if err != nil {
+				if k8serr.IsNotFound(err) {
+					reqLogger.Info("SREIAMUserSecret not ready, trying again")
+					time.Sleep(time.Duration(time.Duration(i*1) * time.Second))
+				} else {
+					reqLogger.Error(err, "unable to retrive SREIAMUserSecret secret")
```
dofinn: retrieve :)

PR #263, opened 2020-01-29
Add finalizer and handling for AWS Federated Account Access

fahlmant:
When an AWSFederatedAccountAccess CR is deleted, the federated role needs to be cleaned up by detaching all policies from it and then deleting the role.


clcollins:
/lgtm

This appears to be working as expected in testing.  No notes about the code itself.

clcollins:
Tested w/BYOC acct and federated access, no problems.

/lgtm

PR #270, opened 2020-02-03
Standardize the different timeouts in the operator

ArielLima:
https://issues.redhat.com/browse/OSD-1962

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -48,7 +48,7 @@ const (
 	iamUserNameSRE          = "osdManagedAdminSRE"
 	awsAMI                  = "ami-000db10762d0c4c05"
 	awsInstanceType         = "t2.micro"
-	createPendTime          = 30 * time.Minute
+	createPendTime          = controllerutils.WaitTime * time.Minute
 	// Fields used to create/monitor AWS case
 	caseCategoryCode              = "other-account-issues"
 	caseServiceCode               = "customer-account"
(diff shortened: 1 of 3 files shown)


ArielLima:
This has been texted and works but there are syntax errors that need fixing

pkg/controller/account/account_controller.go
```diff
@@ -407,11 +406,11 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 			}
 		}
 
-		// see if in creating for longer then 10 minutes
+		// see if in creating for longer then 25 minutes
 		now := time.Now()
 		diff := now.Sub(currentAcctInstance.ObjectMeta.CreationTimestamp.Time)
-		if currentAcctInstance.Status.State == "Creating" && diff > createPendTime {
-			r.setStatusFailed(reqLogger, currentAcctInstance, "Creation pending for longer then 10 minutes")
+		if currentAcctInstance.Status.State == "Creating" && diff > controllerutils.WaitTime * time.Minute {
```
iamkirkbater: Can we pull this "calculation" out?  (I don't specifically know what to call it).  Calculations inside complex if statements make the code harder to grok the second time around.

Maybe we should pull this `controllerutils.WaitTime * time.Minute` part out and just put it on the line above, and maybe call it something like `createTimeout`?

fahlmant: Agree. Put the expression (waitTime * time.Minute) as the definition of createPendTime and leave this line as diff > createPendTime

ArielLima: Done

pkg/controller/account/account_controller.go
```diff
@@ -407,11 +407,11 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 			}
 		}
 
-		// see if in creating for longer then 10 minutes
+		// see if in creating for longer then 25 minutes
```
fahlmant: The comment should point to the createPendTime rather than a specific amount

pkg/controller/account/ec2.go
```diff
@@ -100,9 +100,16 @@ func (r *ReconcileAccount) BuildandDestroyEC2Instances(reqLogger logr.Logger, aw
 
 	// Wait till instance is running
 	var DescError error
+	totalWait := controllerutils.WaitTime * 60
+	currentWait := 1
 	for i := 0; i < 300; i++ {
+		currentWait = currentWait * 2
+		if currentWait > totalWait {
```
fahlmant: I'm a bit confused by this logic. Could you explain what you're doing here?

ArielLima: @fahlmant Oops, It is supposed to be `for totalWait > 0 {`. To put simply we double the time between checking if the EC2 instance is running `2 sec, 4 sec, 8 sec, 16 sec,...` until we reach 25 minutes. However, right now instead of ending the loop at 25 minutes it would reconcile for the rest of 300 loops, wouldn't crash but definitely bad. I'll add a comment and fix the loop. Thanks for catching!

fahlmant: I think the other loop in this file has the same issue.

ArielLima: No it doesn't

PR #274, opened 2020-02-04
Removes IAM specific functions to the iam.go file

iamkirkbater:


pkg/controller/account/account_controller.go
```diff
@@ -234,11 +230,6 @@ type SREConsoleInput struct {
 	SecretName, NameSpace, awsCredsConsoleLoginURL string
 }
 
-// input for new aws client
-type newAwsClientInput struct {
-	awsCredsSecretIDKey, awsCredsSecretAccessKey, awsToken, awsRegion, secretName, nameSpace string
```
iamkirkbater: Looking at this the first time around I thought it might fit better in the iam.go file but now that I'm looking at it in the diff I'm not sure.  Thoughts?

pkg/controller/account/account_controller.go
```diff
@@ -9,10 +9,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/awserr"
-	"github.com/aws/aws-sdk-go/service/iam"
-	"github.com/aws/aws-sdk-go/service/organizations"
 	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/go-logr/logr"
 	"github.com/openshift/aws-account-operator/pkg/controller/utils"
```
iamkirkbater: I also noticed in this file (because I had to add it to the other file as well) we're referencing the controller utils twice, once as `utils` here and again as `controllerutils` in the expanded view below.  Should I fix that in this PR or make another card for that?

pkg/controller/account/account_controller.go
fahlmant: This should stay with the main controller logic.

pkg/controller/account/account_controller.go
fahlmant: I would say consolidate them in this PR

pkg/controller/account/account_controller.go
```diff
@@ -814,211 +696,6 @@ func (r *ReconcileAccount) setStatusFailed(reqLogger logr.Logger, awsAccount *aw
 	return nil
 }
 
-// CreateAccount creates an AWS account for the specified accountName and accountEmail in the orgnization
-func CreateAccount(reqLogger logr.Logger, client awsclient.Client, accountName, accountEmail string) (*organizations.DescribeCreateAccountStatusOutput, error) {
```
fahlmant: This function is mostly related to organizations, not IAM.

clcollins:
/lgtm

Tested locally and was working fine for me as well.

PR #275, opened 2020-02-05
Adds owner refs for created secrets

dofinn:
When an account is deleted, the linked secrets are now also removed.


fahlmant:
Looks good, can you please test this? I think you'll need to rebase due to a bug we fixed after you branched off master.

PR #278, opened 2020-02-07
Implement Route53 hosted zone cleanup for reuse

fahlmant:
This PR cleans up route53 hosted zones during the reuse cleanup code. A hosted zone can only be deleted if all record sets, except for the SOA and NS records, are deleted. Certman operator uses TXT records for certificate setup, so included in this cleanup is the deletion of TXT records within the hosted zone before the deletion is called.

### DIFF
--- a/pkg/controller/accountclaim/reuse.go
+++ b/pkg/controller/accountclaim/reuse.go
@@ -8,6 +8,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/route53"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/go-logr/logr"
 	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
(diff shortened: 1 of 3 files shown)


pkg/controller/accountclaim/reuse.go
```diff
@@ -265,6 +267,83 @@ func (r *ReconcileAccountClaim) cleanUpAwsAccountS3(reqLogger logr.Logger, awsCl
...
+					awsErrors <- recordSetListError
+					return listRecordsError
+				}
+
+				changeBatch := &route53.ChangeBatch{}
+				for _, record := range recordSet.ResourceRecordSets {
+					// Build ChangeBatch
+					// https://docs.aws.amazon.com/sdk-for-go/api/service/route53/#ChangeBatch
+					//https://docs.aws.amazon.com/sdk-for-go/api/service/route53/#Change
+					if *record.Type == "TXT" {
```
rogbas: Considering the DeleteHostedZone call will fail if there are records different from NS/SOA in a hosted zone, shouldn't we clean up all records that are not NS/SOA here, instead of just the TXT ones?

rogbas: That way we avoid failing reuse in case customers add CNAMEs/ALIAS for example.

fahlmant: :+1:

jharrington22: :+1: We should check and delete all records in the hosted zone (except NS/SOA of course).

PR #279, opened 2020-02-10
Put aws accounts in organization ou

ArielLima:
Ticket: https://issues.redhat.com/browse/OSD-1603

For billing and reporting purposes put all AWS Accounts into an OU  in the AWS Organization where the OU represents the customer's organization / account.  This will be based on AccountClaim's legalEntity.id field. 

Done:
- When Account status is updated with a legalEntity.id (and was not set prior):
- Create OU in AWS Organization for the .status.legalEntity.id if it doesn't already exist.
- Move the AWS Account to the OU representing the .status.legalEntity.id

Testing:
- Apply an account CR and wait until the account is ready
- Apply an accountclaim CR
- Check Amazon console for the new OU and check that the account is inside

cc @fahlmant @jharrington22

Edit: description, testing


pkg/controller/accountclaim/accountclaim_controller.go
```diff
@@ -314,7 +319,31 @@ func (r *ReconcileAccountClaim) Reconcile(request reconcile.Request) (reconcile.
...
+			return reconcile.Result{}, err
+		}
 
+		callerIdentityOutput, err := awsClient.GetCallerIdentity(&sts.GetCallerIdentityInput{})
+		if err != nil {
+			reqLogger.Error(err, "Failed to get AWS account parent")
+			return reconcile.Result{}, err
+		}
+		// Move Account to Corresponding OU
+		_, err = awsClient.MoveAccount(&organizations.MoveAccountInput{AccountId: &unclaimedAccount.Spec.AwsAccountID, DestinationParentId: &unclaimedAccount.Spec.LegalEntity.ID, SourceParentId: callerIdentityOutput.Account})
```
fahlmant: Thinking about this... if the create is in the account controller but the move is here, what happens if the OU doesn't exist yet?

ArielLima: I have no idea, I assumed things would be sequential since I was told to do it this way.

fahlmant: I think this function will need to be moved out into the account controller

ArielLima: For me it makes more sense to check/create the OU in the accountclaim controller since that is where we will have the logic to move it.

fahlmant: Or at least some higher level logic will need to be applied

rogbas:
@ArielLima can you please test this PR in a scenario where we reset the account and update the legalID?

ArielLima:
Manually editing the account cr **_legalEntity.id_** does not move an account.
Manually editing the accountclaim cr **_legalEntity.id_** does not create a new ou or move the account.
Resetting the account **_reset_account_cr.sh_** does not move the account back into root.

Currently an account does not get move into or out of an OU unless an accountClaim makes it do so. Also an OU never gets deleted, is this an issue?

@rogbas @fahlmant Do we want to move an account whenever the CRs **_legalEntity.id_** changes? It wouldn't be an issue in stg but would it not induce a security issue in prod?

rogbas:
@jharrington22 @fahlmant @ArielLima @rogbas to sync to review/test that PR.

The use-case where an account changes to a different LegalEntityID is not a normal workflow, so we shouldn't consider it here.

rogbas:
@ArielLima please add unit tests for the new ConfigMap so we can review and merge it

Makefile
```diff
@@ -28,6 +28,15 @@ ifndef AWS_IAM_ARN
 	$(error AWS_IAM_ARN is undefined)
 endif
 
+.PHONY: .check-ou-mapping-configmap-env
+check-ou-mapping-configmap-env:
+ifndef OSD_STAGING_1_OU_ROOT_ID
```
fahlmant, line 33: What should these be set to?

pkg/controller/accountclaim/accountclaim_controller.go
```diff
@@ -28,6 +32,7 @@ import (
 const (
 	AccountClaimed          = "AccountClaimed"
 	AccountUnclaimed        = "AccountUnclaimed"
+	OUName                  = "aws-account-operator-ou-mapping"
```
fahlmant, line 35: Is the the name of the OU or of the configmap?

Makefile
ArielLima: The ID of the root and base (location where new OUs are created) OUs.
2 page in the [design doc](https://docs.google.com/document/d/1IK3M32J5nQkS7eyZ-npU2dSQJxMdyfrFAPra4_SBScM/edit#)

pkg/controller/accountclaim/accountclaim_controller.go
ArielLima: configmap, I agree that the name should be more clear, changing.

fahlmant:
@ArielLima Could you give some instructions on how to use the Makefile target, as well as what the ROOT and BASE parts of the OU name are for?

Makefile
```diff
@@ -222,4 +231,39 @@ endif
...
+	@oc process -p ROOT=${OSD_STAGING_1_OU_ROOT_ID} -p BASE=${OSD_STAGING_1_OU_BASE_ID} -f hack/templates/aws_v1alpha1_configmap_ou_mapping.tmpl | oc apply -f -
+
+# Test delete ou map cr
+.PHONY: delete-ou-map
+delete-ou-map:
+# Delete OU map
+	@oc process -p ROOT=${OSD_STAGING_1_OU_ROOT_ID} -p BASE=${OSD_STAGING_1_OU_BASE_ID} -f hack/templates/aws_v1alpha1_configmap_ou_mapping.tmpl | oc delete -f -
+
+# Test claim-account
+.PHONE: claim-account
```
fahlmant, line 249: Is this different from the `create-accountclaim` target?

Makefile
```diff
@@ -222,4 +231,39 @@ endif
...
+claim-account:
+	# Create account claim CR
+	@oc process -p NAME=${OSD_STAGING_1_ACCOUNT_CR_NAME_OSD}-claim -p NAMESPACE=${NAMESPACE} -f hack/templates/aws_v1alpha1_accountclaim_cr.tmpl | oc apply -f -
+	# Wait for account to become claimed
+	@while true; do CLAIMED=$$(oc get account ${OSD_STAGING_1_ACCOUNT_CR_NAME_OSD} -n aws-account-operator -o json | jq -r '.status.claimed'); if [ "$$CLAIMED" == "true" ]; then break; fi; sleep 1; done
+
+# Test delete account-claim
+.PHONY: delete-account-claim
```
fahlmant, line 257: Do we need this in addition to `delete-accountclaim`?

pkg/controller/accountclaim/accountclaim_controller.go
```diff
@@ -28,6 +32,7 @@ import (
 const (
 	AccountClaimed          = "AccountClaimed"
 	AccountUnclaimed        = "AccountUnclaimed"
+	OUConfigMapName         = "aws-account-operator-ou-mapping"
```
fahlmant, line 35: There are a few other things that the operator could use a config map for. Can we make a single generic config map?  

cc @rogbas

Makefile
ArielLima: yes, `create-accountclaim` uses the `test-claim-namespace` namespace and I am trying to use `aws-account-operator`. I am talking to @jharrington22 about possibly being able to use the target but as of now I don't think it would work.

Makefile
ArielLima: same namespace issues

pkg/controller/accountclaim/accountclaim_controller.go
ArielLima: agreed, I will change this

Makefile
```diff
@@ -220,6 +229,43 @@ ifndef OPERATOR_SECRET_ACCESS_KEY
...
+
+# Test delete ou map cr
+.PHONY: delete-ou-map
+delete-ou-map:
+# Delete OU map
+	@oc process -p ROOT=${OSD_STAGING_1_OU_ROOT_ID} -p BASE=${OSD_STAGING_1_OU_BASE_ID} -f hack/templates/aws_v1alpha1_configmap.tmpl | oc delete -f -
+
+# Test claim-account
+.PHONE: claim-ou-account
+claim-ou-account:
```
fahlmant, line 246: Would it make more sense for this to be
`claim-ou-account: create-accoute create-account-claim-namespace create-account-claim` ?
 That way the logic of each of those doesn't have to be repeated

ArielLima: Since targets can only be used a single time per run I tried to keep these targets as self contained as possible so that we can add it to the `test-all` target.

fahlmant: Why not just test add OU stuff as part of normal account creation/claiming then?

pkg/controller/accountclaim/accountclaim_controller.go
```diff
@@ -308,7 +313,101 @@ func (r *ReconcileAccountClaim) Reconcile(request reconcile.Request) (reconcile.
 		if err != nil {
 			return reconcile.Result{}, err
 		}
+	}
+
+	// Set awsAccountClaim.Spec.AwsAccountOU
+	if unclaimedAccount.Spec.AwsAccountOU == "" {
```
fahlmant, line 319: Can we wrap all the logic under this into a function?

`if unclaimedAccount.Spec.AwsAccountOU == "" {
    placeAccountInOU(...)
}`

pkg/controller/accountclaim/accountclaim_controller.go
```diff
@@ -378,6 +508,15 @@ func AWSAccountIDClaimed(accountClaim *awsv1alpha1.AccountClaim, accountClaimLis
 	return match
 }
 
+func checkOUMapping(cMap *corev1.ConfigMap) (string, string, error) {
+	baseID := cMap.Data["base"]
```
fahlmant, line 512: These should be wrapped in an Ok block in case they don't exist.

ArielLima: I did some manual testing and it did not fail but rather returned an empty string when the values did not exists. That is why I make [this](https://github.com/openshift/aws-account-operator/blob/c274fabb2ac1df6c1546882e570fc71f279c3916/pkg/controller/accountclaim/accountclaim_controller.go#L514-L516) check. If you, however, think it's better in the Ok block I do not have strong feelings about it.

fahlmant: The values or the keys?

ArielLima: the keys, sorry

Makefile
ArielLima: Done

pkg/controller/accountclaim/accountclaim_controller.go
ArielLima: I split the logic into better components. I personally prefer the way it is now over dumping it all inside a function but let me know if you disagree.

pkg/controller/account/account_controller.go
```diff
@@ -467,11 +467,11 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 		}
 
 		// Initialize all supported regions by creating and terminating an instance in each
-		err = r.InitializeSupportedRegions(reqLogger, awsv1alpha1.CoveredRegions, creds)
-		if err != nil {
-			r.setStatusFailed(reqLogger, currentAcctInstance, "Failed to build and destroy ec2 instances")
-			return reconcile.Result{}, err
-		}
+		// err = r.InitializeSupportedRegions(reqLogger, awsv1alpha1.CoveredRegions, creds)
```
fahlmant, line 470: Was this meant to be removed?

ArielLima: lol yes

pkg/controller/accountclaim/accountclaim_controller.go
```diff
@@ -257,7 +259,7 @@ func (r *ReconcileAccountClaim) Reconcile(request reconcile.Request) (reconcile.
 	}
 
 	// Return if this claim has been satisfied
-	if accountClaim.Spec.AccountLink != "" && accountClaim.Status.State == awsv1alpha1.ClaimStatusReady {
+	if accountClaim.Spec.AccountLink != "" && accountClaim.Status.State == awsv1alpha1.ClaimStatusReady && accountClaim.Spec.AccountOU != "" {
```
fahlmant, line 262: Could we move this check into a function that returns a bool? Similar to what Christ recently did for the BYOC logic.

pkg/controller/accountclaim/accountclaim_controller.go
fahlmant: If you wrap both in ok blocks, that elimates the need for 436 and is semantically the "golang" way. Curious about other's opinions though

pkg/controller/account/account_controller.go
ArielLima: fixed

pkg/controller/accountclaim/accountclaim_controller.go
fahlmant: It clutters up the accountclaim reconcile function. We recently cleaned up the account reconciler to better organize the steps. Opinions from others?

Makefile
fahlmant: I don't see where this is done. I am suggesting to combine this ou testing with the current account/accountclaim targets

pkg/controller/accountclaim/organizational_units.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/go-logr/logr"
+
+	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
+	awsclient "github.com/openshift/aws-account-operator/pkg/awsclient"
+)
+
+const (
+	byocPolicy        = "BYOCEC2Policy"
+	arnIAMPrefix      = "arn:aws:iam::"
+	byocUserArnSuffix = ":user/byocSetupUser"
```
fahlmant, line 17: Where are these used?

ArielLima: They are not, removing

pkg/controller/accountclaim/accountclaim_controller.go
ArielLima: :+1:

hack/templates/aws_v1alpha1_ccs_accountclaim_cr.tmpl
```diff
@@ -27,4 +27,4 @@ objects:
       namespace: ${NAMESPACE}
     legalEntity:
       id: "111111"
-      name: ${NAME}
+      name: ${NAME}
```
fahlmant, line 30: Please revert this change

hack/templates/aws_v1alpha1_configmap.tmpl
```diff
@@ -0,0 +1,16 @@
...
+  name: test-aws-ou-mapping-configmap-template
+objects:
+- apiVersion: v1
+  kind: ConfigMap
+  metadata:
+    name: aws-account-operator-configmap
+    namespace: aws-account-operator
+  data:
+    root: ${ROOT}
+    base: ${BASE}
```
fahlmant, line 16: Please add a newline

ArielLima: :+1:

hack/templates/aws_v1alpha1_ccs_accountclaim_cr.tmpl
ArielLima: :+1:

pkg/controller/accountclaim/accountclaim_controller.go
ArielLima: Done

Makefile
ArielLima: sorry, I forgot to remove this target.

pkg/apis/aws/v1alpha1/accountclaim_types.go
```diff
@@ -18,9 +18,10 @@ type AccountClaimSpec struct {
 	AwsCredentialSecret SecretRef   `json:"awsCredentialSecret"`
 	Aws                 Aws         `json:"aws"`
 	AccountLink         string      `json:"accountLink"`
-	BYOC                bool        `json:"byoc,omitempty"`
-	BYOCSecretRef       SecretRef   `json:"byocSecretRef,omitempty"`
-	BYOCAWSAccountID    string      `json:"byocAWSAccountID,omitempty"`
+	AccountOU           string
```
fahlmant, line 21: Is this used? I don't see any JSON for this

fahlmant: Nvm it does show up, but please add the json to have `accountOU`

pkg/apis/aws/v1alpha1/account_types.go
```diff
@@ -38,7 +38,9 @@ type AccountSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
-	AwsAccountID  string `json:"awsAccountID"`
+	AwsAccountID string `json:"awsAccountID"`
+	// +optional
+	AwsAccountOU  string `json:"awsAccountOU,omitempty"`
```
fahlmant, line 43: You need to regen the CRDs for this to show up I think

Makefile
```diff
@@ -220,6 +229,28 @@ ifndef OPERATOR_SECRET_ACCESS_KEY
...
+	@oc process -p ROOT=${OSD_STAGING_1_OU_ROOT_ID} -p BASE=${OSD_STAGING_1_OU_BASE_ID} -f hack/templates/aws_v1alpha1_configmap.tmpl | oc apply -f -
+
+# Test delete ou map cr
+.PHONY: delete-ou-map
+delete-ou-map:
+# Delete OU map
+	@oc process -p ROOT=${OSD_STAGING_1_OU_ROOT_ID} -p BASE=${OSD_STAGING_1_OU_BASE_ID} -f hack/templates/aws_v1alpha1_configmap.tmpl | oc delete -f -
+
+# Test aws ou logic
+.PHONY: test-aws-ou-logic
```
fahlmant, line 245: Needs to ensure the `test-claim-namespace` namespace exists.

pkg/apis/aws/v1alpha1/account_types.go
```diff
@@ -38,7 +38,8 @@ type AccountSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
-	AwsAccountID  string `json:"awsAccountID"`
+	AwsAccountID string `json:"awsAccountID"`
+	// +optional
```
fahlmant, line 42: Why was this added?

pkg/apis/aws/v1alpha1/accountclaim_types.go
```diff
@@ -18,6 +18,7 @@ type AccountClaimSpec struct {
 	AwsCredentialSecret SecretRef   `json:"awsCredentialSecret"`
 	Aws                 Aws         `json:"aws"`
 	AccountLink         string      `json:"accountLink"`
+	AccountOU           string      `json:"accountou,omitempty"`
```
fahlmant, line 21: Please regenerate the CRDs

pkg/controller/accountclaim/organizational_units.go
```diff
@@ -0,0 +1,209 @@
...
+	friendlyOUName := account.Spec.LegalEntity.ID
+	baseID, rootID, err := checkOUMapping(instance)
+	if err != nil {
+		invalidOUErrorMsg := fmt.Sprintf("Invalid OU ConfigMap, missing root and/or base fields: %s", instance.Data)
+		reqLogger.Error(err, invalidOUErrorMsg)
+		return err
+	}
+
+	// Log ConfigMap
+	configmapMsg := fmt.Sprintf("OU: OU mapping ConfigMap was found in namespace: %s, root: %s, base: %s", OUConfigMapName, rootID, baseID)
```
fahlmant, line 48: I'm not sure how helpful this log will be since it will be nearly the same info for every accountcaim.

pkg/controller/accountclaim/organizational_units.go
```diff
@@ -0,0 +1,209 @@
...
+	accountMovedMsg := fmt.Sprintf("OU: Account %s successfully moved to OU %s", account.Name, account.Spec.LegalEntity.ID)
+	reqLogger.Info(accountMovedMsg)
+
+	// Update unclaimedAccount.Spec.AwsAccountOU
+	accountClaim.Spec.AccountOU = OUID
+	return r.specUpdate(reqLogger, accountClaim)
+}
+
+// CreateorFindOU will create or find an existing OU and return its ID
+func CreateorFindOU(reqLogger logr.Logger, client awsclient.Client, account *awsv1alpha1.Account, friendlyOUName string, baseID string) (string, error) {
```
fahlmant, line 80: Please use camel case, i.e. `CreateOrFindOU`

pkg/controller/accountclaim/organizational_units.go
```diff
@@ -0,0 +1,209 @@
...
+		ParentId: &baseID,
+	}
+	ouOutput, ouErr := client.CreateOrganizationalUnit(&createCreateOrganizationalUnitInput)
+	if ouErr != nil {
+		if aerr, ok := ouErr.(awserr.Error); ok {
+			switch aerr.Code() {
+			case "DuplicateOrganizationalUnitException":
+				duplicateOUMsg := fmt.Sprintf("OU: %s Already exists", account.Spec.LegalEntity.ID)
+				reqLogger.Info(duplicateOUMsg)
+				return findOUIDFromName(reqLogger, client, baseID, friendlyOUName)
```
fahlmant, line 93: What happens if this returns an error? Does it need any special handling?

pkg/apis/aws/v1alpha1/account_types.go
ArielLima: forgot to remove it after removing `AwsAccountOU`

pkg/apis/aws/v1alpha1/accountclaim_types.go
ArielLima: thanks for the link, doing it now

pkg/controller/accountclaim/organizational_units.go
ArielLima: you're right, removing this log

pkg/controller/accountclaim/organizational_units.go
ArielLima: thanks for pointing this out

pkg/controller/accountclaim/organizational_units.go
ArielLima: Possible errors are already handled

fahlmant: Where?

ArielLima: In the [method](https://github.com/openshift/aws-account-operator/blob/7f5d2f824385e8c296998fca99ee1577c829bb1e/pkg/controller/accountclaim/organizational_units.go#L170) itself I log unexpected errors and I handle the return [values] when I get them(https://github.com/openshift/aws-account-operator/blob/7f5d2f824385e8c296998fca99ee1577c829bb1e/pkg/controller/accountclaim/organizational_units.go#L52-L55)

fahlmant: In `findOUIDFromName` you only handle if there's an aws error, not if there's some generic error. There's also no specific handling for any aws error, so why unwrap it if you're going to return it straight up to the parent function? Line 176 of this file. I'm wondering if you're going to specifically handle the error from findOUIDFromName anywhere.

ArielLima: I see what you’re saying about making sure I catch generic errors as well, I’ll fix that. I always unwrap AWS errors because I was told it is good practice + it helps with debugging if there is ever a need but I am open to other opinions.

fahlmant: You should unwrap it. I just want to verify that there's nothing specific you want to do with that error based on its message.

ArielLima: I see, well unlike some other parts of this PR where I use specific errors as a queue to apply different logic this particular case any error would be bad. That's why I just log and return.

rogbas:
@ArielLima can you please add the ConfigMap objects for stage, integration and production environments and get a +1 from @aweiteka ?

aweiteka:
LGTM

ArielLima:
ConfigMap added to hive, stage and int

rogbas:
@rogbas sync with @ArielLima to merge and track changes in stage/int

PR #282, opened 2020-02-12
Fix bug where missing role causes finalizer inifinite loop

fahlmant:
Found a bug where if the AWS Federated Account Access CR fails before its able to create the role in the target account, then the finalize code looped forever. This PR fixes it by returning from the role cleanup code if the role doesn't exist.

### DIFF
--- a/pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
+++ b/pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
@@ -469,6 +469,13 @@ func (r *ReconcileAWSFederatedAccountAccess) cleanFederatedRoles(reqLogger logr.
 		if err != nil {
 			if aerr, ok := err.(awserr.Error); ok {
 				switch aerr.Code() {
+				case "NoSuchEntity":
+					// Delete any custom policies made
+					err = r.deleteNonAttachedCustomPolicy(reqLogger, awsClient, federatedRoleCR)
+					if err != nil {
+						return err
+					}
+					return nil
 				default:
 					reqLogger.Error(
 						aerr,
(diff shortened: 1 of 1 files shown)


fahlmant:
Realized that even though the role might not exist, the custom policies may exist. Going to handle custom policy deletion in the case the role doesn't exist.

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -469,6 +469,13 @@ func (r *ReconcileAWSFederatedAccountAccess) cleanFederatedRoles(reqLogger logr.
 		if err != nil {
 			if aerr, ok := err.(awserr.Error); ok {
 				switch aerr.Code() {
+				case "NoSuchEntity":
+					// Delete any custom policies made
+					err = r.deleteNonAttachedCustomPolicy(reqLogger, awsClient, federatedRoleCR)
+					if err != nil {
+
```
rafael-azevedo: your missing some logic here or the return is in the wrong long.

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -550,3 +557,52 @@ func (r *ReconcileAWSFederatedAccountAccess) cleanFederatedRoles(reqLogger logr.
...
+	var policyMarker *string
+	// Paginate through custom policies
+	for {
+		policyListOutput, err := awsClient.ListPolicies(&iam.ListPoliciesInput{Scope: aws.String("Local"), Marker: policyMarker})
+		if err != nil {
+			if aerr, ok := err.(awserr.Error); ok {
+				switch aerr.Code() {
+				default:
+					reqLogger.Error(
```
rafael-azevedo: Why are there two logging lines here , and why are the formats different in both?

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -550,3 +557,52 @@ func (r *ReconcileAWSFederatedAccountAccess) cleanFederatedRoles(reqLogger logr.
...
+
+func (r *ReconcileAWSFederatedAccountAccess)  deleteNonAttachedCustomPolicy(reqLogger logr.Logger, awsClient awsclient.Client, federatedRoleCR *awsv1alpha1.AWSFederatedRole) error {
+
+	var policyMarker *string
+	// Paginate through custom policies
+	for {
+		policyListOutput, err := awsClient.ListPolicies(&iam.ListPoliciesInput{Scope: aws.String("Local"), Marker: policyMarker})
+		if err != nil {
+			if aerr, ok := err.(awserr.Error); ok {
+				switch aerr.Code() {
```
rafael-azevedo: Since we are not doing any filtering here we shouldn't be using the switch we should just return the error. If we need more granular error handling in the future we can add it at that point.

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -550,3 +557,52 @@ func (r *ReconcileAWSFederatedAccountAccess) cleanFederatedRoles(reqLogger logr.
...
+		}
+
+		for _, policy := range policyListOutput.Policies {
+			if *policy.PolicyName == federatedRoleCR.Spec.AWSCustomPolicy.Name {
+				_, err = awsClient.DeletePolicy(&iam.DeletePolicyInput{PolicyArn: policy.Arn})
+				if err != nil {
+					if aerr, ok := err.(awserr.Error); ok {
+						switch aerr.Code() {
+						default:
+							reqLogger.Error(
```
rafael-azevedo: See above comment about switch case and the logging.

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -550,3 +557,52 @@ func (r *ReconcileAWSFederatedAccountAccess) cleanFederatedRoles(reqLogger logr.
...
+							reqLogger.Error(err, fmt.Sprintf("%v", err))
+							return err
+						}
+					}
+				}
+			}
+		}
+
+		if *policyListOutput.IsTruncated {
+			policyMarker = policyListOutput.Marker
```
rafael-azevedo: I would put a continue in here to say if the output is truncated set the marker and continue. Then you can remove the else below and just have break.

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
fahlmant: Bad cop paste. Removing 575 and 592

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
fahlmant: This makes it easily extensible in the future. I.e. line 472 above.

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -550,3 +553,44 @@ func (r *ReconcileAWSFederatedAccountAccess) cleanFederatedRoles(reqLogger logr.
...
+					return err
+				}
+			}
+		}
+
+		for _, policy := range policyListOutput.Policies {
+			if *policy.PolicyName == federatedRoleCR.Spec.AWSCustomPolicy.Name {
+				_, err = awsClient.DeletePolicy(&iam.DeletePolicyInput{PolicyArn: policy.Arn})
+				if err != nil {
+					if aerr, ok := err.(awserr.Error); ok {
```
jharrington22: What happens if its not an `awserr` ?

fahlmant: Whoop, forgot those.

fahlmant:
Tested and cleaned up

PR #283, opened 2020-02-14
remove dep for createBYOCAdminAccessRole

dofinn:
This change removes a dependency on the aws-account-operator-credentials
details when processing a BYOC account claim. This enables BYOC to run
completely independent. For example, aws-account-operator-credentials
can be for an AWS Global account while being able to process BYOC for
AWS China.


dofinn:
`reqLogger.Error(err, "Failed to get non-BYOC IAM User info")`

Why are we actually testing this?

fahlmant:
The only function of the BYOC user creds passed to us is to create a role that our AWS user can assume and create the necessary IAM users, instances etc.

PR #284, opened 2020-02-14
Local dev mode

iamkirkbater:
I took @rafael-azevedo's PR on Local Dev Mode and ran with it to fix the other points in the code that need to get modified for local testing.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -11,6 +11,7 @@ import (
 	routev1 "github.com/openshift/api/route/v1"
 	"github.com/openshift/aws-account-operator/pkg/apis"
 	"github.com/openshift/aws-account-operator/pkg/controller"
+	"github.com/openshift/aws-account-operator/pkg/controller/utils"
 	"github.com/openshift/aws-account-operator/pkg/credentialwatcher"
 	"github.com/openshift/aws-account-operator/pkg/localmetrics"
 	"github.com/openshift/aws-account-operator/pkg/totalaccountwatcher"
@@ -79,7 +80,6 @@ func main() {
 	// Become the leader before proceeding
 	err = leader.Become(ctx, "aws-account-operator-lock")
 	if err != nil {
-		log.Error(err, "")
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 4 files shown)


dofinn:
/lgtm

if we ever update the operator-sdk we can also use something like this: https://github.com/openshift/certman-operator/pull/87

cmd/manager/main.go
```diff
@@ -29,13 +29,16 @@ import (
...
-	totalWatcherInterval          = time.Duration(15) * time.Minute
+	metricsPort               = "8080"
+	metricsPath               = "/metrics"
+	secretWatcherScanInterval = time.Duration(10) * time.Minute
+	hours                     = 1
+	totalWatcherInterval      = time.Duration(15) * time.Minute
 )
 
+// Set Local evn "FORCE_DEV_MODE = local" in order to run locally
+const envDevMode = "FORCE_DEV_MODE"
```
ArielLima: Maybe I am missing something but couldn't we inject this value into another package? that way we don't need to retrieve it through `os` i.e. `utils.envDevMode="FORCE_DEV_MODE"`? I would even consider using a boolean if testing is the only "DevMode" we use. `utils.testMode`?

iamkirkbater: I'm not completely sure what you're suggesting when you say 'injecting into another package'.

Do you mean something like editing `utils.go` and adding a line like `testMode := false` at the top of the file and changing that line when you're trying to run it locally?

ArielLima: @iamkirkbater Yes

ArielLima: I am way more stoked than I should be for this

iamkirkbater: The only problem I have with that is then you start modifying the code again to change the flag.  Part of the idea here is that we don't want to physically modify the code before committing or after running our tests. 

---

With that said, though, I don't think it would be a _bad_ idea to pull this functionality out to somewhere like `utils.go` but still utilize something like an environment variable or command option there, and then the rest of the codebase will just do something like `if ! utils.TestMode`.

So in that case in `utils.go` we would have `TestMode := os.getEnv('FORCE_DEV_MODE')` instead of setting the same variable and such throughout the codebase.

ArielLima: I see, then I agree with you. Move it to utils but keep it the same

ArielLima:
Do we need to initiate and terminate all supported regions when performing local tests?
https://github.com/openshift/aws-account-operator/pull/284/files#diff-32f12988d73ded1726ea7c67c8cb1f90R584
cc/ @jharrington22 @fahlmant @rogbas

clcollins:
> Do we need to initiate and terminate all supported regions when performing local tests?

I would say yes, @ArielLima.  It might be a somewhat misleading name, but developing with the operator-sdk locally still needs to let us test against AWS resources.  This is more a reference to the fact that the operator is not running in a cluster, rather than doing totally self-contained tests.

ArielLima:
> > Do we need to initiate and terminate all supported regions when performing local tests?
> 
> I would say yes, @ArielLima. It might be a somewhat misleading name, but developing with the operator-sdk locally still needs to let us test against AWS resources. This is more a reference to the fact that the operator is not running in a cluster, rather than doing totally self-contained tests.

@clcollins I understand that it is important to test API calls but doesn't it cost us money every time? It seems to me that it should be something that is tested separate from the local tests we make all the time.

dofinn:
The totalaccountwatcher could/can also be negligible, I commented it out for AWS China testing as all it did was slow me down and didnt need to be apart of testing. We have some refactoring tickets, should we look to up the SDK and use this: https://github.com/openshift/certman-operator/pull/87

ArielLima:
> The totalaccountwatcher could/can also be negligible, I commented it out for AWS China testing as all it did was slow me down and didnt need to be apart of testing. We have some refactoring tickets, should we look to up the SDK and use this: [openshift/certman-operator#87](https://github.com/openshift/certman-operator/pull/87)

is secretWatcher another one that can be skipped unless you are specifically testing for it? In the last couple of days it's been getting in my way (creating a lot of noise) since initiating and terminating instances in all regions can sometimes hang for a few minutes.

dofinn:
> > The totalaccountwatcher could/can also be negligible, I commented it out for AWS China testing as all it did was slow me down and didnt need to be apart of testing. We have some refactoring tickets, should we look to up the SDK and use this: [openshift/certman-operator#87](https://github.com/openshift/certman-operator/pull/87)
> 
> is secretWatcher another one that can be skipped unless you are specifically testing for it? In the last couple of days it's been getting in my way (creating a lot of noise) since initiating and terminating instances in all regions can sometimes hang for a few minutes.

maybe local dev script should detect local dev, then prompt for each major functionality that we could disable?

iamkirkbater:
If we want to do something like that we might as well just go all in and figure out a good way to do some feature flags.

rogbas:
@iamkirkbater can you please address the createCase comment and rebase it?

After that we should be able to merge.

PR #285, opened 2020-02-17
Update Federated Role CRs to include roleDisplayName

fahlmant:
Both our defined Federated Roles are missing the roleDisplayName needed to apply them.


jharrington22:
@fahlmant should we add `namespace: aws-account-operator` here too?

PR #288, opened 2020-02-19
Enable AWS China support for BYOC

dofinn:
This PR is to remove dependency on the aws-account-operator-credentials
and allow byoc account_claims to be processed independently using only
the credentials from the claim.

Major changes:
* a byoc client is always instantiated for byoc instances
* aws keys are retrieved pre/post key rotation and compared to avoid
race conditions that were present in testing
* a new client is created with the rotated keys
* the new clients keys are validated to assure the keys have propogated
in AWS infra: https://stackoverflow.com/questions/20156043/how-long-should-i-wait-after-applying-an-aws-iam-policy-before-it-is-valid
* tagging was added to the ec2 instances created to provide clarity
while testing with other IAM users of the same account

These changes were tested with all enpoints and arn's updated for AWS
China while using a secret for an AWS global account for the
aws-account-operator-credentials secret. This confirms the byoc path as
indepedent.


pkg/controller/utils/utils.go
```diff
@@ -113,3 +116,21 @@ func Remove(list []string, s string) []string {
...
+		if err != nil {
+			if aerr, ok := err.(awserr.Error); ok {
+				if aerr.Code() == "InvalidClientTokenId" {
+					reqLogger.Info(fmt.Sprintf("New credentials not yet available after rotating. Attempt %b/6", retry))
+				} else {
+					reqLogger.Error(err, "Failed to create client with rotated credentials")
+					return err
+				}
+			}
+			time.Sleep(10 * time.Second)
```
jharrington22: 10 seconds is a long time to wait, the reconcile loop will stop and wait here. Could we change this to something like;

```
duration := 500 * time.Duration(retry) * time.Millisecond
time.Sleep(duration)
```

pkg/controller/account/account_controller.go
```diff
@@ -313,16 +316,67 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
+					} else {
+						// reque request if keys match
+						time.Sleep(500 * time.Millisecond)
+					}
+					// This may need tuning.
+					if i == 20 {
+						reqLogger.Info("Timeout attempting to validate rotated keys")
+					}
+				}
+				reqLogger.Info("Rotated keys successfully validated")
```
jharrington22: Should this logic go into a separate function and but put inside `byocRotateAccessKeys()` ?

pkg/controller/account/account_controller.go
```diff
@@ -285,19 +285,22 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
+		byocSetupClient, accountClaim, err := r.getBYOCClient(currentAcctInstance)
+		if err != nil {
+			if accountClaim != nil {
+				r.accountClaimBYOCError(reqLogger, accountClaim, err)
 			}
+			return reconcile.Result{}, err
+		}
 
+		// re-assign the client to byoc instantiation
+		awsSetupClient = byocSetupClient
```
jharrington22: Can we just update the `awsSetupClient` reference to `byocSetupClient` everywhere?

pkg/controller/account/account_controller.go
dofinn: It could. I would like to submit as is and make a note to refactor this and every other occasion of key rotation. Compare secrets/ validate client as a utility.

pkg/controller/account/account_controller.go
dofinn: I would think not, the byoc client is created once the account_claim is identified as a BYOC type. Only then, do we reassign `byocSetupClient` to `awsSetupClient` so the rest of the code can remain unchanged.

pkg/controller/utils/utils.go
dofinn: Yes. good spot, will update this.

pkg/controller/account/account_controller.go
```diff
@@ -860,3 +914,22 @@ func matchSubstring(roleID, role string) (bool, error) {
 	matched, err := regexp.MatchString(roleID, role)
 	return matched, err
 }
+
+// getAWSAccessKeyId retrieves the awsCredsSecretIDKey for the current account
+func (r *ReconcileAccount) getAWSAccessKeyId(currentAccountClaim *awsv1alpha1.AccountClaim) (accessKeyID string, err error) {
```
fahlmant: I think this would make more sense in the `iam.go` file.

fahlmant:
@dofinn The code generally looks good at a glance and the account CR becomes Ready successfully. So if I understand this correctly, we're using the customer's provided user's creds for every call instead of the creds in `aws-account-operator-credentials`.  Doing this eliminates the reliance on our creds, but this also potentially locks us completely out of the BYOC account. The role created in the customer's target account is meant to be assumed by our user. By eliminating that, if the customer changes the access key of the IAM user they provide, or delete that user, SRE has no other path for managing the customer's account. I'm not sure what the solution is though. What are the reqs from AWS China?

dofinn:
Doesn't the BYOC implementation leave us vulnerable to being cut off from managing the account anyway? deletion of roles/users for BYOC account.

dofinn:
Closing. will submit another with revised changes.

PR #293, opened 2020-02-27
Refactor byocRotateAccessKeys method

ArielLima:
Ticket: https://issues.redhat.com/browse/OSD-2840

Refactor byocRotateAccessKeys function into smaller logical components

https://github.com/openshift/aws-account-operator/blob/e1cd076bfbf6d306ada944787c619f6f8db9171d/pkg/controller/account/byoc.go#L245

Done criteria:
- byocRotateAccessKeys should be concise and easy to understand
- write unit test for at least one of the new functions

**Still need to make unit tests for the new functions**
**This is a fork of #314 so that should be merged first**

Edit: Description


ArielLima:
just rebased still needs unit tests

rogbas:
@ArielLima will add a unit test to this PR.

pkg/controller/account/byoc.go
```diff
@@ -283,49 +284,23 @@ func (r *ReconcileAccount) byocRotateAccessKeys(reqLogger logr.Logger, byocAWSCl
 
 	// Get the BYOC credentials secret to update
 	accountClaimSecret := &corev1.Secret{}
-
 	err = r.Client.Get(context.TODO(),
 		types.NamespacedName{Name: accountClaim.Spec.BYOCSecretRef.Name, Namespace: accountClaim.Spec.BYOCSecretRef.Namespace},
 		accountClaimSecret)
 
+	// Get secret access key
```
fahlmant, line 291: This gets the access key part, not the secret part

pkg/controller/account/byoc.go
```diff
@@ -362,3 +337,47 @@ func validateBYOCClaim(accountClaim *awsv1alpha1.AccountClaim) error {
 
 	return nil
 }
+
+func deleteOtherAccessKeys(reqLogger logr.Logger, client awsclient.Client, accessKeyID string) error {
```
fahlmant, line 341: I think `iam.go` would be a better place for this function.

pkg/controller/account/byoc.go
```diff
@@ -362,3 +337,47 @@ func validateBYOCClaim(accountClaim *awsv1alpha1.AccountClaim) error {
...
+	// Update secret
+	secret.Data =
+		map[string][]byte{
+			"aws_access_key_id":     []byte(*userSecretInfo.AccessKey.AccessKeyId),
+			"aws_secret_access_key": []byte(*userSecretInfo.AccessKey.SecretAccessKey),
+		}
+	return secret, nil
+}
+
+// func renewBYOCAccessKey(reqLogger logr.Logger, client awsclient.Client, username string) (*iam.CreateAccessKeyOutput, error) {
```
fahlmant, line 381: Is this still needed?

pkg/controller/account/byoc.go
```diff
@@ -362,3 +337,47 @@ func validateBYOCClaim(accountClaim *awsv1alpha1.AccountClaim) error {
...
+			if err != nil {
+				reqLogger.Error(err, "Failed to delete BYOC access keys")
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func replaceSecretAccessKey(reqLogger logr.Logger, client awsclient.Client, secret corev1.Secret, userName string, accessKeyID *string) (corev1.Secret, error) {
```
fahlmant, line 356: Please leave a comment about what this function does. The name suggest to me it just deals with the secret, but the function creates and deletes IAM resources as well.

rogbas:
@ArielLima please add the additional unit test before we can merge this PR

pkg/controller/account/byoc.go
ArielLima: @fahlmant I might be confused but added a comment with what I believe this method does

pkg/controller/account/byoc.go
```diff
@@ -387,3 +362,33 @@ func validateBYOCClaim(accountClaim *awsv1alpha1.AccountClaim) error {
 
 	return nil
 }
+
+// replaceSecretAccessKey will create a new access key and use it to replace a users current one
```
fahlmant, line 366:
```suggestion
// replaceSecretAccessKey deletes an existing access key, then creates a new one and places it in the supplied secret
```

pkg/controller/account/byoc.go
iamkirkbater: Maybe a better name for this function would be `RotateIAMUserKeys`?

pkg/controller/account/byoc.go
```diff
@@ -387,3 +362,29 @@ func validateBYOCClaim(accountClaim *awsv1alpha1.AccountClaim) error {
...
+		return secret, err
+	}
+
+	// Update secret
+	secret.Data =
+		map[string][]byte{
+			"aws_access_key_id":     []byte(*userSecretInfo.AccessKey.AccessKeyId),
+			"aws_secret_access_key": []byte(*userSecretInfo.AccessKey.SecretAccessKey),
+		}
+	return secret, nil
```
iamkirkbater, line 389: I'm wondering if we want to return a secret here, or if we should keep the part about the secret in the function calling this one and just return the AccessKeyID and SecretAccessKey and an error here instead?  If we were pulling data out of the secret at first to be able to rotate the keys then I'd have different feelings about this, but I feel the part about putting the data into the secret just smells wrong here.  I'm not dying on this hill, though.

pkg/controller/account/byoc.go
```diff
@@ -387,3 +362,29 @@ func validateBYOCClaim(accountClaim *awsv1alpha1.AccountClaim) error {
...
+	// Create new BYOC access keys
+	userSecretInfo, err := CreateUserAccessKey(reqLogger, client, userName)
+	if err != nil {
+		failedToCreateUserAccessKeyMsg := fmt.Sprintf("Failed to create IAM access key for %s", userName)
+		reqLogger.Info(failedToCreateUserAccessKeyMsg)
+		return secret, err
+	}
+
+	// Delete original BYOC access key
+	_, err = client.DeleteAccessKey(&iam.DeleteAccessKeyInput{AccessKeyId: accessKeyID})
```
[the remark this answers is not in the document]
iamkirkbater: Actually, I'm now seeing that we're deleting all other access keys first.  Is there any reason we just don't delete ALL access keys, every time?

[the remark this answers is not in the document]
fahlmant: Because we still need one access key at all times.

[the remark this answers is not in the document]
fahlmant: The other keys are already deleted in line 320

pkg/controller/account/byoc.go
fahlmant: I had a similar thought, but this function rotates the keys AND places it in a secret. Ideally this would be two functions.

rogbas:
@iamkirkbater @rogbas can you please take a look and review?

This will need a rebase after we merge #351 today.

ArielLima:
PR #351 did what this would have done. I suggest we close this @rogbas @jharrington22

jharrington22:
@ArielLima yes I think this can be closed due to #351

PR #294, opened 2020-02-28
Aws customer admin role

iamkirkbater:
https://issues.redhat.com/browse/OSD-2793

Generates JSON such as the following:

```json
{
  "Version":"2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Action": ["ec2:DescribeInstances"],
      "Resource":["*"],
      "Condition": {
        "StringEquals": {
          "ram:RequestedResourceType": "route53resolver:ResolverRule"
        }
      }
    }
  ]
}
```


pkg/controller/account/iam.go
```diff
@@ -388,6 +395,71 @@ func deleteAllAccessKeys(reqLogger logr.Logger, client awsclient.Client, userNam
 	return nil
 }
 
+// CreateAccount creates an AWS account for the specified accountName and accountEmail in the orgnization
+func CreateAccount(reqLogger logr.Logger, client awsclient.Client, accountName, accountEmail string) (*organizations.DescribeCreateAccountStatusOutput, error) {
```
fahlmant: Is this move from another PR?

pkg/controller/utils/utils.go
```diff
@@ -19,26 +19,30 @@ const (
...
-		Action    []string               `json:"Action"`
-		Resource  []string               `json:"Resource,omitempty"`
-		Principal *awsv1alpha1.Principal `json:"Principal,omitempty"`
-	}
+// The JSON tags as captials due to requirements for the policydoc
+type awsStatement struct {
+	Effect    string                       `json:"Effect"`
+	Action    []string                     `json:"Action"`
+	Resource  []string                     `json:"Resource,omitempty"`
+	Condition map[string]map[string]string `json:"Condition,omitempty"`
```
fahlmant: Can I suggest that we implement a Condition type that will wrap this? This type is hard to parse visually.

It might be a good idea to just implement all of the possible elements of the JSON policy struct in AWS
https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements.html

pkg/controller/account/iam.go
iamkirkbater: Yeah.  It's one of the reasons I made this a WIP because I still have to clean this up, but I wanted to get this up so I remembered what I was doing on Friday when Monday came 😂

pkg/controller/utils/utils.go
iamkirkbater: Hmm.  If we go that far I wonder if it would be worth it to just try to pull this entire package out and make it a standalone package, as we'd be just generalizing the entire IAM yaml -> json conversion.

fahlmant: I'm :+1: to that. It doesn't look like AWS go sdk has any way to do this. But maybe better in a different PR. For this PR, I think at least making Condition its own type will help with readability

fahlmant: I kinda hate the map syntax

iamkirkbater: Yeah, that's fair.  I definitely wasn't a fan of trying to do the map of a map thing, but it was the quickest way to do it without having to declare the condition operator in the type.  I wonder if I can use the `json: -` thing to get that to work.  I'll test that out.

fahlmant:
/hold

This lgtm but I want mwoodson to review. Could you also rebase?

PR #296, opened 2020-03-02
Fix Role/Policy creation and deletion to support multiple users 

fahlmant:
This PR has 2 commits to address 2 different but related bugs
1) If two users share a role via an AWS Federated Account Access CR, currently deleting one will remove the Roles/Policies from the target account. This leaves the other user unable to use the role. To fix this, two labels are appended to each Federated Account Access when they become ready: role: rolename and account: awsAccountID. Then when an AWS Federated Account Access CR is deleted, it checks first to see if any other FAA CRs have the same label. If not, then the Roles/Policies are cleaned up. If at least one other has the same labels, then the Roles/Policies are NOT cleaned up

2) We currently have code that is supposed to handle if a Policy or Role already exists by using the AWS Error code `EntityAlreadyExists`. However, the functions that were supposed to return this error were returning another non-AWS error. While I was fixing this, I also figured that the code should not try to delete and recreate these since that could affect access for users that already have access to those Roles/Policies. Instead, the code assumes if the Role or Policy already exists in AWS to use them as is. Not sure if this is the best solution but it seems to be the cleanest and safest at the moment.

### DIFF
--- a/pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
+++ b/pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
@@ -20,6 +20,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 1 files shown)


fahlmant:
Closing in favor of #300 and #301

PR #299, opened 2020-03-03
Move constants and types that could shared in the package v1alpha1

ArielLima:
Ticket: https://issues.redhat.com/browse/OSD-2835

Create pkg/apis/aws/v1alpha1/shared_types.go add constants and types that could shared in the package v1alpha1

https://github.com/openshift/aws-account-operator/blob/master/pkg/controller/account/account_controller.go#L79-L138

Done criteria:
- Move types and constants that are shared between APIs to this file
- Any constants that we can reference here to keep controller files clean

Testing:
- To test you just need to apply  an account CR and check that it becomes ready without issues.

Edit: Description

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -2,7 +2,6 @@ package account
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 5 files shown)


ArielLima:
Is it reasonable to move errors out as well? Are we doping this to every controller with this PR?
cc @jharrington22 @rogbas

ArielLima:
**Tested with new make tests**

pkg/apis/shared_types.go
```diff
@@ -0,0 +1,100 @@
+package apis
```
fahlmant: Is there a reason this isn't put into the utils package or something similar? This is not a very descriptive name IMO

ArielLima: Simply followed the ticket, I can move it though. What do you think? @rogbas @jharrington22

fahlmant: The tickets specifies the pkg `v1alpha1` as I read it.

ArielLima: https://issues.redhat.com/browse/OSD-2835 @fahlmant

jharrington22: Yes apis isn't quite right. I wanted it to be part of the v1alpha1 package I just got the path wrong in the ticket. @ArielLima please update.

pkg/controller/account/account_controller.go
```diff
@@ -29,6 +28,7 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	"github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
 	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
```
iamkirkbater: Is there a reason we're importing this again?

pkg/controller/account/iam.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/go-logr/logr"
+	"github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
 	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
```
iamkirkbater: Same question, we're importing the same package twice here.

pkg/controller/account/account_controller.go
ArielLima: No, fixed

pkg/controller/account/iam.go
ArielLima: No, fixed

fahlmant:
@ArielLima Can you update the PR title and the commit message?

ArielLima:
> @ArielLima Can you update the PR title and the commit message?

:+1:

PR #300, opened 2020-03-04
Fix bug where using same role in account with second user fails

fahlmant:
We currently have code that is supposed to handle if a Policy or Role already exists by using the AWS Error code EntityAlreadyExists. However, the functions that were supposed to return this error were returning another non-AWS error. 

In order to fix these, each FAA with get a unique UUID appended as a label at creation time. Then every role and policy created will get that uuid appended to the end of the name during creation and deletion actions.

### DIFF
--- a/pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
+++ b/pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
@@ -12,6 +12,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/rand"
 
 	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
 	"github.com/openshift/aws-account-operator/pkg/awsclient"
@@ -157,6 +158,35 @@ func (r *ReconcileAWSFederatedAccountAccess) Reconcile(request reconcile.Request
 		return reconcile.Result{}, nil
 	}
 
+	// Check if the FAA has the uid label
+	if !hasUIDLabel(currentFAA) {
+		// Generate a new UID
+		uid := rand.String(6)
+
+		reqLogger.Info(fmt.Sprintf("Adding UID %s to AccountAcess %s", uid, currentFAA.Name))
+		newLabel := map[string]string{"uid": uid}
+
+		// Join the new UID label with any current labels
+		if currentFAA.Labels != nil {
+			currentFAA.Labels = joinLabelMaps(currentFAA.Labels, newLabel)
+		} else {
+			currentFAA.Labels = newLabel
+		}
+
+		// Update the CR with new labels
+		err = r.client.Update(context.TODO(), currentFAA)
+		if err != nil {
+			reqLogger.Error(err, fmt.Sprintf("Lable update for %s failed", currentFAA.Name))
+			return reconcile.Result{}, err
+		}
+
+	}
+
+	uidLabel, ok := currentFAA.Labels["uid"]
+	if !ok {
+		return reconcile.Result{}, err
+	}
+
 	// Get aws client
 	awsClient, err := awsclient.GetAWSClient(r.client, awsclient.NewAwsClientInput{
 		SecretName: currentFAA.Spec.AWSCustomerCredentialSecret.Name,
@@ -173,7 +203,7 @@ func (r *ReconcileAWSFederatedAccountAccess) Reconcile(request reconcile.Request
 	gciOut, err := awsClient.GetCallerIdentity(&sts.GetCallerIdentityInput{})
 	if err != nil {
 		SetStatuswithCondition(currentFAA, "Failed to get account ID information", awsv1alpha1.AWSFederatedAccountFailed, awsv1alpha1.AWSFederatedAccountStateFailed)
-		utils.LogAwsError(log, fmt.Sprintf("Unable to create role requested by '%s'", currentFAA.Name), err, err)
+		utils.LogAwsError(log, fmt.Sprintf("Failed to get account ID information for '%s'", currentFAA.Name), err, err)
 		err := r.client.Status().Update(context.TODO(), currentFAA)
 		if err != nil {
 			reqLogger.Error(err, fmt.Sprintf("Status update for %s failed", currentFAA.Name))
@@ -226,12 +256,12 @@ func (r *ReconcileAWSFederatedAccountAccess) Reconcile(request reconcile.Request
 	// Get policy arns for managed policies
 	policyArns := createPolicyArns(accountID, awsManagedPolicyNames, true)
 	// Get custom policy arns
-	customPolicy := []string{requestedRole.Spec.AWSCustomPolicy.Name}
+	customPolicy := []string{requestedRole.Spec.AWSCustomPolicy.Name + "-" + uidLabel}
 	customerPolArns := createPolicyArns(accountID, customPolicy, false)
 	policyArns = append(policyArns, customerPolArns[0])
 
 	// Attach the requested policy to the newly created role
-	err = r.attachIAMPolices(awsClient, currentFAA.Spec.AWSFederatedRole.Name, policyArns)
+	err = r.attachIAMPolices(awsClient, currentFAA.Spec.AWSFederatedRole.Name+"-"+uidLabel, policyArns)
 	if err != nil {
 		//TODO() role should be deleted here so that we leave nothing behind.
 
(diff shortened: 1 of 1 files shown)


pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -157,6 +158,35 @@ func (r *ReconcileAWSFederatedAccountAccess) Reconcile(request reconcile.Request
...
+	if !hasUUIDLabel(currentFAA) {
+		// Generate a new UUID
+		uuid := rand.String(6)
+
+		reqLogger.Info(fmt.Sprintf("Adding UUID %s to AccountAcess %s", uuid, currentFAA.Name))
+		newLabel := map[string]string{"uuid": uuid}
+
+		// Join the new UUID label with any current labels
+		if currentFAA.Labels != nil {
+			currentFAA.Labels = joinLabelMaps(currentFAA.Labels, newLabel)
```
fahlmant: This could be cleaned up by putting the check login in the joinLabelMaps function. note for future improvements.

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -157,6 +158,35 @@ func (r *ReconcileAWSFederatedAccountAccess) Reconcile(request reconcile.Request
...
+		// Update the CR with new labels
+		err = r.client.Update(context.TODO(), currentFAA)
+		if err != nil {
+			reqLogger.Error(err, fmt.Sprintf("Lable update for %s failed", currentFAA.Name))
+			return reconcile.Result{}, err
+		}
+
+	}
+
+	uuidLabel, ok := currentFAA.Labels["uuid"]
```
fahlmant: Could make this a function that returns a label and err since this same 4 lines is done a lot through code. Future improvement.

sam-nguyen7: +1

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -289,13 +319,22 @@ func (r *ReconcileAWSFederatedAccountAccess) createIAMPolicy(awsClient awsclient
 		return &iam.Policy{}, fmt.Errorf("Error marshalling jsonPolicy doc : Error %s", err.Error())
 	}
 
+	var policyName string
+	// Try and build policy name
+	if uuidLabel, ok := afaa.Labels["uuid"]; ok {
+		policyName = afr.Spec.AWSCustomPolicy.Name + "-" + uuidLabel
```
iamkirkbater: Do we need to do any checks here to make sure that that the `afr.Spec.AWSCustomPolicy.Name` is not longer than 128 characters minus the 37 characters from the `-[uuid]`?

Maybe even shorter? (Are we using the Switch Role feature or are we assuming this role through another mechanism?)

https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-limits.html#w477aac29b9c11b5

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -326,54 +365,93 @@ func (r *ReconcileAWSFederatedAccountAccess) createIAMRole(awsClient awsclient.C
...
 	jsonAssumeRolePolicyDoc, err := json.Marshal(&assumeRolePolicyDoc)
 	if err != nil {
-		return nil, fmt.Errorf("Error marshalling jsonPolicy doc : Error %s", err.Error())
+		return nil, err
+	}
+
+	var roleName string
+	// Try and build role name
+	if uuidLabel, ok := afaa.Labels["uuid"]; ok {
+		roleName = afr.Name + "-" + uuidLabel
```
iamkirkbater: Similar to above, do we need to do any checks here to make sure that that the `afr.Spec.AWSCustomPolicy.Name` is not longer than 64 characters minus the 37 characters from the `-[uuid[?

Maybe even shorter? (Are we using the Switch Role feature or are we assuming this role through another mechanism?)

https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-limits.html#w477aac29b9c11b5

fahlmant: Probably same as above, we should validate length before trying to create. We're using switch role since we put the url in the CR

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
fahlmant: Yeah, probably should be checked before the code gets to this point, i.e. check at the start of the reconcile and fail if the length is too long.

fahlmant: ah, the uuid I'm using is only 6  chars fyi

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -157,6 +158,35 @@ func (r *ReconcileAWSFederatedAccountAccess) Reconcile(request reconcile.Request
 		return reconcile.Result{}, nil
 	}
 
+	// Check if the FAA has the uuid label
+	if !hasUUIDLabel(currentFAA) {
+		// Generate a new UUID
+		uuid := rand.String(6)
```
iamkirkbater: If we're not using a full 36-char UUID here could we maybe call this something else?  Maybe just UID?  At first glance I was expecting this to be a UUID and not just 6 chars.

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
iamkirkbater: This was a mistake on my part where I thought UUID was a full 36 char UUID string and not just 6 characters.

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
iamkirkbater: This was a mistake on my part where I missed that UUID was only 6 random characters and not a full 36 char UUID string.  We should still add a check here for length, but it's definitely not as bad now with us only using up 7 out of 64 chars.

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
fahlmant: :+1: Changing

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -446,12 +446,11 @@ func (r *ReconcileAWSFederatedAccountAccess) createOrUpdateIAMRole(awsClient aws
...
 					return nil, err
 				}
-				if role == nil {
-					return nil, errors.New("Create Role Output is nil")
-				}
-
 				return role, nil
 			}
+		} else {
+			// Catch any other unexpected errors from the AWS API
```
clcollins: Handles: https://github.com/openshift/aws-account-operator/pull/300#pullrequestreview-369920242

jharrington22:
> Reporting the occasional crash that we found during pair debugging:
> 
> ```
> E0305 16:12:08.956319   67831 runtime.go:69] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
> /Users/kbater/Projects/go/src/github.com/openshift/aws-account-operator/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:76
> /Users/kbater/Projects/go/src/github.com/openshift/aws-account-operator/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:65
> /Users/kbater/Projects/go/src/github.com/openshift/aws-account-operator/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:51
> /usr/local/Cellar/go/1.13.6/libexec/src/runtime/panic.go:679
> /usr/local/Cellar/go/1.13.6/libexec/src/runtime/panic.go:199
> /usr/local/Cellar/go/1.13.6/libexec/src/runtime/signal_unix.go:394
> /Users/kbater/Projects/go/src/github.com/openshift/aws-account-operator/pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go:249
> /Users/kbater/Projects/go/src/github.com/openshift/aws-account-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:215
> /Users/kbater/Projects/go/src/github.com/openshift/aws-account-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:158
> /Users/kbater/Projects/go/src/github.com/openshift/aws-account-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:133
> /Users/kbater/Projects/go/src/github.com/openshift/aws-account-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:134
> /Users/kbater/Projects/go/src/github.com/openshift/aws-account-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:88
> /usr/local/Cellar/go/1.13.6/libexec/src/runtime/asm_amd64.s:1357
> panic: runtime error: invalid memory address or nil pointer dereference [recovered]
>         panic: runtime error: invalid memory address or nil pointer dereference
> [signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x21c2a8c]```
> ```

This will be resolved in the next PR.

PR #301, opened 2020-03-04
Add labels to FAAs

fahlmant:
If two users share a role via an AWS Federated Account Access CR, currently deleting one will remove the Roles/Policies from the target account. This leaves the other user unable to use the role. To fix this, two labels are appended to each Federated Account Access when they become ready: role: rolename and account: awsAccountID. Then when an AWS Federated Account Access CR is deleted, it checks first to see if any other FAA CRs have the same label. If not, then the Roles/Policies are cleaned up. If at least one other has the same labels, then the Roles/Policies are NOT cleaned up


fahlmant:
Closing in favor of #300 
Since every FAA will manage its own role, the deletion of the FAA will handle its own role. That makes this code obsolete, as no roles are shared.

PR #303, opened 2020-03-06
Adds handling of unexpected AWS API errors

clcollins:
Currently, if there is an error creating the role, and the error is an AWS API error (awserr.Error), only "EntityAlreadyExists" is handled, and all other errors are slilently dropped on the floor.  This adds an `else` case to return the awserr.Error.Code() string as an error if it isn't the EntityAlreadyExists error.

REF: panics[1] due to role being nil, ultimately tracked down to unhandled "MalformedPolicyDocument" errors returned by AWS that were being ignored by the code.

[1] https://github.com/openshift/aws-account-operator/pull/300#pullrequestreview-369920242


pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -439,18 +439,22 @@ func (r *ReconcileAWSFederatedAccountAccess) createOrUpdateIAMRole(awsClient aws
 		if aerr, ok := err.(awserr.Error); ok {
```
jharrington22: What happens here if the error is not cast as an AWS error?

PR #304, opened 2020-03-06
Handles non-AWS errors returned by createIAMRole function

clcollins:
This change handles errors that do not fall under the `awserr.Error`
type, and ensures they are returned if encountered, and not dropped.

### DIFF
--- a/pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
+++ b/pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
@@ -230,7 +230,7 @@ func (r *ReconcileAWSFederatedAccountAccess) Reconcile(request reconcile.Request
 	}
 
 	// Create role and apply custom policys and awsmanagedpolicies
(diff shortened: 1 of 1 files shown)


pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -451,12 +451,12 @@ func (r *ReconcileAWSFederatedAccountAccess) createOrUpdateIAMRole(awsClient aws
 				}
 
 				return role, nil
-
-			} else {
-				// Handle unexpected AWS API errors
-				return nil, errors.New(aerr.Code())
 			}
+			// Handle unexpected AWS API errors
+			return nil, errors.New(aerr.Code())
```
jharrington22: One comment, in other places we explicitly log that this is an unknown error from AWS could you do the same? https://github.com/openshift/aws-account-operator/blob/master/pkg/controller/account/iam.go#L422

clcollins: Updated.  Adds a switch/case for known, and default for unknown, using the `util.LogAwsError` as requested.

PR #307, opened 2020-03-10
Only rotate STS credentials when Account CR state is Ready

jharrington22:
This PR should change secretwatcher to only rotate STS credentials if the Account CR is in a Ready state.

I'm also reducing the scan interval to 1 minute so that we're renewing less secrets on every run.

### DIFF
--- a/pkg/credentialwatcher/secretwatcher.go
+++ b/pkg/credentialwatcher/secretwatcher.go
@@ -123,6 +123,12 @@ func (s *secretWatcher) updateAccountRotateCredentialsStatus(log logr.Logger, ac
 		return
 	}
 
+	// Only rotate STS credentials if the account CR is in a Ready state
+	if accountInstance.Status.State != string(awsv1alpha1.AccountReady) {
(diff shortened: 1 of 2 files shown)


pkg/credentialwatcher/secretwatcher.go
```diff
@@ -123,6 +123,12 @@ func (s *secretWatcher) updateAccountRotateCredentialsStatus(log logr.Logger, ac
 		return
 	}
 
+	// Only rotate STS credentials if the account CR is in a Ready state
+	if accountInstance.Status.State != string(awsv1alpha1.AccountReady) {
```
clcollins: Does `string(awsv1alpha1.AccountReady)` have to be wrapped in `string()`?  It's already a string, isn't it?

clcollins:
Left a small comment, but otherwise:

/lgtm

pkg/credentialwatcher/secretwatcher.go
jharrington22: It should be fixed as its actually of type `AccountConditionType`

PR #295, opened 2020-02-28
Moves complex logic and simple checks to function that return boolean results

clcollins:
This change moves some simple checks and complex logic to functions that
    return boolean results for the different possible states of the account,
    allowing more testable checks readable code.
    
    Also includes tests for the new functions to validate they are returning
    the correct output given the expected account input.
    
    Signed-off-by: Christopher Collins <collins.christopher@gmail.com>

PR #316, opened 2020-03-18
Separates BYOC Logic from account_controller.go

clcollins:
This moves the BYOC initialization and logic from the account_controller.go file to byoc.go via a new `initializeNewBYOCAccount` function.  This has been tested locally and is working as expected.

This depends on PR #295 and should not be merged before that.

REF: [https://issues.redhat.com/browse/OSD-2833](https://issues.redhat.com/browse/OSD-2833)

### DIFF
--- a/pkg/controller/account/byoc.go
+++ b/pkg/controller/account/byoc.go
@@ -36,12 +36,9 @@ var roleID = ""
 
 // BYOC Accounts are determined by having no state set OR not being claimed
 // Returns true if either are true AND Spec.BYOC is true
-func accountIsNewBYOC(currentAcctInstance *awsv1alpha1.Account) bool {
-	if currentAcctInstance.Spec.BYOC {
-		if !accountHasState(currentAcctInstance) {
-			return true
-		}
-		if !accountIsClaimed(currentAcctInstance) {
+func newBYOCAccount(currentAcctInstance *awsv1alpha1.Account) bool {
+	if accountIsBYOC(currentAcctInstance) {
+		if !accountHasState(currentAcctInstance) || !accountIsClaimed(currentAcctInstance) {
 			return true
 		}
 	}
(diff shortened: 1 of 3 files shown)


clcollins:
Rebased from #295 (dependency) and ran the `make test-all` tests with no issues.

clcollins:
#295 has been merged

rogbas:
@rogbas @jharrington22 could you please take a look and lgtm?

rogbas:
@rogbas @jharrington22 can you please re-test and review the PR as well?

ArielLima:
It seems to work for me. I applied a BYOC AccountClaim and it successfully spun and claimed an account. However, it will need to be tested again when the conflicts are resolved. @rogbas @jharrington22 @clcollins

clcollins:
Rebased and retested OK.

pkg/controller/account/byoc.go
```diff
@@ -59,6 +56,61 @@ func claimBYOCAccount(r *ReconcileAccount, reqLogger logr.Logger, currentAcctIns
 	return nil
 }
 
+// Checks whether or not the access keys need to be rotated based on state, and if so, rotates them
+func (r *ReconcileAccount) initializeCredentials(reqLogger logr.Logger, awsSetupClient, client awsclient.Client, currentAcctInstance *awsv1alpha1.Account, accountClaim *awsv1alpha1.AccountClaim, adminAccessArn string) (string, error) {
+
+	err := r.byocRotateAccessKeys(reqLogger, client, accountClaim)
```
fahlmant, line 62: This was removed in master, lets remove here as well

pkg/controller/account/byoc.go
```diff
@@ -59,6 +56,61 @@ func claimBYOCAccount(r *ReconcileAccount, reqLogger logr.Logger, currentAcctIns
...
+	if err != nil {
+		return roleID, err
+	}
+
+	return roleID, err
+}
+
+func (r *ReconcileAccount) initializeNewBYOCAccount(reqLogger logr.Logger, currentAcctInstance *awsv1alpha1.Account, awsSetupClient awsclient.Client, adminAccessArn string) (string, error) {
+	client, accountClaim, err := r.getBYOCClient(currentAcctInstance)
+	if err != nil && accountClaim != nil {
```
fahlmant, line 76: Should this be an `or`?

fahlmant: I see what's happening. But I think we need to check if the error is nil, or if we can't find the accountclaim. Not sure what that logic should look like

pkg/controller/account/byoc.go
fahlmant: Since this is being removed, does it make sense to remove this function totally?

clcollins: Removed `initializeCredentials` function and call to `byocRotateAccessKeys`.  Moved `createBYOCAdminAccessRole` to parent loop.

Removed `byocRotateAccessKeys` function as it is no longer used.

pkg/controller/account/byoc.go
clcollins: Good catch, thanks.  Addressed so that if there is an error, and the account claim is not nil, it calls `r.accountClaimBYOCError`, and then regardless, returns the error.

clcollins:
Re-tested after implementing your suggestions @fahlmant, and looks OK.

rogbas:
I was able to run it locally and validate the PR.
+1 to merge it next Monday Jun 1st.

PR #317, opened 2020-03-19
Handle the new UID assignment to FAA Roles in the Makefile tests

clcollins:
This adds the ability to retrieve the FAA Role UID and use it for the `test-switch-role` target, automating that step where it was previously a manual step.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,8 +78,10 @@ create-awsfederatedaccountaccess: check-aws-account-id-env
 
 .PHONY: test-switch-role
 test-switch-role:
+	# Retrieve role UID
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -78,8 +78,10 @@ create-awsfederatedaccountaccess: check-aws-account-id-env
 
 .PHONY: test-switch-role
 test-switch-role:
+	# Retrieve role UID
+	$(eval UID=$(shell oc get awsfederatedaccountaccesses.aws.managed.openshift.io  test-federated-user -o=json |jq -r .metadata.labels.uid))
```
iamkirkbater: Is there a reason we use the full domain for this?  Just curious.

clcollins: The CRD has it as a valid name, and I think that was just a tab-complete from the command I was using.  Both the full one and `awsfederatedaccountaccess` works.  I don't have a preference for which to use, though.

PR #302, opened 2020-03-04
Finalizer refactor

iamkirkbater:
Refactors Finalizers to their own files.

Tested by creating a BYOC accountclaim and by creating an account and accountclaim.

PR #318, opened 2020-03-20
Enable AWS China for CCS/BYOC

dofinn:
This PR enabled the aws-account-operator to spawn AWS clients, based on
the region passed in the acccountClaim. The platform (AWS)
now has two default configs, one for AWS Global and one for AWS China.
Each PlatformConfig struct contains platform specific:
* AWS Client
* ARN Prefix
* Covered Regions
* Federation Config

EC2 instance tagging is also included to provide clarity over what
process created and destroyed the instances. `AaoInitRegion` = Aws
Account Operator Initialise Region.

### DIFF
--- a/pkg/apis/aws/v1alpha1/shared_types.go
+++ b/pkg/apis/aws/v1alpha1/shared_types.go
@@ -1,55 +1,68 @@
 package v1alpha1
 
-import "errors"
+import (
+	"errors"
+)
 
-// CoveredRegions map
-var CoveredRegions = map[string]map[string]string{
-	"us-east-1": {
-		"initializationAMI": "ami-000db10762d0c4c05",
-	},
-	"us-east-2": {
-		"initializationAMI": "ami-094720ddca649952f",
-	},
-	"us-west-1": {
-		"initializationAMI": "ami-04642fc8fca1e8e67",
-	},
-	"us-west-2": {
-		"initializationAMI": "ami-0a7e1ebfee7a4570e",
-	},
-	"ca-central-1": {
-		"initializationAMI": "ami-06ca3c0058d0275b3",
-	},
-	"eu-central-1": {
-		"initializationAMI": "ami-09de4a4c670389e4b",
-	},
-	"eu-west-1": {
-		"initializationAMI": "ami-0202869bdd0fc8c75",
-	},
-	"eu-west-2": {
-		"initializationAMI": "ami-0188c0c5eddd2d032",
-	},
-	"eu-west-3": {
-		"initializationAMI": "ami-0c4224e392ec4e440",
-	},
-	"ap-northeast-1": {
-		"initializationAMI": "ami-00b95502a4d51a07e",
-	},
-	"ap-northeast-2": {
-		"initializationAMI": "ami-041b16ca28f036753",
-	},
-	"ap-south-1": {
-		"initializationAMI": "ami-0963937a03c01ecd4",
-	},
-	"ap-southeast-1": {
-		"initializationAMI": "ami-055c55112e25b1f1f",
-	},
-	"ap-southeast-2": {
-		"initializationAMI": "ami-036b423b657376f5b",
-	},
-	"sa-east-1": {
-		"initializationAMI": "ami-05c1c16cac05a7c0b",
-	},
-}
+var (
+	// AWSRegionsGlobal provides a map of AWS Region to initializationAMI for AWS Global.
+	AWSRegionsGlobal = map[string]map[string]string{
+		"us-east-1": {
+			"initializationAMI": "ami-000db10762d0c4c05",
+		},
+		"us-east-2": {
+			"initializationAMI": "ami-094720ddca649952f",
+		},
+		"us-west-1": {
+			"initializationAMI": "ami-04642fc8fca1e8e67",
+		},
+		"us-west-2": {
+			"initializationAMI": "ami-0a7e1ebfee7a4570e",
+		},
+		"ca-central-1": {
+			"initializationAMI": "ami-06ca3c0058d0275b3",
+		},
+		"eu-central-1": {
+			"initializationAMI": "ami-09de4a4c670389e4b",
+		},
+		"eu-west-1": {
+			"initializationAMI": "ami-0202869bdd0fc8c75",
+		},
+		"eu-west-2": {
+			"initializationAMI": "ami-0188c0c5eddd2d032",
+		},
+		"eu-west-3": {
+			"initializationAMI": "ami-0c4224e392ec4e440",
+		},
+		"ap-northeast-1": {
+			"initializationAMI": "ami-00b95502a4d51a07e",
+		},
+		"ap-northeast-2": {
+			"initializationAMI": "ami-041b16ca28f036753",
+		},
+		"ap-south-1": {
+			"initializationAMI": "ami-0963937a03c01ecd4",
+		},
+		"ap-southeast-1": {
+			"initializationAMI": "ami-055c55112e25b1f1f",
+		},
+		"ap-southeast-2": {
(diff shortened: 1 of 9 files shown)


pkg/controller/utils/utils.go
```diff
@@ -8,6 +8,7 @@ import (
```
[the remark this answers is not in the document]
dofinn: Might be better in your referenced spot or in a seperate platform_types.

pkg/controller/utils/utils.go
```diff
@@ -17,6 +18,130 @@ const (
...
+	AWSARNPrefixGlobal = "arn:aws:"
+	AWSARNPrefixChina  = "arn:aws-cn:"
+
+	AWSIAMPolicyAdministrator = "iam::aws:policy/AdministratorAccess"
+
+	AWSFedEndpointURLGlobal = "https://signin.aws.amazon.com/federation"
+	AWSFedEndpointURLChina  = "https://signin.amazonaws.cn/federation"
+
+	AWSFedConsURLGlobal = "https://console.aws.amazon.com/"
+	AWSFedConsURLChina  = "https://console.amazonaws.cn/"
```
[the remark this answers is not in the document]
dofinn: I'm not sure what you are requesting? To not use consts and use strings in the structs?

rogbas:
@jharrington22 @fahlmant @rogbas will sync to review/test that PR.

dofinn:
will rebase after https://github.com/openshift/aws-account-operator/pull/302 merged.

rogbas:
@dofinn could you please rebase this PR? A couple of automated make target have been added to the operator recently, making it easier for us to do automated testing of your PR.

pkg/controller/account/account_controller.go
```diff
@@ -165,25 +163,50 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
+
+	// Based on the awsRegion bool, import either the Global or China configuration.
+	AwsPlatformConfig := awsv1alpha1.AwsPlatformConfig{}
+
+	if !checkAwsProviderIsGlobal(awsRegion) {
+		AwsPlatformConfig = awsv1alpha1.AwsPlatformConfigChina
+	} else {
+		AwsPlatformConfig = awsv1alpha1.AwsPlatformConfigGlobal
+	}
+	reqLogger.Info(fmt.Sprintf("AWS Platform configured region: %s", AwsPlatformConfig.ClientInput.AwsRegion))
```
jharrington22, line 193: @dofinn Currently there are 1000's of Account CR's that don't have claims or have a `.spec.ClaimLink` that points to non existing claims. When the accountpool_controller creates Account CRs satisfying the pool they don't have claims, they are awaiting to be claimed. Merging this as is would have caused the account_controller to continually error since it would requeue that Account CR with no ClaimLink and just error again forever.

pkg/controller/account/account_controller.go
```diff
@@ -165,25 +163,50 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
+
+	// Pass awsPlatformConfig to the setup client to create platform appropriate client.
+	awsSetupClient, err := awsclient.GetAWSClient(r.Client, AwsPlatformConfig.ClientInput)
```
jharrington22, line 196: @dofinn Instead of taking the `AwsPlatformConfig` you could create a new function on the awsClient type that would take the AccountClaim and return a client that connects to the correct region based on the `spec.aws.regions`. You would need two functions something like a `GetAwsClientFromClaim` and a `GetAwsBYOCClientFromClaim` these respectively would return the correct client, the first returning one using the credentials from the root (payer) account and the latter returning a client from the credentials in referenced in the AccountClaim. With that we would not need to pass around the `AwsPlatformConfig` struct as the client is created correctly when its instantiated.

pkg/controller/account/account_controller.go
```diff
@@ -165,25 +163,50 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
 
-	// If the account is BYOC, needs some different set up
+	// If the account is BYOC, needs some different set up.
 	if currentAcctInstance.Spec.BYOC {
 		reqLogger.Info("BYOC account")
 		if currentAcctInstance.Status.State == "" || currentAcctInstance.Status.Claimed != true {
-			// Create client for BYOC account
-			byocAWSClient, accountClaim, err := r.getBYOCClient(currentAcctInstance)
+			// Create client for BYOC account.
+			byocAWSClient, err := r.getBYOCClient(currentAcctInstance, accountClaim, AwsPlatformConfig)
```
jharrington22, line 209: See comment on the `GetAWSClient` above.

pkg/apis/aws/v1alpha1/shared_types.go
```diff
@@ -1,56 +1,138 @@
...
+	}
+
+	// AwsPlatformConfigChina provides a default struct with fields configured for AWS China.
+	AwsPlatformConfigChina = AwsPlatformConfig{
+		ClientInput: awsclient.NewAwsClientInput{
+			AwsRegion:  AWSRegionDefaultChina,
+			SecretName: AWSSecretNameChina,
+			NameSpace:  v1alpha1.AccountCrNamespace,
+		},
+		ARNPrefix:      AWSARNPrefixChina,
```
rogbas, line 67: Since we're proposing an alternative to the approach of having different `AwsPlatformConfig` objects, probably it would be better having the ARNs coming from a method called from `awsclient` interface. 
Since in the awsclient has the region, we could have a `GetARNPrefix()` method

pkg/controller/account/ec2.go
```diff
@@ -167,6 +167,17 @@ func CreateEC2Instance(reqLogger logr.Logger, client awsclient.Client, ami strin
...
+				{
+					ResourceType: aws.String("instance"),
+					Tags: []*ec2.Tag{
+						{
+							Key:   aws.String("Name"),
+							Value: aws.String("AaoInitRegion"),
+						},
+					},
+				},
+			},
```
rogbas, line 180: Any special AWS China req for this tag being added to EC2 instances? 
I'm asking because there's already another card/PR taking care of that: https://github.com/openshift/aws-account-operator/pull/339

pkg/controller/account/account_controller.go
rogbas: It doesn't seem a new function has been created in the awsclient interface, receiving the accountClaim (which contains the region) and returning the client awsClient. Any reason for that?

pkg/controller/account/account_controller.go
```diff
@@ -141,71 +139,113 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
-		SecretName: utils.AwsSecretName,
-		NameSpace:  awsv1alpha1.AccountCrNamespace,
-		AwsRegion:  "us-east-1",
-	})
+	// Set these vars in this scope so they can be passed as required.
+	AwsPlatformConfig := utils.AwsPlatformConfig{}
+	AwsPlatformConfig = utils.AwsPlatformConfigGlobal
+
+	//// Return associated accountClaim.
+	//accountClaim, err = r.ReturnAccountClaim(currentAcctInstance)
```
rogbas, line 147: Can you please remove commented code from the PR?

pkg/controller/account/ec2.go
dofinn: Nope, it was there for the exact reason. I will remove.

pkg/apis/aws/v1alpha1/shared_types.go
dofinn: This would require more changes to pass around the region. I recommend doing this in a separate PR for refactoring.

PR #320, opened 2020-03-23
Adds deploy target for automating install of the aws-account-operator

clcollins:
This change adds a deploy target to the Makefile, to automate the install
of the aws-account-operator pieces required for operation.

`make deploy` will create the operator secret, deploy the CRDs, and create an accountPool with zero size.  It's also idempotent, so it can be run multiple times with no issue and will just replace the resources that differ from what it expects (in case you create a bad secret, or whatnot).

REF: [https://issues.redhat.com/browse/OSD-3077](https://issues.redhat.com/browse/OSD-3077)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -184,3 +184,33 @@ list-s3-bucket:
 # Test reuse
 .PHONY: test-reuse
 test-reuse: check-aws-account-id-env create-account create-accountclaim-namespace create-accountclaim create-s3-bucket delete-accountclaim delete-accountclaim-namespace list-s3-bucket delete-account
+
+# Deploy the operator secrets, CRDs and namesapce.
+.PHONY: deploy-aws-account-operator-credentials
+deploy-aws-account-operator-credentials:
+# Base64 Encode the AWS Credentials
+	$(eval ID=$(shell echo -n ${OPERATOR_ACCESS_KEY_ID} | base64 ))
+	$(eval KEY=$(shell echo -n ${OPERATOR_SECRET_ACCESS_KEY} | base64))
+
+.PHONY: deploy-aws-account-operator
+deploy-aws-account-operator:
+# Create aws-account-operator namespace
+	@oc get namespace ${NAMESPACE} || oc new-project ${NAMESPACE}
+# Create the aws-account-operator-credentials secret
+	@oc process -p OPERATOR_ACCESS_KEY_ID=${ID} -p OPERATOR_SECRET_ACCESS_KEY=${KEY} -f hack/templates/aws_v1alpha1_aws_account_operator_credentials.tmpl | oc apply -f -
+# Create aws-account-operator CRDs
+	@ls deploy/crds/*crd.yaml | xargs -L1 oc apply -f
+# Create zero size account pool
+	@oc apply -f hack/files/aws_v1alpha1_zero_size_accountpool.yaml
+
+.PHONY: deploy
+deploy: check-aws-credentials deploy-aws-account-operator-credentials deploy-aws-account-operator
+
+.PHONY: .check-aws-credentials
+check-aws-credentials:
+ifndef OPERATOR_ACCESS_KEY_ID
+	$(error OPERATOR_ACCESS_KEY_ID is undefined)
+endif
+ifndef OPERATOR_SECRET_ACCESS_KEY
+	$(error OPERATOR_SECRET_ACCESS_KEY is undefined)
+endif
--- a/hack/templates/aws_v1alpha1_aws_account_operator_credentials.tmpl
+++ b/hack/templates/aws_v1alpha1_aws_account_operator_credentials.tmpl
@@ -1,0 +1,17 @@
+apiVersion: v1
+kind: Template
+parameters:
+- name: OPERATOR_ACCESS_KEY_ID
+- name: OPERATOR_SECRET_ACCESS_KEY
+metadata:
+  name: test-federated-user-template
+objects:
+- apiVersion: v1
+  data:
+    aws_access_key_id: "${OPERATOR_ACCESS_KEY_ID}"
+    aws_secret_access_key: "${OPERATOR_SECRET_ACCESS_KEY}"
+  kind: Secret
+  metadata:
+    name: aws-account-operator-credentials
+    namespace: aws-account-operator
+    type: Opaque
--- a/README.md
+++ b/README.md
@@ -91,9 +91,24 @@ The below commands can be used to test payer account credentials where we create
 
 ## 1.5. Local Development
 
+### 1.5.1 Operator Install
+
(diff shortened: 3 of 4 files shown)


Makefile
```diff
@@ -92,3 +92,32 @@ delete-awsfederatedaccountaccess: check-aws-account-id-env
 
 .PHONY: test-awsfederatedaccountaccess
 test-awsfederatedaccountaccess: check-aws-account-id-env create-account create-awsfederatedrole create-awsfederatedaccountaccess test-switch-role delete-awsfederatedaccountaccess delete-awsfederatedrole delete-account
+
+.PHONY: deploy-aws-account-operator-credentials
+deploy-aws-account-operator-credentials:
+# Base64 Encode the AWS Credentials
+	$(eval ID=$(shell base64 <<< ${AWS_ACCESS_KEY_ID}))
```
jharrington22: Are you assuming here that people have their `.aws/credentials` setup without profiles?

Makefile
```diff
@@ -92,3 +92,32 @@ delete-awsfederatedaccountaccess: check-aws-account-id-env
...
+.PHONY: deploy-aws-account-operator
+deploy-aws-account-operator:
+# Create aws-account-operator namespace
+	@oc get namespace ${NAMESPACE} || oc new-project ${NAMESPACE}
+# Create the aws-account-operator-credentials secret
+	@oc process -p AWS_ACCESS_KEY_ID=${ID} -p AWS_SECRET_ACCESS_KEY=${KEY} -f hack/templates/aws_v1alpha1_aws_account_operator_credentials.tmpl | oc apply -f -
+# Create aws-account-operator CRDs
+	@ls deploy/crds/*crd.yaml | xargs -L1 oc apply -f
+# Create zero size account pool
+	@oc apply -f - <<< '{"apiVersion": "aws.managed.openshift.io/v1alpha1", "kind": "AccountPool", "metadata": {"name": "test-account-pool"}, "spec":{"poolSize": 0}}'
```
jharrington22: We should change and update the account pool CR in the deploy folder to use here

Makefile
```diff
@@ -92,3 +92,32 @@ delete-awsfederatedaccountaccess: check-aws-account-id-env
...
+
+.PHONY: deploy
+deploy: check-aws-credentials deploy-aws-account-operator-credentials deploy-aws-account-operator
+
+.PHONY: .check-aws-credentials
```
jharrington22: Should we add this to `check-aws-account-id-env`?

Makefile
clcollins: No, the ID and Key are needed as parameters for the credentials secret created from the `aws_v1alpha1_aws_account_operator_credentials.yaml` further down.  I was following the precedent of defining a variable as done for the check-aws-account-id-env target, for consistency.

A deploy would look like: `AWS_ACCESS_KEY_ID="something" AWS_SECRET_KEY_ID="something else" make deploy`

This was also an easy way to base64 encode the credentials before using them in the secret template.

Makefile
clcollins: Done.  I created a zero-size accountpool YAML file in the `hack/` directory and apply that instead.

Makefile
clcollins: I avoided doing that because the AWS ID & Key are not required for the test-awsfederatedaccountaccess target, and the AWS Account IDs are not required for the deploy target.  This way a user doesn't need to set or export them all to do one or the other.

Makefile
ArielLima: I think it would be nice if it assumed `.aws/credentials` was setup. I don't know if this is possible but being able to apply secrets interchangeably between osd-1 & osd-2 would be nice as well.

clcollins: @ArielLima: It's not an AWS limitation, though, otherwise I would just use the `.aws/credentials` file directly as you and James have suggested.

As I mentioned above, the Key and ID are required as parameters to create the account secret in MiniShift, so they have to be parsed and used by the Makefile script.  They can't just be left up to the `aws` command to handle.

iamkirkbater:
I ended up having to delete my CRC cluster locally and rebuild a new one.  Used this to start up a new AWS Account Operator to be able to test.  :100: would bootstrap this way again.

Makefile
iamkirkbater: It would be nice to not have to put in the secrets if they're already in the credentials file, but I don't think that should hold this up.  We should add a backlog item to figure out a way to get those out of the credentials file.

clcollins: Going with @jharrington22's request, I've replaced the use of `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` with `OPERATOR_ACCESS_KEY_ID` and `OPERATOR_SECRET_ACCESS_KEY` to avoid conflict with AWS CLI tool creds.

Parsing an `.aws/credentials` file is impractical as it requires mandating the format of everyone's creds file, which I'd like to avoid.

clcollins:
> Can you update the README to reflect how to bootstrap this?

@iamkirkbater: README.md updated with bootstrap documentation

Makefile
```diff
@@ -70,25 +70,54 @@ create-awsfederatedaccountaccess: check-aws-account-id-env
...
 	@oc process -p AWS_IAM_ARN=${AWS_IAM_ARN} -p IAM_USER_SECRET=${IAM_USER_SECRET} -p AWS_FEDERATED_ROLE_NAME=${AWS_FEDERATED_ROLE_NAME} -p NAMESPACE=${NAMESPACE} -f hack/templates/aws_v1alpha1_awsfederatedaccountaccess_cr.tmpl | oc delete -f -
 
 .PHONY: test-awsfederatedaccountaccess
 test-awsfederatedaccountaccess: check-aws-account-id-env create-account create-awsfederatedrole create-awsfederatedaccountaccess test-switch-role delete-awsfederatedaccountaccess delete-awsfederatedrole delete-account
+
+.PHONY: deploy-aws-account-operator-credentials
+deploy-aws-account-operator-credentials:
+# Base64 Encode the AWS Credentials
+	$(eval ID=$(shell base64 <<< ${OPERATOR_ACCESS_KEY_ID}))
+	$(eval KEY=$(shell base64 <<< ${OPERATOR_SECRET_ACCESS_KEY}))
```
iamkirkbater: These commands need to be changed to the `echo -n ${OPERATOR_ACCESS_KEY_ID} | base64` format otherwise it adds the newline problem I ran into earlier.

hack/templates/aws_v1alpha1_aws_account_operator_credentials.tmpl
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+kind: Template
+parameters:
+- name: AWS_ACCESS_KEY_ID
+- name: AWS_SECRET_ACCESS_KEY
+metadata:
+  name: test-federated-user-template
```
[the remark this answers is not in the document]
clcollins: Changed to OPERATOR_ACCESS_KEY_ID and OPERATOR_SECRET_ACCESS_KEY

Makefile
clcollins: Niiiiiice catch @iamkirkbater!

Fixed!

rogbas:
@clcollins could you please remove from the PR the changes that are not related to automating the local deployment of the operator? e.g. extra documentation and Makefile changes.

PR #321, opened 2020-03-25
Adds random requeue time to credential refresh.

iamkirkbater:
Adds a fuzzing time factor to, over time, randomly distribute credential refreshes.

To test:
* `go test -v github.com/openshift/aws-account-operator/pkg/credentialwatcher`
* Create Account CR or two or Three
* Wait for Credentials to Refresh
* Compare Timestamps, they should no longer be within 59-60 minutes of when they were created or last rotated but within 45-55 minutes, or for the Console URL it should be within 12-15 minutes and not 14-15 minutes.

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -3,7 +3,6 @@ package account
 import (
 	"context"
 	"fmt"
-	"math/rand"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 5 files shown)


iamkirkbater:
Removed the WIP label because the "weird" behavior that I was seeing where it was refreshing the credentials every minute was happening on master as well, and was not introduced on this branch.

clcollins:
- Validated functionality in the PR
- Tested awsfederatedaccountaccess and ccs/byoc creation

/LGTM

PR #322, opened 2020-03-30
Ccs make target

jharrington22:
This PR adds a make target `make test-ccs` that will test the CCS workflow as follows:

1. Create CCS namespace
2. Create IAM user osdCcsAdmin if it doesn't exist or rotate the IAM Access Keys if it does
3. Create CCS account claim with an external to the AWS organization account ID (osd-staging-2) new EMV OSD_STAGING_2_AWS_ACCOUNT_ID
4. Wait for CCS claim to become ready
5. Delete what we created


Makefile
```diff
@@ -92,3 +95,47 @@ delete-awsfederatedaccountaccess: check-aws-account-id-env
...
+.PHONY: delete-ccs-secret
+delete-ccs-secret:
+	# Create CCS Secret
+	@oc delete -f ccs-secret.yaml
+	@rm ccs-secret.yaml
+
+.PHONY: create-ccs-accountclaim
+create-ccs-accountclaim:
+	# Create ccs accountclaim
+	@oc process -p CCS_ACCOUNT_ID=${OSD_STAGING_2_AWS_ACCOUNT_ID} -p NAME=${CCS_CLAIM_NAME} -p NAMESPACE=${CCS_NAMESPACE_NAME} -f hack/templates/aws_v1alpha1_ccs_accountclaim_cr.tmpl | oc apply -f -
```
clcollins: I don't think the Account Claim CR template made it into your commit.

```
error: failed to read input object (not a Template?): the path "hack/templates/aws_v1alpha1_ccs_accountclaim_cr.tmpl" does not exist
```

Makefile
```diff
@@ -92,3 +95,47 @@ delete-awsfederatedaccountaccess: check-aws-account-id-env
...
+	@while true; do STATUS=$$(oc get accountclaim ${CCS_CLAIM_NAME} -n ${CCS_NAMESPACE_NAME} -o json | jq -r '.status.state'); if [ "$$STATUS" == "Ready" ]; then break; elif [ "$$STATUS" == "Failed" ]; then echo "Account claim ${CCS_CLAIM_NAME} failed to create"; exit 1; fi; sleep 1; done
+
+.PHONY: delete-ccs-accountclaim
+delete-ccs-accountclaim:
+	# Delete ccs accountclaim
+	@oc process -p CCS_ACCOUNT_ID=${OSD_STAGING_2_AWS_ACCOUNT_ID} -p NAME=${CCS_CLAIM_NAME} -p NAMESPACE=${CCS_NAMESPACE_NAME} -f hack/templates/aws_v1alpha1_ccs_accountclaim_cr.tmpl | oc delete -f -
+
+# Test CCS
+.PHONY: test-ccs
+test-ccs: create-ccs-namespace create-ccs-secret create-ccs-accountclaim delete-ccs-accountclaim delete-ccs-secret delete-ccs-namespace
```
clcollins: This would benefit from using the `check-aws-account-id-env` target like we do for `test-awsfederatedaccountacccess`, to validate that the `OSD_STAGING_2_AWS_ACCOUNT_ID` variable is set before attempting to create things.

Makefile
```diff
@@ -92,3 +95,47 @@ delete-awsfederatedaccountaccess: check-aws-account-id-env
...
+.PHONY: delete-ccs-namespace
+delete-ccs-namespace:
+	# Create namespace
+	@oc process -p NAME=${CCS_NAMESPACE_NAME} -f hack/templates/namespace.tmpl | oc delete -f -
+
+# Create CCS (BYOC) Secret
+.PHONY: create-ccs-secret
+create-ccs-secret:
+	# Create CCS Secret
+	./hack/scripts/aws/rotate_iam_access_keys.sh -p osd-staging-2 -u osdCcsAdmin -a ${OSD_STAGING_2_AWS_ACCOUNT_ID} -o ccs-secret.yaml
```
clcollins: `ccs-secret.yaml` ends up being created in the root of the code base, and stays there if the make target fails before getting to the `delete-ccs-secret` target for whatever reason.  This could accidentally lead to credentials committed to the repository if someone is not expecting the file to be there, or not paying attention.

Maybe the creds could be written from STDIN to avoid a file being written?  Or they could be placed in /tmp, so they're cleaned up on reboot if there's a deletion failure?  Or, better, private tmp, on the (slim) chance that the user is on a shared machine.

Makefile
jharrington22: :man_facepalming: thanks.

Makefile
jharrington22: Yeap already doing this https://github.com/openshift/aws-account-operator/pull/322/files#diff-b67911656ef5d18c4ae36cb6741b7965R28

jharrington22: I'll just reuse that same check, good catch.

Makefile
jharrington22: Good catch, I did think that these should go to stdout I will do that its the right thing to do.

jharrington22:
@clcollins I added old so it doesn't auto merge but your comments have been addressed if you could take another look please. Thanks.

jharrington22:
@fahlmant @clcollins @iamkirkbater have you been able to review this?

PR #324, opened 2020-04-01
Update gopkg lock to remove unneeded packages

fahlmant:
There are a few extra deps in Gopkg.lock. Ran `dep ensure` to clean them up.


ArielLima:
The only test I performed creating an account.
/lgtm

PR #328, opened 2020-04-07
Add test-all make target

jharrington22:
This PR adds a make target that should run the following tests/targets

- CCS
- Reuse
- Account Creation
- Federated Access


jharrington22:
@clcollins looks good now after #330 

```
make test-all  52.19s user 7.60s system 19% cpu 4:59.54 total 
```

jharrington22:
@clcollins going to need another :bacon: I just removed the hold.

PR #329, opened 2020-04-08
Set IAM_USER var to create access keys

jharrington22:
This PR fixes a bug where the IAM_USER var isn't set if the IAM user didn't exist and is created by the script before the access keys are created/rotated.


jharrington22:
cc/ @2uasimojo @dustman9000

PR #330, opened 2020-04-08
Test scripts

jharrington22:
This PR removes duplicated make targets and moves them into scripts. It also introduces a test directory where we will later add Go integration tests replacing these bash scripts.

### DIFF
--- a/test/integration/api/delete_account_secrets.sh
+++ b/test/integration/api/delete_account_secrets.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+# Load Environment vars
+source hack/scripts/test_envs 
+
+if [ -z "$OSD_STAGING_1_ACCOUNT_CR_NAME_OSD" ]; then
+    "OSD_STAGING_1_ACCOUNT_CR_NAME_OSD not set"
+    exit 1
+fi
+
+for secret in $(oc get secrets -n "${NAMESPACE}" | awk "/${OSD_STAGING_1_ACCOUNT_CR_NAME_OSD}/"'{ print $1 }')
+do
+    oc delete secret $secret -n ${NAMESPACE}
+done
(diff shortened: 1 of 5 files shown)


test/integration/api/delete_account_secrets.sh
```diff
@@ -0,0 +1,14 @@
...
+
+# Load Environment vars
+source hack/scripts/test_envs 
+
+if [ -z "$OSD_STAGING_1_ACCOUNT_CR_NAME_OSD" ]; then
+    "OSD_STAGING_1_ACCOUNT_CR_NAME_OSD not set"
+    exit 1
+fi
+
+for secret in $(oc get secrets -n "${NAMESPACE}" | grep ${OSD_STAGING_1_ACCOUNT_CR_NAME_OSD} | awk '{ print $1 }')
```
clcollins, line 11:
```suggestion
for secret in $(oc get secrets -n "${NAMESPACE}" | awk '/${OSD_STAGING_1_ACCOUNT_CR_NAME_OSD}/ { print $1 }')
```
Removes unnecessary `grep`.  Just a suggestion - not a big deal.  I think just a second ago, you were not using the `grep`, but had a duplicated $ in the `{print $$1}` command.  I was going to suggest removing that but Github said the diff was out of date.

jharrington22: @clcollins Ha so I removed the second `$` and it wasn't working, I converted the quotes to `"` but still no love. Is your suggestion tested and working? Happy to do it if so! I was trying to do the right thing.

clcollins: Ah, you're right, I made a mistake in the suggestion.  This works:

`for secret in $(oc get secrets -n "${NAMESPACE}" | awk "/${OSD_STAGING_1_ACCOUNT_CR_NAME_OSD}/"'{ print $1 }')`

I think the second $ was actually breaking the awk printing, and printing the whole line, so it was deleting the secret by name, but also trying to delete secrets with names based on the other columns - working in spite of itself, heh.

jharrington22: Thanks @clcollins updated the PR

PR #335, opened 2020-04-16
Add label for the target awsAccountID where the role is created

fahlmant:
Currently, there is no direct reference to the target AWS account for an AWS Federated Account Access. Instead, it relies on using the `aws` secret in the namespace of the accountclaim to call `aws sts get-caller-identity`. This causes an issue if the aws secret is deleted before the finalizer code is run.

This PR fixes this by adding a label to the CR with the account ID. It still relies on the secret to get the account ID, but only needs to happen once. From then on, the ID can be referenced in the CR by the label. 

This also includes a new function to check an arbitrary label on the CR, rather than just the `uid` label.

### DIFF
--- a/pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
+++ b/pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
@@ -159,11 +159,11 @@ func (r *ReconcileAWSFederatedAccountAccess) Reconcile(request reconcile.Request
 	}
 
 	// Check if the FAA has the uid label
-	if !hasUIDLabel(currentFAA) {
+	if !hasLabel(currentFAA, awsv1alpha1.UIDLabel) {
 		// Generate a new UID
 		uid := rand.String(6)
 
-		reqLogger.Info(fmt.Sprintf("Adding UID %s to AccountAcess %s", uid, currentFAA.Name))
+		reqLogger.Info(fmt.Sprintf("Adding UID %s to AccountAccess %s", uid, currentFAA.Name))
 		newLabel := map[string]string{"uid": uid}
 
 		// Join the new UID label with any current labels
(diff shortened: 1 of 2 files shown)


pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -159,19 +159,13 @@ func (r *ReconcileAWSFederatedAccountAccess) Reconcile(request reconcile.Request
 	}
 
 	// Check if the FAA has the uid label
-	if !hasUIDLabel(currentFAA) {
+	if !hasLabel(currentFAA, "uid") {
```
jharrington22, line 162: Can we make the `"uid"` a var in `shared_types.go` for v1alpha1?

fahlmant: :+1:

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -213,6 +207,22 @@ func (r *ReconcileAWSFederatedAccountAccess) Reconcile(request reconcile.Request
 		return reconcile.Result{}, err
 	}
 
+	accountID := *gciOut.Account // Add requested aws managed policies to the role
+
+	if !hasLabel(currentFAA, awsv1alpha1.AccountIDLabel) {
+
+		reqLogger.Info(fmt.Sprintf("Adding awsAccountID %s to AccountAcess %s", accountID, currentFAA.Name))
```
iamkirkbater, line 214: Typo in AccountAcess

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -213,6 +213,28 @@ func (r *ReconcileAWSFederatedAccountAccess) Reconcile(request reconcile.Request
...
+		if currentFAA.Labels != nil {
+			currentFAA.Labels = joinLabelMaps(currentFAA.Labels, newLabel)
+		} else {
+			currentFAA.Labels = newLabel
+		}
+
+		// Update the CR with new labels
+		err = r.client.Update(context.TODO(), currentFAA)
+		if err != nil {
+			reqLogger.Error(err, fmt.Sprintf("Lable update for %s failed", currentFAA.Name))
```
iamkirkbater, line 233: Lable -> Label

PR #336, opened 2020-04-17
Fix example CRs for Account and FederatedAccountAccess

fahlmant:
This PR takes out specific user information from the AWS Federated Account Access example CR, as well as fills out a basic spec for the Account example CR.

### DIFF
--- a/deploy/crds/aws_v1alpha1_awsfederatedaccountaccess_cr.yaml
+++ b/deploy/crds/aws_v1alpha1_awsfederatedaccountaccess_cr.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: aws-account-operator
 spec:
   awsCustomerCredentialSecret: 
-    name: osd-creds-mgmt-w2tclb-secret
-    namespace: aws-account-operator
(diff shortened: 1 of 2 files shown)


deploy/crds/aws_v1alpha1_awsfederatedaccountaccess_cr.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   namespace: aws-account-operator
 spec:
   awsCustomerCredentialSecret: 
-    name: osd-creds-mgmt-w2tclb-secret
-    namespace: aws-account-operator
-  externalCustomerAWSIAMARN: arn:aws:iam::414830704956:user/james.harrington
+    name: secret-name
+    namespace: example-namespace
+  externalCustomerAWSIAMARN: arn:aws:iam::{$AWS_ACCOUNT_ID}:user/{$AWS_IAM_USER}
```
jharrington22, line 10: Are these meant to be templatable? either way `${AWS_ACCOUNT_ID}` seems better?

fahlmant: Not intended specifically to be templateable. Just to indicate where to place what

PR #340, opened 2020-04-22
Added functionality for easily generating tag types

ArielLima:
While I was working on a ticket regarding tagging aws resources the operator initializes on account creation. @jharrington22  noticed that different aws resources use different tag types with identical values. This PR gives us a way for easily creating tags for different resources


jharrington22:
@ArielLima I opened https://github.com/openshift/aws-account-operator/pull/341 which would add a similar functionality to the awsclient and supports multiple tags and initalization.

ArielLima:
Closing in favour of better #341

PR #343, opened 2020-04-23
Multipe ccs cluster support

jharrington22:


pkg/apis/aws/v1alpha1/shared_types.go
```diff
@@ -85,3 +85,9 @@ var ErrCreateEC2Instance = errors.New("EC2CreationTimeout")
 
 // ErrFailedAWSTypecast indicates that there was a failure while typecasting to aws error
 var ErrFailedAWSTypecast = errors.New("FailedToTypecastAWSError")
+
+// IAMUserIDLabel label key for IAM user suffix
+var IAMUserIDLabel = "iam_user_id"
```
fahlmant, line 90: I suggest`iamUserID` to keep label names consistent with k8s labels.

pkg/awsclient/client.go
```diff
@@ -66,6 +66,9 @@ type Client interface {
 	DeleteUser(*iam.DeleteUserInput) (*iam.DeleteUserOutput, error)
 	DeleteUserPolicy(*iam.DeleteUserPolicyInput) (*iam.DeleteUserPolicyOutput, error)
 	GetUser(*iam.GetUserInput) (*iam.GetUserOutput, error)
+	ListUsers(*iam.ListUsersInput) (*iam.ListUsersOutput, error)
+	ListUsersPages(*iam.ListUsersInput, func(*iam.ListUsersOutput, bool) bool) error
```
fahlmant, line 70: Where is this function defined? I'm not seeing it in the SDK but I maybe be blind.

fahlmant: https://docs.aws.amazon.com/sdk-for-go/api/service/iam/#IAM.ListUsersPages

pkg/controller/utils/utils.go
```diff
@@ -128,3 +107,36 @@ func Remove(list []string, s string) []string {
 	}
 	return list
 }
+
+// GenerateShortUUID Generates a short UID
+func GenerateShortUUID() string {
```
fahlmant, line 112: Semantics: we called a 6 char string a  UID in the AWS Federated Account Access controller and that's what the label uses. Thoughts on changing all UUID to UID to match and then the AFAA controller can use this as well? 

https://github.com/openshift/aws-account-operator/blob/master/pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go#L163-L164

pkg/controller/account/account_controller.go
```diff
@@ -364,10 +366,13 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 	if accountReadyForInitialization(currentAcctInstance) {
 
 		reqLogger.Info(fmt.Sprintf("Initalizing account: %s AWS ID: %s", currentAcctInstance.Name, currentAcctInstance.Spec.AwsAccountID))
-		//var awsAssumedRoleClient awsclient.Client
 		var roleToAssume string
 
 		if accountIsBYOC(currentAcctInstance) {
+			// Use the same ID applied to the account name for IAM usernames
+			currentAccInstanceID := currentAcctInstance.Labels[awsv1alpha1.IAMUserIDLabel]
```
fahlmant, line 373: This label should be unwrapped/checked before trying to use it for safety.

pkg/controller/accountclaim/reuse.go
```diff
@@ -32,6 +34,45 @@ func (r *ReconcileAccountClaim) finalizeAccountClaim(reqLogger logr.Logger, acco
...
+		connErr := fmt.Sprintf("Unable to create aws client for region %s", clusterAwsRegion)
+		reqLogger.Error(err, connErr)
+		return err
+	}
+
+	// Remove IAM user we'll remove the IAM user for CCS
+	if utils.AccountCRHasIAMUserIDLabel(reusedAccount) && accountClaim.Spec.BYOC {
+		err = r.cleanUpIAM(reqLogger, awsClient, reusedAccount, accountClaim)
+		if err != nil {
+			reqLogger.Error(err, "Failed to delete IAM user during finalizer cleanup")
```
rogbas, line 70: Do we want to return from the reconcile loop with error in case we fall here?

pkg/controller/account/iam.go
```diff
@@ -596,13 +467,13 @@ func (r *ReconcileAccount) BuildIAMUser(reqLogger logr.Logger, awsClient awsclie
 	secretName := account.Name
 
 	// Append to secret name if we're if its the SRE admin user secret
-	if iamUserName == iamUserNameSRE {
-		secretName = account.Name + "-" + strings.ToLower(iamUserName)
+	if strings.Contains(iamUserNameSRE, iamUserName) {
+		secretName = account.Name + "-" + strings.ToLower("osdManagedAdminSRE")
```
fahlmant, line 471: I think we should still have the base names `osdManagedAdmin` and `osdManagedAdminSRE` in the `shared_types` file.

pkg/controller/utils/utils.go
jharrington22: Not sure why I didn't do that as I was looking at the AFAA, will fix.

pkg/controller/account/iam.go
jharrington22: So I'm not doing that here because of the way the secret is created, I don't want the `osd-creds-mgmt-asdfx-osdmanagedadminsre-secret` for instance to look like `osd-creds-mgmt-asdfx-osdmanagedadminsre-asdfx-secret`. We would create a card to decouple the secret creation process here?

fahlmant: I'm saying just make
var OsdManagedAdminSRE = "osdManagedAdminSRE" and use that.

jharrington22:
This PR works with https://github.com/openshift/aws-account-operator/pull/351 which will be need to be merged before this. Cherry-picking the commit from #351 and running `make test-all` succeeds. Credential rotate fine also.

rogbas:
@rogbas @fahlmant can you please review and leave feedback?

jharrington22:
@clcollins @rogbas please review.

pkg/awsclient/iam.go
```diff
@@ -0,0 +1,179 @@
...
+
+// checkIAMUserExists checks if a given IAM user exists within an account
+// Takes a logger, an AWS client for the target account, and a target IAM username
+func CheckIAMUserExists(reqLogger logr.Logger, client Client, userName string) (bool, *iam.GetUserOutput, error) {
+	// Retry when getting IAM user information
+	// Sometimes we see a delay before credentials are ready to be user resulting in the AWS API returning 404's
+	var iamGetUserOutput *iam.GetUserOutput
+	var err error
+
+	attempt := 1
```
ArielLima, line 87: Do we really need a second variable for this? It doesn't seem like `attempt` is used in any way that `i` couldn't be used.

pkg/awsclient/iam.go
```diff
@@ -0,0 +1,179 @@
...
+	return true, iamGetUserOutput, nil
+}
+
+// CreateIAMUser creates a new IAM user in the target AWS account
+// Takes a logger, an AWS client for the target account, and the desired IAM username
+func CreateIAMUser(reqLogger logr.Logger, client Client, account *awsv1alpha1.Account, userName string) (*iam.CreateUserOutput, error) {
+	var createUserOutput = &iam.CreateUserOutput{}
+	var err error
+
+	attempt := 1
```
ArielLima, line 137: same as above.

pkg/controller/account/generate.go
```diff
@@ -0,0 +1,38 @@
...
+
+	uuid := utils.GenerateShortUID()
+
+	accountName := GenerateAccountCRName(uuid)
+
+	return &awsv1alpha1.Account{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      accountName,
+			Namespace: namespace,
+			Labels: utils.GenerateLabel(awsv1alpha1.IAMUserIDLabel, uuid),
```
rogbas, line 23: Wouldn't we want to label the Account CR and also suffix the IAM user with the full account name (emailID + uuid) instead of just the uuid?

That way if we start supporting multiple email IDs we're covered.

pkg/controller/accountclaim/reuse.go
jharrington22: No I don't think we want to fail the reuse for this since its CCS, these users are unique so leaving them behind isn't a problem

Makefile
```diff
@@ -145,24 +161,50 @@ create-ccs-secret:
 	# Wait for AWS to propogate IAM credentials
 	sleep ${SLEEP_INTERVAL}
 
+# Create CCS (BYOC) Secret
+.PHONY: create-ccs-2-secret
+create-ccs-2-secret:
+	# Create CCS Secret
+	./hack/scripts/aws/rotate_iam_access_keys_2.sh -p osd-staging-2 -u osdCcsAdmin -a ${OSD_STAGING_2_AWS_ACCOUNT_ID} -o /dev/stdout | oc apply -f -
```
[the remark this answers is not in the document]
jharrington22: Fixed.

pkg/controller/account/generate.go
jharrington22: The UID is the same as whats in the account name which is emailID + UID

PR #346, opened 2020-04-24
Use root account creds to assume a role in target account for federated role cleanup

fahlmant:
This PR moves the AWS Federated Account Access finalizer cleanup to use the root account (aws-account-operator-credentials) access keys to assume the `OrganizationAccountAccessRole` in the target AWS account rather than the secret specified in the CR. This solves an issue where the secret is deleted and the AFAA finalizer can't complete properly and leaves extra Federated Roles.

### DIFF
--- a/pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
+++ b/pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
@@ -558,22 +558,51 @@ func (r *ReconcileAWSFederatedAccountAccess) removeFinalizer(reqLogger logr.Logg
 func (r *ReconcileAWSFederatedAccountAccess) cleanFederatedRoles(reqLogger logr.Logger, currentFAA *awsv1alpha1.AWSFederatedAccountAccess, federatedRoleCR *awsv1alpha1.AWSFederatedRole) error {
 
 	// Get the UID
-	uidLabel, ok := currentFAA.Labels["uid"]
+	uidLabel, ok := currentFAA.Labels[awsv1alpha1.UIDLabel]
 	if !ok {
 		labelErr := fmt.Sprintf("Unable to get UID label")
 		return errors.New(labelErr)
 	}
 
+	// Get the AWS Account ID
+	accountIDLabel, ok := currentFAA.Labels[awsv1alpha1.AccountIDLabel]
+	if !ok {
+		labelErr := fmt.Sprintf("Unable to get AWS Account ID label")
+		return errors.New(labelErr)
+	}
+
 	roleName := currentFAA.Spec.AWSFederatedRole.Name + "-" + uidLabel
 
-	// Build AWS client
-	awsClient, err := awsclient.GetAWSClient(r.client, awsclient.NewAwsClientInput{
-		SecretName: currentFAA.Spec.AWSCustomerCredentialSecret.Name,
-		NameSpace:  currentFAA.Spec.AWSCustomerCredentialSecret.Namespace,
+	// Build AWS client from root secret
+	rootAwsClient, err := awsclient.GetAWSClient(r.client, awsclient.NewAwsClientInput{
+		SecretName: "aws-account-operator-credentials",
+		NameSpace:  awsv1alpha1.AccountCrNamespace,
 		AwsRegion:  "us-east-1",
 	})
 	if err != nil {
-		awsClientErr := fmt.Sprintf("Unable to create aws client for region ")
+		awsClientErr := fmt.Sprintf("Unable to create root aws client for region ")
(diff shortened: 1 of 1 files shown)


pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -558,22 +558,47 @@ func (r *ReconcileAWSFederatedAccountAccess) removeFinalizer(reqLogger logr.Logg
...
 	// Get the UID
-	uidLabel, ok := currentFAA.Labels["uid"]
+	uidLabel, ok := currentFAA.Labels[awsv1alpha1.UIDLabel]
 	if !ok {
 		labelErr := fmt.Sprintf("Unable to get UID label")
 		return errors.New(labelErr)
 	}
 
+	// Get the AWS Account ID
+	accountIDLabel, ok := currentFAA.Labels[awsv1alpha1.AccountIDLabel]
```
fahlmant, line 568: Need to handle the `ok` block

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -558,22 +558,51 @@ func (r *ReconcileAWSFederatedAccountAccess) removeFinalizer(reqLogger logr.Logg
...
+	}
+
+	awsClient, err := awsclient.GetAWSClient(r.client, awsclient.NewAwsClientInput{
+		AwsCredsSecretIDKey:     *assumeRoleOutput.Credentials.AccessKeyId,
+		AwsCredsSecretAccessKey: *assumeRoleOutput.Credentials.SecretAccessKey,
+		AwsToken:                *assumeRoleOutput.Credentials.SessionToken,
+		AwsRegion:               "us-east-1",
+	})
+	if err != nil {
+		awsClientErr := fmt.Sprintf("Unable to create aws client for target linked account in region ")
```
iamkirkbater, line 605: Using `fmt.Sprintf` here but we're not adding any variables to be displayed in the text.  Is this intended?

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -558,22 +558,51 @@ func (r *ReconcileAWSFederatedAccountAccess) removeFinalizer(reqLogger logr.Logg
...
-		NameSpace:  currentFAA.Spec.AWSCustomerCredentialSecret.Namespace,
+	// Build AWS client from root secret
+	rootAwsClient, err := awsclient.GetAWSClient(r.client, awsclient.NewAwsClientInput{
+		SecretName: "aws-account-operator-credentials",
+		NameSpace:  awsv1alpha1.AccountCrNamespace,
 		AwsRegion:  "us-east-1",
 	})
 	if err != nil {
-		awsClientErr := fmt.Sprintf("Unable to create aws client for region ")
+		awsClientErr := fmt.Sprintf("Unable to create root aws client for region ")
```
iamkirkbater, line 583: Using fmt.Sprintf here but we're not adding any variables to be displayed in the text. Is this intended?

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -558,22 +558,51 @@ func (r *ReconcileAWSFederatedAccountAccess) removeFinalizer(reqLogger logr.Logg
...
+	}
+
 	roleName := currentFAA.Spec.AWSFederatedRole.Name + "-" + uidLabel
 
-	// Build AWS client
-	awsClient, err := awsclient.GetAWSClient(r.client, awsclient.NewAwsClientInput{
-		SecretName: currentFAA.Spec.AWSCustomerCredentialSecret.Name,
-		NameSpace:  currentFAA.Spec.AWSCustomerCredentialSecret.Namespace,
+	// Build AWS client from root secret
+	rootAwsClient, err := awsclient.GetAWSClient(r.client, awsclient.NewAwsClientInput{
```
clcollins, line 577: This doesn't affect @ArielLima's work on the awsclient refactor, correct?  Since it's a different client with different credentials?

fahlmant: This will need to be changed after the AWS Client refactor PR is merged I think

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
fahlmant: Yeah, I just coppied what we do with other errors in case we need to add in vars late

PR #347, opened 2020-04-27
Update rotate credentials bool only if false

jharrington22:
This PR ensures the credential watcher is only updating the Account CR status object if the credentials bool its updating is false.

### DIFF
--- a/pkg/credentialwatcher/secretwatcher.go
+++ b/pkg/credentialwatcher/secretwatcher.go
@@ -139,20 +139,24 @@ func (s *secretWatcher) updateAccountRotateCredentialsStatus(log logr.Logger, ac
 		return
 	}
 
-	if accountInstance.Status.RotateCredentials != true {
+	if (accountInstance.Status.RotateConsoleCredentials && credentialType == "console") || (accountInstance.Status.RotateCredentials && credentialType == "cli") {
+		// We don't want to update the status since its already been updated
+		return
+	}
 
-		if credentialType == "console" {
-			accountInstance.Status.RotateConsoleCredentials = true
-			log.Info(fmt.Sprintf("AWS console credentials secret was created ago requeueing to be refreshed"))
-		} else if credentialType == "cli" {
-			accountInstance.Status.RotateCredentials = true
-			log.Info(fmt.Sprintf("AWS cli credentials secret was created ago requeueing to be refreshed"))
-		}
+	if !accountInstance.Status.RotateConsoleCredentials && credentialType == "console" {
+		accountInstance.Status.RotateConsoleCredentials = true
+		log.Info(fmt.Sprintf("%s %s credentials secret to be refreshed", accountName, credentialType))
+	}
 
-		err = s.UpdateAccount(accountInstance)
-		if err != nil {
-			log.Error(err, fmt.Sprintf("Error updating account %s", accountName))
-		}
+	if !accountInstance.Status.RotateCredentials && credentialType == "cli" {
+		accountInstance.Status.RotateCredentials = true
+		log.Info(fmt.Sprintf("%s %s credentials secret to be refreshed", accountName, credentialType))
+	}
+
+	err = s.UpdateAccount(accountInstance)
+	if err != nil {
+		log.Error(err, fmt.Sprintf("Unable to update status to trigger account %s to rotate %s credentials", accountName, credentialType))
 	}
 }
 


pkg/credentialwatcher/secretwatcher.go
```diff
@@ -139,20 +139,24 @@ func (s *secretWatcher) updateAccountRotateCredentialsStatus(log logr.Logger, ac
...
-	if accountInstance.Status.RotateCredentials != true {
-
-		if credentialType == "console" {
-			accountInstance.Status.RotateConsoleCredentials = true
-			log.Info(fmt.Sprintf("AWS console credentials secret was created ago requeueing to be refreshed"))
-		} else if credentialType == "cli" {
-			accountInstance.Status.RotateCredentials = true
-			log.Info(fmt.Sprintf("AWS cli credentials secret was created ago requeueing to be refreshed"))
-		}
+	if (accountInstance.Status.RotateConsoleCredentials && credentialType == "console") || (accountInstance.Status.RotateCredentials && credentialType == "cli") {
```
iamkirkbater, line 142: Can we simplify this into functions similar to what @clcollins did?  It took me a bit of time to parse what we were checking for here.

pkg/credentialwatcher/secretwatcher.go
```diff
@@ -139,20 +139,24 @@ func (s *secretWatcher) updateAccountRotateCredentialsStatus(log logr.Logger, ac
...
+		return
+	}
 
-		err = s.UpdateAccount(accountInstance)
-		if err != nil {
-			log.Error(err, fmt.Sprintf("Error updating account %s", accountName))
-		}
+	if !accountInstance.Status.RotateConsoleCredentials && credentialType == "console" {
+		accountInstance.Status.RotateConsoleCredentials = true
+		log.Info(fmt.Sprintf("AWS console credentials secret was created ago requeueing to be refreshed"))
```
iamkirkbater, line 149: Would it be helpful here to put account name that we're requeueing so we can grep the logs better?

pkg/credentialwatcher/secretwatcher.go
```diff
@@ -139,20 +139,24 @@ func (s *secretWatcher) updateAccountRotateCredentialsStatus(log logr.Logger, ac
...
+	}
+
+	if !accountInstance.Status.RotateCredentials && credentialType == "cli"  {
+		accountInstance.Status.RotateCredentials = true
+		log.Info(fmt.Sprintf("AWS cli credentials secret was created ago requeueing to be refreshed"))
```
iamkirkbater, line 154: Would it be helpful here to put account name that we're requeueing so we can grep the logs better?

pkg/credentialwatcher/secretwatcher.go
iamkirkbater: We'll open a new card for this refactor so we can add tests and track appropriately.

pkg/credentialwatcher/secretwatcher.go
```diff
@@ -139,20 +139,24 @@ func (s *secretWatcher) updateAccountRotateCredentialsStatus(log logr.Logger, ac
 		return
 	}
 
-	if accountInstance.Status.RotateCredentials != true {
```
2uasimojo, line 142: If I'm understanding the previous code correctly, there was a hole where, if the CLI creds had recently been rotated, we would miss rotating the console creds completely. And this code change makes it so we're checking both of them explicitly. Do I have that right?

pkg/credentialwatcher/secretwatcher.go
2uasimojo: +1. (This message doesn't make a lot of sense as written anyway.)

pkg/credentialwatcher/secretwatcher.go
```diff
@@ -139,20 +139,24 @@ func (s *secretWatcher) updateAccountRotateCredentialsStatus(log logr.Logger, ac
...
-		err = s.UpdateAccount(accountInstance)
-		if err != nil {
-			log.Error(err, fmt.Sprintf("Error updating account %s", accountName))
-		}
+	if !accountInstance.Status.RotateConsoleCredentials && credentialType == "console" {
+		accountInstance.Status.RotateConsoleCredentials = true
+		log.Info(fmt.Sprintf("AWS console credentials secret was created ago requeueing to be refreshed"))
+	}
+
+	if !accountInstance.Status.RotateCredentials && credentialType == "cli"  {
```
2uasimojo, line 152: Oh, you have a gofmt fail here.
```suggestion
	if !accountInstance.Status.RotateCredentials && credentialType == "cli" {
```

pkg/credentialwatcher/secretwatcher.go
jharrington22: We now check both explicitly but we also do not update the Account CR's status sub resource if its already the desired value `true` since this would cause an increment in resource version and an error in the reconciler.

PR #350, opened 2020-04-29
Fix Double Logging for Secrets Refresh

iamkirkbater:
Since this log message is logging twice, let's just remove it because it's not really adding much value since the account isn't ready anyway, and with failed accounts this is just causing a bunch of noise in the logs.


2uasimojo:
I got pretty excited when I saw the commit message -- but the PR looks like it only has one commit that's a trivial logging change. Did the substance get lost somewhere?

[Later] Yeah, it looks like this PR is pulling the wrong branch: `double-logging` instead of `secrets-refresh-random`.

iamkirkbater:
@2uasimojo Yeah, I tried using `hub` to create the PR from my command line and I thought I cancelled it but it looks like it just added a previous PR message that I had created.

clcollins:
/lgtm

I'm in favor of removing that log line entirely

PR #351, opened 2020-04-30
Refactor BuildIAMUser

jharrington22:
This PR is a first pass at refactoring the BuildIAMUser function. It removes the logger from outer functions where possible and cleans up the logic within the function by moving out as much as possible into smaller functions. A couple of TODO's but this makes it much easier to read.

I think that we should move the logger completely from BuildIAMUser and return any errors out to the reconcile and parse there. I didn't do that here because of 1. time and 2. PR size.

All errors are returned to the reconcile function which will then fail the account if there is a problem.

`make test-all` is passing locally and I've verified that the AWS credentials are working.


jharrington22:
This needs to be merged before https://github.com/openshift/aws-account-operator/pull/343 as I just confirmed it fixes issues that I was having on that branch with secret names and creation (I cherry-picked this commit to test). Everything works now.

ArielLima:
Passed locally for me as well with `make test-all`.

PR #354, opened 2020-04-30
add resource limits/requests for the operator per OSD-2109

bmeng:
Current usage on both stg and prod

STG:
aws-account-operator                           aws-account-operator-7578d56b5-l7cqg             45m          158Mi

PROD:
aws-account-operator                              aws-account-operator-647d5b8d7c-h5v2x              17m          112Mi

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -21,6 +21,13 @@ spec:
           command:
           - aws-account-operator
           imagePullPolicy: Always
+          resources:
+            requests:
+              memory: "200Mi"
+              cpu: "50m"
+            limits:
+              memroy: "400Mi"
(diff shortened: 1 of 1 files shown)


rogbas:
@rogbas to create a spike for eval resource consumption by the operator, to confirm the proposed requests/limits are OK in the long term

jharrington22:
@bmeng how did you compute these limits? Were you able to look at the CPU usage overtime? Secret renewal is taking longer than usual and I'm wondering if this is causing issues. cc/ @rogbas

bmeng:
I just checked the realtime usage of the operator on both stg and prod env. Did not watch them overtime

PR #355, opened 2020-04-30
Add additional permissions for network-mgmt role

maorfr:
This PR adds additional permissions required to provision resources to support VPC peerings.

It also removes some duplicates.

### DIFF
--- a/deploy/crds/aws_v1alpha1_awsfederatedrole_networkmgmt_cr.yaml
+++ b/deploy/crds/aws_v1alpha1_awsfederatedrole_networkmgmt_cr.yaml
@@ -14,10 +14,13 @@ spec:
     awsStatements:
       - effect: Allow
         action:
+          - "ec2:AssociateRouteTable"
+          - "ec2:AttachInternetGateway"
           - "ec2:AcceptVpcPeeringConnection"
           - "ec2:AttachVpnGateway"
           - "ec2:CreateCustomerGateway"
           - "ec2:CreateRoute"
+          - "ec2:CreateRouteTable"
           - "ec2:CreateTags"
           - "ec2:CreateVPCEndpoint"
           - "ec2:CreateVpcPeeringConnection"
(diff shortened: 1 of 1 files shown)


rogbas:
@rogbas please ping Will to validate the added permissions and confirm it's compatible with the Network  Manager supported use cases.

mwoodson:
This PR looks like it lets the user ADD ROUTE TABLES, not just routes.    I have a concern about this.

We currently allow network admins to create routes, but not delete or modify routes.   Does this need to change?

Adding a whole new route table seems risky here.  It could hold up cluster deletions if not tagged appropriately.  We would also have to assign subnets to routestables.  I'm unclear if this does anything.

I think there is an issue we have now is how do we let the customer manage (modify/delete) the routes.  By messing with the default route table, the customer can mess up the cluster and things can be broken.  It would be on the customer.

Looking for opinions, is adding/creating route tables the best way to go here?

maorfr:
when it comes to automation, using the existing route tables is not very easy to do.
if using terraform, one would have to `terraform import` the existing route table and use it.

the approach of creating a new route table can be automated very easily by customers.

if you describe the required tagging, I can try to shape the policy accordingly. that way, a new route table can only be created if it is according to these rules.

jfchevrette:
It seems that AWS recommends using custom route tables rather than the main table when adding subnets so as to not risk breaking the main routing table.

https://docs.aws.amazon.com/vpc/latest/userguide/VPC_Route_Tables.html#CustomRouteTables

OTOH the aws docs do state that a VPC cannot be deleted if associated resources (such as custom route tables) are not deleted first. This could indeed create a situation where the cluster aws resources are not properly deleted if any custom route tables assigned to the VPC are not deleted first.

https://aws.amazon.com/premiumsupport/knowledge-center/troubleshoot-dependency-error-delete-vpc/

iamkirkbater:
Could we do the opposite here and explicitly tag _our_ route tables and then explicitly DENY the network admin role access to those tables?  Then we give our customers the flexibility to modify their network however they see fit, without being able to mess with the OSD provisioned stuff?

fahlmant:
Discussing next week

PR #357, opened 2020-05-02
CreateIAMUser panic resolution

dofinn:
745ef75 Enables the CreateIAMUser the possibilty to return a null
pointer even when the IAM user is successfully created. It appears that
the for loop was not returned in the correct position.


pkg/controller/account/iam.go
```diff
@@ -493,6 +493,7 @@ func CreateIAMUser(reqLogger logr.Logger, client awsclient.Client, userName stri
 				return &iam.CreateUserOutput{}, err
 			}
 		}
+		break
```
iamkirkbater, line 496: If we break here without adding an explicit continue in the first aerr case above this will break out of the for loop every time. So we can either add a continue there or we need to wrap this break in an “else” block

dofinn: good catch. again.

PR #358, opened 2020-05-04
Fix account failure race condition

iamkirkbater:
Sometimes, when attempting to run the finalizer on an accountclaim the secrets watcher will update the CR in an attempt to rotate the credentials, which will cause the finalizer to fail and the account to go into a failed state.

This PR is designed to catch that specific error and requeue instead of failing the account.

### DIFF
--- a/pkg/controller/accountclaim/accountclaim_controller.go
+++ b/pkg/controller/accountclaim/accountclaim_controller.go
@@ -2,6 +2,7 @@ package accountclaim
 
 import (
 	"context"
+	rawError "errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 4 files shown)


pkg/controller/accountclaim/accountclaim_controller.go
```diff
@@ -136,6 +137,12 @@ func (r *ReconcileAccountClaim) Reconcile(request reconcile.Request) (reconcile.
 					// If the finalize/cleanup process fails for an account we don't want to return
 					// we will flag the account with the Failed Reuse condition, and with state = Failed
 
+					// First we want to see if this was an update race condition where the credentials rotator will update the CR while the finalizer is trying to run.  If that's the case, we want to requeue and retry, before outright failing the account.
+					if strings.HasSuffix(err.Error(), "the object has been modified; please apply your changes to the latest version and try again") {
```
iamkirkbater, line 141: This should use the k8serr package to see if it's a NotFound err instead of using strings.HasSuffix.

fahlmant:
@iamkirkbater Please review

PR #360, opened 2020-05-07
Use shared types key name vars

jharrington22:
CCS IAM clean up isn't using the correct tag names to search and delete users.

### DIFF
--- a/pkg/controller/accountclaim/reuse.go
+++ b/pkg/controller/accountclaim/reuse.go
@@ -400,10 +400,10 @@ func (r *ReconcileAccountClaim) cleanUpIAM(reqLogger logr.Logger, awsClient awsc
 		}
 		user = getUser.User
 		for _, tag := range user.Tags {
-			if *tag.Key == "ClusterName" && *tag.Value == accountCR.Name {
+			if *tag.Key == awsv1alpha1.ClusterAccountNameTagKey && *tag.Value == accountCR.Name {
 				clusterNameTag = true
 			}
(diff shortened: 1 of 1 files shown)


jharrington22:
@2uasimojo yes there should be testing but would like to get this fixed before any CCS clusters are created in stage.

[1] 100% agree I was looking at this last night thinking why are these not constants. I will change them in another PR but it'll require testing of all workflows in case these are being changed anywhere.

2uasimojo:
> @2uasimojo yes there should be testing but would like to get this fixed before any CCS clusters are created in stage.

Fair. In that case I'll need to dig deeper to convince myself this is right. (If someone else who already has context would like to approve this first, feel free.)

2uasimojo:
Okay, I see the (only) other [place these are used](https://github.com/openshift/aws-account-operator/blob/7360448b468187cad3a5e99e5bf88df872faaee4/pkg/awsclient/tags.go#L50-L57) is where we're [creating the user](https://github.com/openshift/aws-account-operator/blob/7360448b468187cad3a5e99e5bf88df872faaee4/pkg/awsclient/iam.go#L137-L140), and this is how we identify users to clean up in the [accountclaim finalizer](https://github.com/openshift/aws-account-operator/blob/7360448b468187cad3a5e99e5bf88df872faaee4/pkg/controller/accountclaim/reuse.go#L69).
So this is a correct change; and also should be upgrade-safe because it'll only pick up things we missed before, not revoke buggy-but-working behavior.

/lgtm
/approve

PR #362, opened 2020-05-07
Fix Typo for memory limit

iamkirkbater:
Fixes the typo that's preventing the operator from deploying.


2uasimojo:
I'm watching @jharrington22 debug this live. Fix is legit (although only solves part of the problem). Let's land it.

/lgtm

PR #363, opened 2020-05-13
change the IAM policy for osdManagedAdmin user - OSD-2360

bmeng:


bmeng:
cc @jharrington22 Could you help review the change?

fahlmant:
Is there a reason for this change? Is there a card to go along with it?

bmeng:
> Is there a reason for this change? Is there a card to go along with it?

https://issues.redhat.com/browse/OSD-2360 Yes, here is the related card.

clcollins:
@rogbas Please follow up with this PR, thank you.

jharrington22:
@bmeng part of the reason we use the AWS policy is so that we can ensure that the policy is the same every time, its the same policy we ask customers to use for CCS.

In reality without modifying logic to make this operator aware of layered products we cannot do this (and I don't think we should make it aware of LP). 

We could however ask CCS customers to use restricted SCP's and or use a different OU for LP and apply a less restrictive policy that the OSD policy at an account level for non CCS.

+1 to close this PR for the time being if everyone agrees :smile:

PR #365, opened 2020-05-15
increase the resource limits/requests for the operator to make sure it…

bmeng:
… would be running well all the time

After discussing with @drewandersonnz and based on the problem @jharrington22 raised in https://github.com/openshift/aws-account-operator/pull/354

Too make sure the operator can be ran without any risk, bump the values of the memory and cpu to the same, large enough value for limits and requests.

cc @rogbas @jharrington22

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -23,11 +23,11 @@ spec:
           imagePullPolicy: Always
           resources:
             requests:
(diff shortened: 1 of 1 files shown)


rogbas:
@rogbas review PR and comment

jharrington22:
+1 to closing until we do further testing.

PR #366, opened 2020-05-15
Fix EBS volume failed to delete error

ArielLima:
If an EBS volume is attached to an instance, force detach it before attempting to delete.
Ticket: https://issues.redhat.com/browse/OSD-3311
Test:
- Create an account -> Claim the account created -> Delete account claim

This ticket is WIP because I want others to test it, it works for me but my environment hasn't been working properly.


fahlmant:
@ArielLima Is there a way we can replicate this? EBS volumes aren't always stuck attached.

PR #367, opened 2020-05-15
Adds secret testing to Account Creation tests

iamkirkbater:
Adds the ability to test that

* Expected secrets are present at the end of account creation
* Expected secrets have expected keys

To Test:
* `make test-account-creation`
* It should now verify that the secrets are present.

Or if you already have an `osd-staging-1` account created:
* `make test-secrets`

### DIFF
--- a/test/integration/api/delete_account_secrets.sh
+++ b/test/integration/api/delete_account_secrets.sh
@@ -10,5 +10,5 @@ fi
 
 for secret in $(oc get secrets -n "${NAMESPACE}" | awk "/${OSD_STAGING_1_ACCOUNT_CR_NAME_OSD}/"'{ print $1 }')
 do
-    oc delete secret $secret -n ${NAMESPACE}
+    oc delete secret $secret -n ${NAMESPACE} || true
 done
--- a/test/integration/test_secrets.sh
+++ b/test/integration/test_secrets.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-TEST_ACCOUNT="osd-staging-1"
+TEST_ACCOUNT_CR_NAME="osd-staging-1"
 TEST_NAMESPACE="aws-account-operator"
 EXIT_STATUS="PASS"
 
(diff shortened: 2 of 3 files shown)


ArielLima:
Can we get some sort of output indicating the test passes?

iamkirkbater:
@ArielLima - Added a message that the test has passed.

rogbas:
@rogbas @fahlmant can you please test it and comment?

test/integration/api/delete_account_secrets.sh
```diff
@@ -10,5 +10,5 @@ fi
 
 for secret in $(oc get secrets -n "${NAMESPACE}" | awk "/${OSD_STAGING_1_ACCOUNT_CR_NAME_OSD}/"'{ print $1 }')
 do
-    oc delete secret $secret -n ${NAMESPACE}
-done
\ No newline at end of file
+    oc delete secret $secret -n ${NAMESPACE} || true
```
fahlmant, line 13: What is this doing? My brain is having trouble parsing this

test/integration/test_secrets.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+
+TEST_ACCOUNT="osd-staging-1"
```
fahlmant, line 3: `TEST_AWS_ACCOUNT` just to make it more clear? Trying to be explicit to avoid overlap between AWS Account and Account CR

fahlmant: Oh, if this is the Acount CR name, then can we change the name of it. And the var

fahlmant:
@iamkirkbater Can the bash script also try to use the credentials in the various secrets to ensure the credentials are valid?

test/integration/api/delete_account_secrets.sh
iamkirkbater: It's a failsafe.  It deletes the secret and if that fails (typically due to the secret not being present) it goes into the `|| true` and just continues on with execution.

Specifically I have a race-condition on my machine where sometimes this script is called while the secrets are being deleted with the accountclaim, so the for loop will pick up the secrets but the secret will be gone by the time the `oc delete` call is run.

iamkirkbater:
We could do that.  I think the original scope of this was to ensure the validity of the structure of the secrets, and not the validity of the secrets themselves, but we can expand the scope.

test/integration/api/delete_account_secrets.sh
fahlmant: Got it 👍

test/integration/test_secrets.sh
rogbas: @iamkirkbater can we please rename this `TEST_ACCOUNT` CR Name? Having the same name as the Payer account can be misleading.

fahlmant:
@fahlmant Please review

PR #368, opened 2020-05-18
Add const label name to metrics

yeya24:
Signed-off-by: yeya24 <yb532204897@gmail.com>

`name=aws-account-operator` can just be a const label. It is not necessary to use a `GaugeVec` in this case.

### DIFF
--- a/pkg/localmetrics/localmetrics.go
+++ b/pkg/localmetrics/localmetrics.go
@@ -23,54 +23,66 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 )
 
+const operatorName = "aws-account-operator"
+
 var log = logf.Log.WithName("localmetrics")
 
 var (
-	MetricTotalAWSAccounts = prometheus.NewGaugeVec(prometheus.GaugeOpts{
-		Name: "aws_account_operator_total_aws_accounts",
-		Help: "Report how many accounts have been created in AWS org",
-	}, []string{"name"})
-	MetricTotalAccountCRs = prometheus.NewGaugeVec(prometheus.GaugeOpts{
-		Name: "aws_account_operator_total_account_crs",
-		Help: "Report how many account CRs have been created",
-	}, []string{"name"})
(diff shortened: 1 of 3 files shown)


yeya24:
@fahlmant  Could you please take a look? 😄

pkg/localmetrics/localmetrics.go
```diff
@@ -23,54 +23,66 @@ import (
...
-	}, []string{"name"})
-	MetricPoolSizeVsUnclaimed = prometheus.NewGaugeVec(prometheus.GaugeOpts{
-		Name: "aws_account_operator_pool_size_vs_unclaimed",
-		Help: "Report the difference between the pool size and the number of unclaimed account CRs",
-	}, []string{"name"})
-	MetricTotalAccountPendingVerification = prometheus.NewGaugeVec(prometheus.GaugeOpts{
-		Name: "aws_account_operator_total_aws_account_pending_verification",
-		Help: "Report the number of accounts waiting for enterprise support and EC2 limit increases in AWS",
-	}, []string{"name"})
+	MetricTotalAWSAccounts = prometheus.NewGauge(prometheus.GaugeOpts{
```
iamkirkbater, line 31: Would it make sense to abstract these to a function?

```
func newGauge(name string, help string) prometheus.Gauge { #or whatever the type would be
    return prometheus.NewGauge(prometheus.GaugeOpts{
        Name: name,
        Help: help,
    })
}
```

yeya24: Thanks for your review. IMO a function is not necessary since the current way is already a common pattern.

ArielLima:
This looks good to me, @fahlmant input?

fahlmant:
At a quick glance this looks good. There were issues adding labels directly to our metrics I vaguely remember, so I'll need to test this before merging.

yeya24:
Is it ready to go?

fahlmant:
Looks good, I'll merge it next week 
/hold

PR #370, opened 2020-05-21
Ensures godoc compatibility with iam.go

iamkirkbater:
This PR fixes a few comments to ensure compatibility with Godoc standards.

### DIFF
--- a/pkg/controller/account/iam.go
+++ b/pkg/controller/account/iam.go
@@ -39,20 +39,21 @@ type awsStatement struct {
 	Principal *awsv1alpha1.Principal `json:"Principal,omitempty"`
 }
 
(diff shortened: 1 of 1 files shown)


pkg/controller/account/iam.go
```diff
@@ -174,7 +177,7 @@ func (r *ReconcileAccount) BuildSTSUser(reqLogger logr.Logger, awsSetupClient aw
 	// with eventual consisency on AWS' side
 	time.Sleep(10 * time.Second)
 
-	// Create STS user for SRE admins
+	// Create the sre-admin user credentials
```
fahlmant, line 180: This is specifically STS credentials, which is different from the normal Access Keys

PR #372, opened 2020-05-29
Adds validity test to AWSFederatedRoles

iamkirkbater:
Tests that a given role CR has valid IAM permissions.  

TODO:
- [x] Add Make test
- [x] Validate test works with both CRs
- [x] Validate test works with `make test-all`

## To Test:
- Change an action name to something misspelled or add a fake one to the bottom of either of the federated role CRs
- `make test-awsfederatedrole`

## Example Outputs:

*Valid CR: *
```
aws-account-operator [$]
> TEST_CR=test-federated-user TEST_ROLE_FILE=deploy/crds/aws_v1alpha1_awsfederatedrole_networkmgmt_cr.yaml go test github.com/openshift/aws-account-operator/test/integration/
ok      github.com/openshift/aws-account-operator/test/integration      2.630s
```

*Misspelled Action: *
```
aws-account-operator [$]
> TEST_CR=test-federated-user TEST_ROLE_FILE=deploy/crds/aws_v1alpha1_awsfederatedrole_networkmgmt_cr.yaml go test github.com/openshift/aws-account-operator/test/integration/
--- FAIL: TestFederatedAccessRolePermissions (0.88s)
    federated_access_cr_test.go:137: ec2:CreateRouet is not allowed by RoleARN: arn:aws:iam::423379392308:role/network-mgmt-4pcgsz
        -- Possibly a missing context or the action does not exist?
        {
          EvalActionName: "ec2:CreateRouet",
          EvalDecision: "implicitDeny",
          EvalResourceName: "*",
          MissingContextValues: ["ram:RequestedResourceType"],
          OrganizationsDecisionDetail: {
            AllowedByOrganizations: true
          }
        }
FAIL
FAIL    github.com/openshift/aws-account-operator/test/integration      1.530s
FAIL
```


rogbas:
@rogbas @fahlmant @ArielLima @clcollins can you please give it a run and comment?

fahlmant:
Kirk this looks awesome! 
Just a few questions:
1. Are there any known limitations to the SimulatePrincipal call? Anything we can't test? 
2. For the go file, any reason to not use the structs we have to build the policies in the controllers? And any specific reason to target yaml over JSON in this case?

iamkirkbater:
Thanks @fahlmant!  

I don't specifically know of any limitations, but I do know that it starts getting significantly more complex as you add conditions or target principals, but they should still be testable, I'm just not sure how to add that into our testing framework here.

For the go file, I went with `yaml` because I wanted to be able to test the CR directly from the file rather than through the controller.  I guess since creating the CR is part of the setup of the test this could be refactored to utilize the controller by both creating the CR and then getting the CR and using the built-in structs.  I'm not sure what the LOE would be to make that change, compared to just leaving it as it is for now.  We shouldn't have to change anything here unless the structure of the yaml changes, but it is introducing tech debt.  I guess the question here is more along the lines of how much effort we want to put into this.

I think when I started this before I didn't have the understanding of the kubernetes client that I've gained within the last few weeks, so I just went with the "easiest" solution to me at the time.

fahlmant:
@fahlmant Please review

PR #373, opened 2020-06-01
[WIP] Set AFAAs to be updated when FederatedRole is updated

fahlmant:
When a FederatedRole, currently there is no mechanism to update any FederatedAccountAccess using the FederatedRole. This PR adds that mechanism by adding a label to any FederatedAccountAccess using the FederatedRole, forcing the FederatedAccountAccess reconcile loop to be triggered to update the role in AWS.

### DIFF
--- a/pkg/controller/awsfederatedrole/awsfederatedrole_controller.go
+++ b/pkg/controller/awsfederatedrole/awsfederatedrole_controller.go
@@ -108,6 +108,7 @@ func (r *ReconcileAWSFederatedRole) Reconcile(request reconcile.Request) (reconc
 		return reconcile.Result{}, nil
 	}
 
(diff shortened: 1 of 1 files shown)


rogbas:
@fahlmant please evaluate if we can use the `metadata.generation`  field to keep track of changes to the AFAA CR, after confirming it gets incremented only with changes to the CR's spec.

clcollins:
Testing with a vanilla Kube cluster, adding an annotation incrementes both resourceVersion and generation.

Adding a label only increments resourceVersion

Neither increment `.metadata.annotation.deployment.kubernetes.io/revision`

fahlmant:
Closing, need to spike on if this is even possible with our current setup.

PR #374, opened 2020-06-01
Use configmap to determine aws account limit

ArielLima:
[ticket](https://issues.redhat.com/browse/OSD-3924)

Remove hard coded account limit from the operator. Since different environments have different account limits we should look for the given limit in the operator's ConfigMap.

### DIFF
--- a/pkg/apis/aws/v1alpha1/shared_types.go
+++ b/pkg/apis/aws/v1alpha1/shared_types.go
@@ -92,7 +92,7 @@ var ErrFailedAWSTypecast = errors.New("FailedToTypecastAWSError")
 var ErrMissingDefaultConfigMap = errors.New("MissingDefaultConfigMap")
 
 // ErrInvalidConfigMap indicates that the ConfigMap has invalid fields
-var ErrInvalidConfigMap = errors.New("OUConfigMapInvalid")
+var ErrInvalidConfigMap = errors.New("ConfigMapInvalid")
 
 // ErrNonexistentOU indicates that an OU does not exist
 var ErrNonexistentOU = errors.New("OUWithNameNotFound")
--- a/pkg/apis/aws/v1alpha1/shared_types.go
+++ b/pkg/apis/aws/v1alpha1/shared_types.go
@@ -88,8 +88,11 @@ var ErrCreateEC2Instance = errors.New("EC2CreationTimeout")
 // ErrFailedAWSTypecast indicates that there was a failure while typecasting to aws error
 var ErrFailedAWSTypecast = errors.New("FailedToTypecastAWSError")
 
-// ErrInvalidOUMap indicates that the OU ConfigMap is not valid
-var ErrInvalidOUMap = errors.New("OUConfigMapInvalid")
+// ErrMissingDefaultConfigMap indicates that the expected default confimap was not found
+var ErrMissingDefaultConfigMap = errors.New("MissingDefaultConfigMap")
+
+// ErrInvalidConfigMap indicates that the ConfigMap has invalid fields
+var ErrInvalidConfigMap = errors.New("OUConfigMapInvalid")
 
 // ErrNonexistentOU indicates that an OU does not exist
 var ErrNonexistentOU = errors.New("OUWithNameNotFound")
@@ -128,3 +131,9 @@ var EmailID = "osd-creds-mgmt"
 
 // InstanceResourceType is the resource type used when building Instance tags
 var InstanceResourceType = "instance"
+
+// DefaultConfigMap holds the expected name for the operator's ConfigMap
+var DefaultConfigMap = "aws-account-operator-configmap"
+
+// DefaultConfigMapAccountLimit holds the fallback limit of aws-accounts
+var DefaultConfigMapAccountLimit = 100
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -36,7 +36,6 @@ var log = logf.Log.WithName("controller_account")
 
 const (
 	// AwsLimit tracks the hard limit to the number of accounts; exported for use in cmd/manager/main.go
-	AwsLimit                = 4800
 	awsCredsUserName        = "aws_user_name"
 	awsCredsSecretIDKey     = "aws_access_key_id"
 	awsCredsSecretAccessKey = "aws_secret_access_key"
@@ -286,7 +285,7 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 
 			if !accountHasAwsAccountID(currentAcctInstance) {
 				// before doing anything make sure we are not over the limit if we are just error
-				if totalaccountwatcher.TotalAccountWatcher.Total >= AwsLimit {
+				if ok, _ := checkAWSAccountsLimitReached(r, reqLogger, totalaccountwatcher.TotalAccountWatcher.Total); ok {
 					reqLogger.Error(awsv1alpha1.ErrAwsAccountLimitExceeded, "AWS Account limit reached", "Account Total", totalaccountwatcher.TotalAccountWatcher.Total)
 					return reconcile.Result{}, awsv1alpha1.ErrAwsAccountLimitExceeded
 				}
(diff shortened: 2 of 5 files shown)


pkg/apis/aws/v1alpha1/shared_types.go
```diff
@@ -88,8 +88,11 @@ var ErrCreateEC2Instance = errors.New("EC2CreationTimeout")
 // ErrFailedAWSTypecast indicates that there was a failure while typecasting to aws error
 var ErrFailedAWSTypecast = errors.New("FailedToTypecastAWSError")
 
-// ErrInvalidOUMap indicates that the OU ConfigMap is not valid
-var ErrInvalidOUMap = errors.New("OUConfigMapInvalid")
+// ErrMissingDefaultConfigMap indicates that the expected default confimap was not found
+var ErrMissingDefaultConfigMap = errors.New("MissingDefaultConfigMap")
+
+// ErrInvalidConfigMap indicates that the ConfigMap has invalid fields
+var ErrInvalidConfigMap = errors.New("OUConfigMapInvalid")
```
fahlmant, line 95: Is this different than the account limit being invalid?  Any reason not to change to `ConfigMapInvalid`?

pkg/controller/account/account_controller.go
```diff
@@ -605,6 +604,27 @@ func CreateAccount(reqLogger logr.Logger, client awsclient.Client, accountName,
...
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Namespace: awsv1alpha1.AccountCrNamespace, Name: awsv1alpha1.DefaultConfigMap}, instance)
+	if err != nil {
+		unexpectedErrorMsg := fmt.Sprintf("%s: Failed to retrieve default ConfigMap, account limit defaulting to 100", awsv1alpha1.ErrMissingDefaultConfigMap)
+		reqLogger.Info(unexpectedErrorMsg)
+	} else {
+		if limit, ok := instance.Data["account-limit"]; ok {
+			if i, err := strconv.Atoi(limit); err == nil {
+				return i <= currentAccounts, nil
+			}
+			unexpectedErrorMsg := fmt.Sprintf("Account: Failed to convert ConfigMap 'account-limit' string field to int, account limit defaulting to 100")
```
fahlmant, line 618: You don't need this or the else below. If the error != nil, then the code will continue, so you can put this line and the reqlogger at the end of the else block from 613

pkg/apis/aws/v1alpha1/shared_types.go
ArielLima: @fahlmant I missed updating the error name thanks for catching that! I use this error to encompass all cases of missing or bad field values. Are you saying we should have errors specific to each field that is missing/invalid?

fahlmant: No I'm saying we should just have one `ConfigMapInvalid`

pkg/controller/account/account_controller.go
ArielLima: @fahlmant Actually this logs when converting the `account-limit` field fails while the else bellow logs when the `account-limit` field is missing entirely. Since they are logging two different issues I think leaving it this way is best. Are you concerned about too many logs?

fahlmant: Ah I see the different errors. I don't see a big difference between both errors, as they both say that the config map is misconfigured essentially. I guess it doesn't matter too much. Curious on other's thoughts

pkg/apis/aws/v1alpha1/shared_types.go
ArielLima: Does my latest commit fix that concern?

fahlmant: 👍

pkg/controller/account/account_controller.go
```diff
@@ -605,6 +604,27 @@ func CreateAccount(reqLogger logr.Logger, client awsclient.Client, accountName,
 	return accountStatus, nil
 }
 
+func checkAWSAccountsLimitReached(r *ReconcileAccount, reqLogger logr.Logger, currentAccounts int) (bool, error) {
+	instance := &corev1.ConfigMap{}
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Namespace: awsv1alpha1.AccountCrNamespace, Name: awsv1alpha1.DefaultConfigMap}, instance)
+	if err != nil {
+		unexpectedErrorMsg := fmt.Sprintf("%s: Failed to retrieve default ConfigMap, account limit defaulting to 100", awsv1alpha1.ErrMissingDefaultConfigMap)
+		reqLogger.Info(unexpectedErrorMsg)
+	} else {
```
iamkirkbater, line 613: I'm not a huge fan of this giant else block (or else statements 98% of the time, really).  This is pretty confusing to read and understand, at least at first glance.  Is there any way we could just return more often rather than rely on else statements through here?

pkg/controller/account/account_controller.go
iamkirkbater: It probably wouldn't hurt to be more explicit with the errors, to help anyone who may be on-call who's debugging a potential problem in the future who's not as familiar with the operator.

I do, however, like the idea of returning more often rather than these else blocks, as I've indicated in another comment earlier.

pkg/controller/account/account_controller.go
```diff
@@ -605,6 +604,27 @@ func CreateAccount(reqLogger logr.Logger, client awsclient.Client, accountName,
 	return accountStatus, nil
 }
 
+func checkAWSAccountsLimitReached(r *ReconcileAccount, reqLogger logr.Logger, currentAccounts int) (bool, error) {
```
iamkirkbater, line 607: Can we add some unit tests with this function before we merge it?

iamkirkbater: Let me know if you want to :pear: on the testing if you're getting blocked or something.

pkg/controller/account/account_controller.go
ArielLima: @iamkirkbater Currently the method returns twice; when it works and we know exactly what the account limit is or when it fails and we want to default that limit to 100. Normally we would return at every error but since any other return would be nearly identical to the one that is already there I decided to return only once. We don't need to worry about "missing" an error since this method runs in a linearly fashion the `error` variable always has the newest value. This is why I approached this the way I did, logging any issues that were encounter and return only once.

pkg/controller/account/account_controller.go
ArielLima: @iamkirkbater If you think it would be appropriate to add comments I'm happy to do that but as far as the control flow goes I don't know how it can be any more efficient or clean. If you have any suggestions I'm all ears.

pkg/controller/account/account_controller.go
ArielLima: Good point! I'll add some in my next PR.

pkg/controller/account/account_controller.go
iamkirkbater: You know, maybe it was the fact that I was looking at this in the diff page and it had some of the comments in it so it looked a lot longer than I originally thought, when I went to the actual file to look at this it doesn't look as bad.  I can live with this.  If we need to add another else in here, though, I'm going to suggest refactoring again, but we shouldn't do that until there's a few unit tests in place anyway.

iamkirkbater:
@fahlmant @ArielLima If we create a configmap locally we should be able to test this, right?  Then update the configmap through the separate paths we expect it to take?

pkg/controller/account/account_controller.go
ArielLima: I will take care of this request on a separate PR https://issues.redhat.com/browse/OSD-4073

ArielLima:
I manually tested this by running the `checkAWSAccountsLimitReached` function at the beginning of the reconcile loop and everything worked as expected. Aside from creating unit tests which already have a separate dedicated [card](https://issues.redhat.com/browse/OSD-4073) this looks good to me.
@rogbas @fahlmant @iamkirkbater

PR #377, opened 2020-06-04
Add finalizer to the byoc secret, ensure it can be used before deletion

fahlmant:
Currently, the byoc secret can be deleted before the accountclaim has a chance to finalize. This PR adds a finalizer to the byoc secret to ensure the secret is still around when the accountclaim tries to use it. Once the claim has used the secret, it removes the finalizer.

### DIFF
--- a/pkg/controller/accountclaim/accountclaim_controller.go
+++ b/pkg/controller/accountclaim/accountclaim_controller.go
@@ -35,6 +35,7 @@ const (
 	awsCredsAccessKeyId     = "aws_access_key_id"
 	awsCredsSecretAccessKey = "aws_secret_access_key"
 	accountClaimFinalizer   = "finalizer.aws.managed.openshift.io"
+	byocSecretFinalizer     = accountClaimFinalizer + "/byoc"
 	waitPeriod              = 30
 )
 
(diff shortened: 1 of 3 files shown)


pkg/controller/accountclaim/accountclaim_controller.go
```diff
@@ -167,6 +168,27 @@ func (r *ReconcileAccountClaim) Reconcile(request reconcile.Request) (reconcile.
 
 		reqLogger.Info("Reconciling BYOC AccountClaim")
 
+		// Ensure BYOC secret has finalizer
```
iamkirkbater, line 171: Is there any way we could take this added functionality and add it as a separate function, possibly with a unit test or two?

pkg/controller/accountclaim/reuse.go
```diff
@@ -85,6 +85,26 @@ func (r *ReconcileAccountClaim) finalizeAccountClaim(reqLogger logr.Logger, acco
 		if err != nil {
 			reqLogger.Error(err, "Failed to delete BYOC account from accountclaim cleanup")
 		}
+
```
iamkirkbater, line 88: Same as the other comment, can we abstract this out to another function and unit test it?

PR #378, opened 2020-06-07
Log failed accounts and don't reconcile

jharrington22:
This PR sets the account controller to log and ignore accounts with the `.status.state` "Failed"


jharrington22:
@rogbas @clcollins could you please test, this is a result of https://issues.redhat.com/browse/OSD-3992

PR #382, opened 2020-06-11
Refactor metrics collector

yeya24:
Signed-off-by: yeya24 <yb532204897@gmail.com>

Fixes:

https://issues.redhat.com/browse/OSD-4030
https://issues.redhat.com/browse/OSD-4033
https://issues.redhat.com/browse/OSD-4032
https://issues.redhat.com/browse/OSD-4031

## What are changed and how it works?

In the pervious metrics implementation, the metrics are updated in each reconcile loops. This way is not very reusable because if we add a new controller, maybe we have to add metrics update logic to its reconcile loop as well.

The main idea of refactoring is to move the metrics update logic out of the reconcile loop and it is also better to put all metrics collecting stuff in one place so it can be easily reused .

A common practice in controller metrics is https://github.com/coreos/prometheus-operator/blob/master/pkg/alertmanager/collector.go. We just need to implement a collector struct with two interfaces, then the struct can be registered to the Prometheus registry and collected at the configured Prometheus scrape interval.

```
Describe(ch chan<- *prometheus.Desc)
Collect(ch chan<- prometheus.Metric)
```

## label changes

In previous metrics, we have some metrics like `aws_account_operator_total_accounts_crs_unclaimed` or `aws_account_operator_total_accounts_crs_claimed`, I put some of the status fields in Prometheus metrics label.

Labels we have 
Screenshot from 2020-06-11 10-11-37

### DIFF
--- a/pkg/localmetrics/localmetrics.go
+++ b/pkg/localmetrics/localmetrics.go
@@ -15,176 +15,219 @@
 package localmetrics
 
 import (
-	"math"
-	"sync"
+	"context"
 
 	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
+
 	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 )
 
-const operatorName = "aws-account-operator"
-
-var log = logf.Log.WithName("localmetrics")
-
-var (
-	MetricTotalAWSAccounts = prometheus.NewGauge(prometheus.GaugeOpts{
-		Name:        "aws_account_operator_total_aws_accounts",
-		Help:        "Report how many accounts have been created in AWS org",
-		ConstLabels: prometheus.Labels{"name": operatorName},
-	})
-	MetricTotalAccountCRs = prometheus.NewGauge(prometheus.GaugeOpts{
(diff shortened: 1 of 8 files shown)


pkg/localmetrics/localmetrics.go
```diff
@@ -15,176 +15,152 @@
...
+		}, []string{"ccs", "legal_id"}),
+
+		// pool_name is not a good label because it may cause
+		// high cardinality. But in our use case it is okay
+		// since we only has one account pool in the cluster.
+		accountPoolSize: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name:        "aws_account_operator_account_pool_size",
+			Help:        "Report the size of account pool CR",
+			ConstLabels: prometheus.Labels{"name": operatorName},
+		}, []string{"namespace", "pool_name"}),
```
fahlmant, line 74: In practice, we will only ever have one pool due to the design of hive. The cluster name gets appended to the metric in the app-sre prometheus (i.e. `hive-stage`). I'm pretty sure we have the same pool name across all our environments anyway so I'm not sure if this label is useful

yeya24: You are totally right. But I think the design of `account pool cr` is quite weird, if we can only have one in the cluster, then why do we need that cr?

fahlmant: The CR was added when we had a different design. This is just the way its worked out.

yeya24: I think it is okay to keep the `pool_name` label. Though it is not very useful, it won't cause cardinality problems. What do you think?

PR #384, opened 2020-06-15
Create and Delete BYOCAdminAccess role per cluster

fahlmant:
This PR alters the creation of the BYOCAdminAccess role to match osdManagedAdmin and osdManagedAdminSRE users creation to be unique per cluster/accountclaim.

### DIFF
--- a/pkg/controller/accountclaim/reuse.go
+++ b/pkg/controller/accountclaim/reuse.go
@@ -497,6 +497,53 @@ func (r *ReconcileAccountClaim) cleanUpIAM(reqLogger logr.Logger, awsClient awsc
 		}
 	}
 
+	reqLogger.Info("Cleaning up IAM roles")
+
+	roles, err := awsclient.ListIAMRoles(reqLogger, awsClient)
+	if err != nil {
+		return err
+	}
+
+	for _, role := range roles {
+		clusterNameTag := false
+		clusterNamespaceTag := false
+		getRole, err := awsClient.GetRole(&iam.GetRoleInput{RoleName: role.RoleName})
+		if err != nil {
+			return err
+		}
+
+		for _, tag := range getRole.Role.Tags {
+			if *tag.Key == awsv1alpha1.ClusterAccountNameTagKey && *tag.Value == accountCR.Name {
+				clusterNameTag = true
+			}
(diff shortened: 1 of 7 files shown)


fahlmant:
The `gofmt -s` from `make docker-build` is fighting the `gofmt -s` from `make`. Trying to reconcile them

pkg/controller/accountclaim/reuse.go
```diff
@@ -497,6 +497,53 @@ func (r *ReconcileAccountClaim) cleanUpIAM(reqLogger logr.Logger, awsClient awsc
...
+					return fmt.Errorf(fmt.Sprintf("Unable to detach IAM role policy from role %s", *getRole.Role.RoleName), err)
+				}
+			}
+			_, err = awsClient.DeleteRole(&iam.DeleteRoleInput{RoleName: getRole.Role.RoleName})
+			reqLogger.Info(fmt.Sprintf("Deleting IAM role: %s", *getRole.Role.RoleName))
+			if err != nil {
+				return fmt.Errorf(fmt.Sprintf("Unable to delete IAM role %s", *getRole.Role.RoleName), err)
+			}
+		} else {
+			reqLogger.Info(fmt.Sprintf("Not deleting role: %s", *getRole.Role.RoleName))
```
rogbas, line 544: Why would we not delete the role if it doesn't have the AccountName and Namespace tags? To avoid breaking compatibility with older clusters?

rogbas:
I checked this PR and the workflow LGTM, I +1 merging it after addressing the question added above.

pkg/controller/accountclaim/reuse.go
fahlmant: Or any other cluster in that AWS Account. If Cluster A and Cluster B were both CCS clusters in Account X, deleting all BYOCAdminAccess Roles when Cluster A is deleted would leave us unable to access Cluster B with that role. 

Basically, one BYOC role per CCS cluster, tied only to that cluster, regardless of larger account they're in. This follows the same patterns as the unique IAM users that James added recently.

rogbas: ok, it makes sense, thanks for confirming

rogbas:
@fahlmant I'm +1 to merging it, please do it when you can.

PR #385, opened 2020-06-15
Refactors Secret Creation

iamkirkbater:
Uses secret refresh logic during secret creation.
Also refactors secret refresh logic to not save so that both of the secrets can be updated at the same time.

To Test:
* `make create-account`
* Validate all secrets are present


pkg/controller/account/account_controller.go
```diff
@@ -389,24 +389,15 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 
 		}
 
-		// Create new awsClient with SRE IAM credentials so we can generate STS and Federation tokens from it
-		SREAWSClient, err := r.awsClientBuilder(r.Client, awsclient.NewAwsClientInput{
-			SecretName: *SREIAMUserSecret,
-			NameSpace:  awsv1alpha1.AccountCrNamespace,
-			AwsRegion:  "us-east-1",
-		})
 		if err != nil {
```
fahlmant, line 392: Does this err still need to be here?

iamkirkbater: Good catch.  That's a no.

rogbas:
@rogbas @fahlmant can you please give it a run locally?

fahlmant:
@fahlmant @rogbas

pkg/controller/account/iam.go
```diff
@@ -166,72 +165,6 @@ func (r *ReconcileAccount) CreateSecret(reqLogger logr.Logger, account *awsv1alp
 	return nil
 }
 
-// BuildSTSUser sets up an IAM user with the proper access and creates secrets to hold credentials
-// Takes a logger, an awsSetupClient for the signing token, an awsClient for, an account CR to set ownership of secrets, the namespace to create the secret in, and a role to assume with the creds
-// The awsSetupClient is the client for the user in the target linked account
-// The awsClient is the client for the user in the payer level account
-func (r *ReconcileAccount) BuildSTSUser(reqLogger logr.Logger, awsSetupClient awsclient.Client, awsClient awsclient.Client, account *awsv1alpha1.Account, nameSpace string, iamRole string) (string, error) {
```
fahlmant, line 173: Are the users created using the secret refresh logic in the rotator?

pkg/controller/account/account_controller.go
```diff
@@ -427,29 +412,38 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
 		reqLogger.Info(fmt.Sprintf("rotating console URL credentials for %s", currentAcctInstance.Name))
 		err = r.RotateConsoleCredentials(reqLogger, awsSetupClient, currentAcctInstance)
 		if err != nil {
 			return reconcile.Result{}, err
 		}
+		credentialsRotated = true
 	}
+	if credentialsRotated {
+		// Set `status.RotateCredentials` to false now that they have been updated
+		currentAcctInstance.Status.RotateCredentials = false
```
2uasimojo, line 438: I like that we're consolidating the actual push of the status update, but this isn't clearing the `RotateConsoleCredentials` status. You would want to leave the local status updates (L438, and its counterpart for `RotateConsoleCredentials = false`) inside the conditions above, and make this condition _only_ push the `Status().Update()`. 

Personally, I think it's goodness if `Rotate[Console]Credentials` is still responsible for updating the account status locally. Can we consider changing `Rotate[Console]Credentials` to returna a bool indicating whether it did that or not?

pkg/controller/account/iam.go
2uasimojo, line 173: Confirmed that this method is not called anywhere.

(Perhaps this cleanup belongs in a separate PR?)

pkg/controller/account/account_controller.go
```diff
@@ -442,14 +422,23 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
 		reqLogger.Info(fmt.Sprintf("rotating console URL credentials for %s", currentAcctInstance.Name))
 		err = r.RotateConsoleCredentials(reqLogger, awsSetupClient, currentAcctInstance)
 		if err != nil {
 			return reconcile.Result{}, err
 		}
+		currentAcctInstance.Status.RotateConsoleCredentials = false
 	}
+
+	// Push the updated account to the server.
+	err = r.Client.Status().Update(context.TODO(), currentAcctInstance)
```
2uasimojo, line 437: So now this will always push a (single) status update. But I think there are still code paths where we didn't actually rotate anything and the update would be a no-op, right? That may not be a big deal, but if you wanted to keep the `credentialsRotated` bool from the previous iteration, you could get rid of that.

iamkirkbater: There might be, but this is also where the account creation gets pushed, as part of this refactor was forcing the credentials to be rotated/created when the account is created as well as when the CR is updated so that the credentials get rotated.

With that said, though, the "credential rotation" will be getting refactored right after this PR in another one to move that functionality out of the CR and just into the secret_watcher outright.  We'll still need to create the users and secrets here, but then we'd circle back on whether we want to do one update at the end even if nothing's changed compared to specifically updating when something changes.

pkg/controller/account/iam.go
iamkirkbater: So this got removed accidentally, so I'll need to re-add this in.

pkg/controller/account/account_controller.go
2uasimojo: Okay, sounds like a plan. Thanks!

/lgtm

iamkirkbater:
If #408 is approved then this should be closed.

PR #392, opened 2020-06-25
Adds VPC Peering Permissions

iamkirkbater:
Should be merged after the DirectConnect (#391) and VPN (#390) PRs.

PR #391, opened 2020-06-25
Adds broken down permissions for Direct Connect

iamkirkbater:
Adds DirectConnect permissions based on Customer Facing Documentation.

Needs to be merged after https://github.com/openshift/aws-account-operator/pull/390

PR #390, opened 2020-06-25
Adds Permissions for creating a VPN

iamkirkbater:
This PR begins to add the targeted AWS Statements based on the Customer Facing documentation.

PR #393, opened 2020-06-25
Cleans up Legacy Permissions into blocks.

iamkirkbater:
Instead of completely removing legacy permissions some made sense to keep, so I added them.

I took the liberty of getting rid of GuardDuty reporting.  That's not applicable to Network Management and if we want to enable customer self-service for that then we should consider adding a separate role they can assume.

I do have questions about others, though:

- [ ] Do we want to let our VPCs accept peering connections?  The documentation says it should go one way, outbound from our cluster.  I guess the small argument is that there might be a rogue peering request that someone might blindly accept, but I could see that argument going either way.
- [ ] During my setup I didn't see anything that needed the `ec2:CreateVpnConnectionRoute` or Delete of the same.  Not sure if this is related to troubleshooting or something that needs to also be enabled, maybe someone smarter than me can comment.
- [x] Can someone confirm that RAM and Route53Resolver are both related to VPC Peering?  RAM seems to be something that allows you to "see" resources in other accounts/VPCs that are sharing, and the Route53 Resolver looks to be like it automatically creates DNS routes between VPCs, so these seem useful to keep.  Just looking for confirmation so I can add it to our permissions document for future reference. - [kbater note: RAM was used for DirectConnect, being able to share a transit gateway]

### DIFF
--- a/deploy/crds/aws_v1alpha1_awsfederatedrole_networkmgmt_cr.yaml
+++ b/deploy/crds/aws_v1alpha1_awsfederatedrole_networkmgmt_cr.yaml
@@ -12,58 +12,33 @@ spec:
     description: Description of CustomerAdministratorAccess
     # list of statements for the policy
     awsStatements:
+      ## START LEGACY PERMISSIONS
+      # AWS Transit Gateway
+      - effect: Allow
+        action:
+          - "ec2:CreateTransitGatewayVpcAttachment"
+        resource:
+          - "arn:aws:ec2:*:*:subnet/*"
+          - "arn:aws:ec2:*:*:transit-gateway/*"
+          - "arn:aws:ec2:*:*:vpc/*"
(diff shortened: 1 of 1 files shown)


iamkirkbater:
/hold

Do not merge until #390 #391 and #392 have been merged and questions above have been resolved.

fahlmant:
@fahlmant Please review
Leads review on this one

iamkirkbater:
@mwoodson @jewzaam @jharrington22 @rogbas @fahlmant - This one reorganizes some of the policies in the list, and there's still the open question above about letting our side `Accept` the VPC peering request.  It would be a better customer experience to have us Accept, but at the same time is an extra possible attack vector, however it is quite a minimal one at that because they'd have to know the ARN of our VPC I believe.

Would love any feedback related to this change.

PR #397, opened 2020-07-06
OSD-4181: Unify PHONY targets

lisa:
There were extra hyphens.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,13 +59,13 @@ delete-account:
 test-account-creation: create-account delete-account
 
 # Create account claim namespace
-.PHONY: create-account-claim-namespace
+.PHONY: create-accountclaim-namespace
 create-accountclaim-namespace:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -59,13 +59,13 @@ delete-account:
 test-account-creation: create-account delete-account
 
 # Create account claim namespace
-.PHONY: create-account-claim-namespace
+.PHONY: create-accountclaim-namespace
```
fahlmant, line 62: Line 284 uses `create-account-claim-namespace` and need to be updated

lisa: Fixed. Good catch.

jharrington22:
Thanks @lisa ! looks good to me, squash first and I'll merge.

lisa:
> Thanks @lisa ! looks good to me, squash first and I'll merge.

Squashed.

PR #398, opened 2020-07-06
Change summary metrics to histogram

yeya24:
Signed-off-by: Ben Ye <yb532204897@gmail.com>

Summary metrics are collected in client side, which means every time the operator restarts (maybe triggered by CD update), the current metrics will be lost. Histogram also has some other advantages.

So this pr updates the metrics type to histogram.

### DIFF
--- a/pkg/localmetrics/localmetrics.go
+++ b/pkg/localmetrics/localmetrics.go
@@ -42,9 +42,9 @@ type MetricsCollector struct {
 	accountClaims                   *prometheus.GaugeVec
 	accountReuseAvailable           *prometheus.GaugeVec
 	accountPoolSize                 *prometheus.GaugeVec
-	accountReadyDuration            prometheus.Summary
-	accountClaimReadyDuration       *prometheus.SummaryVec
-	accountReuseCleanupDuration     prometheus.Summary
+	accountReadyDuration            prometheus.Histogram
+	accountClaimReadyDuration       *prometheus.HistogramVec
+	accountReuseCleanupDuration     prometheus.Histogram
 	accountReuseCleanupFailureCount prometheus.Counter
 }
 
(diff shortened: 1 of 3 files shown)


jharrington22:
@yeya24 can you explain how this will effect the current graphs? Currently the summary doesn't seem to change the graph when the operator is restarted. I might just be miss understanding what you are changing here

fahlmant:
> Histogram also has some other advantages.
Can you please explain what those advantages are?

yeya24:
Summary is collected at client side, it has a expiration mechanism. The metrics are aggregated into a 10 min time window, but after 10 mins, all the metrics at that time window will expire. So you will find all the summary metrics return NaN.

Histogram is collected at Prometheus side, so the metrics expiration case won't happen. And it supports ad-hoc query as well.

fahlmant:
@fahlmant @rogbas please review

pkg/localmetrics/localmetrics.go
```diff
@@ -85,21 +85,24 @@ func NewMetricsCollector(store cache.Cache) *MetricsCollector {
 			Help:        "Report the size of account pool cr",
 			ConstLabels: prometheus.Labels{"name": operatorName},
 		}, []string{"namespace", "pool_name"}),
-		accountReadyDuration: prometheus.NewSummary(prometheus.SummaryOpts{
+		accountReadyDuration: prometheus.NewHistogram(prometheus.HistogramOpts{
 			Name:        "aws_account_operator_account_ready_duration_seconds",
 			Help:        "The duration for account cr to get ready",
 			ConstLabels: prometheus.Labels{"name": operatorName},
+			Buckets:     []float64{1, 3, 5, 10, 15, 20, 30},
```
fahlmant, line 92: @fahlmant @rogbas Please review buckets

yeya24: @fahlmant Do we need to add a ccs label for account cr?

yeya24: And the buckets need to update, I saw some metrics with ready time > 100s.

And there is one metric with ready time = 17655s. Seems there is a bug.

pkg/localmetrics/localmetrics.go
```diff
@@ -85,21 +89,40 @@ func NewMetricsCollector(store cache.Cache) *MetricsCollector {
 			Help:        "Report the size of account pool cr",
 			ConstLabels: prometheus.Labels{"name": operatorName},
 		}, []string{"namespace", "pool_name"}),
-		accountReadyDuration: prometheus.NewSummary(prometheus.SummaryOpts{
-			Name:        "aws_account_operator_account_ready_duration_seconds",
+
+		// Use minute as the unit
+		accountReadyDuration: prometheus.NewHistogram(prometheus.HistogramOpts{
+			Name:        "aws_account_operator_account_ready_duration_minutes",
```
2uasimojo, line 95: The [naming conventions](https://prometheus.io/docs/practices/naming/) specifically discourage using anything other than "base units" (in this case "seconds"). Is there a good reason to do otherwise here?

yeya24: Thanks. I think you are totally right! I will update this metric to use seconds.

yeya24: It might be good if we can introduce this linter to ensure our metrics best practice. https://github.com/prometheus/client_golang/blob/master/prometheus/testutil/promlint/promlint.go

2uasimojo: I wasn't able to tell at a glance how that would work, but I like the idea.

yeya24: https://github.com/coredns/coredns/blob/master/test/metric_naming_test.go That's how Coredns integrates this. We can have it as well.

2uasimojo:
[This page](https://prometheus.io/docs/practices/histograms/) may help understand the tradeoffs between Histogram and Summary. It sounds like we really don't want to mess with Summary unless we're angling for very specific SLO-based use cases. Histogram seems more flexible, especially while we're still figuring out exactly how we want the data used/presented.

2uasimojo:
@yeya24 Would you mind updating the PR summary (first comment) to also mention the reworking of how CCS vs non-CCS metrics are reported? I.e.
- Previously "account claim ready duration" was using a label on the same metric to distinguish CCS and non-CCS, and now it's using separate metrics. (Which BTW I think is 100% the right thing to do. As I've said elsewhere, CCS and non-CCS almost seem like they should even be separate CRs.)
- Previously "account ready duration" was not distinguishing CCS/non-CCS at all, and now it is (in the same way as above -- separate metrics rather than labels).

@fahlmant @rogbas Personally, I wouldn't sweat the buckets too much, since we are (I think?) using this more to observe "normals" so we can notice when spikes/anomalies occur, as opposed to predetermining expected durations and sounding alarms when we cross those boundaries. True?

Given that, and having dug a bit into Summar vs Histogram and agreeing with @yeya24's motivations for switching over, this PR
/lgtm

pkg/localmetrics/localmetrics.go
2uasimojo: Nice. Want to throw together a card for that, in case we don't get to it right away? (Seems like a nice-to-have, but probably edged out by more immediate priorities for the forseeable.)

PR #403, opened 2020-07-08
Removes boilerplate operator-sdk comments

clcollins:


pkg/controller/account/account_controller.go
```diff
@@ -110,11 +105,6 @@ type ReconcileAccount struct {
 
 // Reconcile reads that state of the cluster for a Account object and makes changes based on the state read
 // and what is in the Account.Spec
-// TODO(user): Modify this Reconcile function to implement your Controller logic.  This example creates
-// a Pod as an example
-// Note:
-// The Controller will requeue the Request to be processed again if the returned error is non-nil or
-// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
```
2uasimojo, line 117: I kind of like this part of the comment still being there. But meh.

clcollins: :+1: That's fine with me. I'll get that back in there for those.

PR #404, opened 2020-07-08
DRY MetricsCollector Describe and Collect

2uasimojo:
When updating the list of Collectors in MetricsCollector, it is
necessary to make sure that MetricsCollector's Describe and Collect
methods invoke the respective method for each wrapped Collector. This is
slightly tedious and error-prone.

This commit uses some reflection to abstract that part away, so that
adding a new field to MetricsCollector only requires you to do the
things specific to that field.

### DIFF
--- a/pkg/localmetrics/localmetrics.go
+++ b/pkg/localmetrics/localmetrics.go
@@ -16,6 +16,7 @@ package localmetrics
 
 import (
 	"context"
+	"reflect"
 
 	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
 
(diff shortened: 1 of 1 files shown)


pkg/localmetrics/localmetrics.go
```diff
@@ -110,33 +111,43 @@ func NewMetricsCollector(store cache.Cache) *MetricsCollector {
...
+// - Neither method returns anything, so neither does this method.
+func (c *MetricsCollector) runOnCollectors(method collectorMethod, arg interface{}) {
+	// Start reflecting
+	cValue := reflect.ValueOf(*c)
+	// For each field in our MetricsCollector receiver...
+	for i := 0; i < cValue.NumField(); i++ {
+		// ...if the field is an instance of a Collector subclass...
+		if cValue.Type().Field(i).Type.Implements(reflect.TypeOf((*prometheus.Collector)(nil)).Elem()) {
+			// ...Find and invoke the indicated method.
+			cValue.Field(i).MethodByName(string(method)).Call([]reflect.Value{reflect.ValueOf(arg)})
```
2uasimojo, line 137: Bah, this doesn't work unless we export all the fields in MetricsCollector. Which wouldn't be the worst thing, but it's also not pretty.

/hold

fahlmant:
@yeya24 Please review

2uasimojo:
> @yeya24 Please review

I actually got [feedback from Ben in slack](https://coreos.slack.com/archives/CNJPJSG3Z/p1594242964432000):

> But for using reflect to auto-register the metrics, I am not sure if it is good for readability. It is too hacky :joy:

I'm not stuck on doing this, and it would need some more work anyway as noted inline. So if others agree with Ben, or are even ambivalent, I'll just close it out. @fahlmant @iamkirkbater thoughts?

fahlmant:
+1 to what Ben said. I see the reflect as a bit of a hack, even though it does simplify things a bit. But it would be preferable to match best practices, which I believe @yeya24 followed during initial implementation here.

2uasimojo:
Roger that. Thanks for looking!

PR #406, opened 2020-07-09
template for cm

ArielLima:
Template for the ConfigMap used by this operator


maorfr:
this should go in to this template IMO:
https://github.com/openshift/aws-account-operator/blob/master/hack/olm-registry/olm-artifacts-template.yaml

deploy/configurations.yaml
```diff
@@ -0,0 +1,21 @@
...
+metadata:
+  name: aws-account-operator-configmap
+objects:
+- apiVersion: v1
+  kind: ConfigMap
+  metadata:
+    name: aws-account-operator-configmap
+    namespace: aws-account-operator
+  data:
+    account-limit: ${ACCOUNT_LIMIT}
```
maorfr, line 19: I would add quotes around `${ACCOUNT_LIMIT}`

ArielLima: By adding the quote here do I remove them from the app-interface pr? -> https://gitlab.cee.redhat.com/service/app-interface/-/blob/d3f2ce5db392eb9e55d5ddeeee58a70d986034d2/data/services/osd-operators/cicd/saas/saas-aws-account-operator.yaml#L42-42

ArielLima:
> this should go in to this template IMO:
> https://github.com/openshift/aws-account-operator/blob/master/hack/olm-registry/olm-artifacts-template.yaml

Sure

deploy/configurations.yaml
maorfr: nope, keep on both

fahlmant:
@rogbas @fahlmant please review

ArielLima:
@rogbas @fahlmant https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/6309

rogbas:
The way it's set in the app-interface MR linked above, we have account limit == 100 in all environments. It doesn't make sense considering what we have today.

rogbas:
@ArielLima can you please review the app-interface cicd config, making sure we have real account limits and not 100?

PR #407, opened 2020-07-09
Support for 'cluster' dev mode

2uasimojo:
Adds support for `FORCE_DEV_MODE=cluster`, which still skips AWS support
case creation, but serves metrics from the cluster as normal.

Renames the `deploy` make target to `predeploy`.

Introduces a `deploy-local` target which invokes the appropriate
`operator-sdk` command for you, with `FORCE_DEV_MODE_=local`.

Introduces a `deploy-cluster` target which deploys the operator in the
"real" cluster you're logged into, with `FORCE_DEV_MODE=cluster`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -279,8 +279,8 @@ deploy-aws-account-operator-credentials:
 	$(eval ID=$(shell echo -n ${OPERATOR_ACCESS_KEY_ID} | base64 ))
 	$(eval KEY=$(shell echo -n ${OPERATOR_SECRET_ACCESS_KEY} | base64))
 
-.PHONY: deploy-aws-account-operator
-deploy-aws-account-operator:
+.PHONY: predeploy-aws-account-operator
+predeploy-aws-account-operator:
 # Create aws-account-operator namespace
 	@oc get namespace ${NAMESPACE} || oc new-project ${NAMESPACE}
 # Create the aws-account-operator-credentials secret
@@ -290,8 +290,24 @@ deploy-aws-account-operator:
 # Create zero size account pool
 	@oc apply -f hack/files/aws_v1alpha1_zero_size_accountpool.yaml
 
-.PHONY: deploy
-deploy: check-aws-credentials deploy-aws-account-operator-credentials deploy-aws-account-operator
+.PHONY: predeploy
+predeploy: check-aws-credentials deploy-aws-account-operator-credentials predeploy-aws-account-operator
+
+.PHONY: deploy-local
+deploy-local: FORCE_DEV_MODE?=local
+deploy-local:
+	@operator-sdk up local --namespace=$(OPERATOR_NAMESPACE)
+
+.PHONY: deploy-cluster
+deploy-cluster: FORCE_DEV_MODE?=cluster
+deploy-cluster: isclean
+# Deploy things like service account, roles, etc.
+# TODO(efried): Filtering out operator.yaml here is icky, but necessary so we can do the substitutions.
+#               Revisit when templating mentioned below is done.
+	@ls deploy/*.yaml | grep -v operator.yaml | xargs -L1 oc apply -f
+# Deploy the operator resource, using our dev image and the appropriate (or requested) dev mode
+# TODO(efried): template this, but without having to maintain an almost-copy of operator.yaml
+	@hack/scripts/edit_operator_yaml_for_dev.py $(OPERATOR_IMAGE_URI) "$(FORCE_DEV_MODE)" | oc apply -f -
 
 .PHONY: .check-aws-credentials
 check-aws-credentials:
--- a/README.md
+++ b/README.md
@@ -9,9 +9,10 @@
     - [1.2. Requirements](#12-requirements)
     - [1.3. Workflow](#13-workflow)
     - [1.4. Testing your AWS account credentials with the CLI](#14-testing-your-aws-account-credentials-with-the-cli)
-    - [1.5. Local Development](#15-local-development)
+    - [1.5. Development](#15-development)
     - [1.5.1 Operator Install](#151-operator-install)
-    - [1.5.2 Running the Operator and Local Development Mode](#152-running-the-operator-and-local-development-mode)
+      - [1.5.1.1 Local Mode](#1511-local-mode)
+      - [1.5.1.2 Cluster Mode](#1512-cluster-mode)
   - [2. The Custom Resources](#2-the-custom-resources)
     - [2.1. AccountPool CR](#21-accountpool-cr)
     - [2.2. Account CR](#22-account-cr)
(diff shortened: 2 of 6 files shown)


pkg/controller/utils/utils.go
```diff
@@ -19,10 +19,21 @@ const (
...
+	//   . Edit deploy/operator.yaml. Under .spec.template.spec.env, add:
+	//         - name: FORCE_DEV_MODE
+	//           value: cluster
+	//   . `oc apply` all the YAML files in deploy/, including the updated operator.yaml.
+	envDevMode = "FORCE_DEV_MODE"
+
+	// MetricsPort is the port at which we'll serve metrics
+	MetricsPort               = "8080"
+	// MetricsPath is the URI path at which we'll serve metrics
+	MetricsPath               = "/metrics"
```
2uasimojo, line 36: Moved these in from main.go mainly to make them easier to refer to from the comments below.

hack/scripts/edit_operator_yaml_for_dev.py
```diff
@@ -0,0 +1,38 @@
...
+
+container = y['spec']['template']['spec']['containers'][0]
+# Replace the container image
+container['image'] = image_uri
+# Add the FORCE_DEV_MODE env var
+container['env'].append(dict(name='FORCE_DEV_MODE', value=dev_mode))
+
+print(yaml.dump(y, default_flow_style=False))
+
+sys.exit(0)
```
yeya24, line 38: A small nit, please add a new line here.

fahlmant:
@fahlmant Please set up review time with Eric + team

README.md
```diff
@@ -9,11 +9,16 @@
...
-    - [1.5. Local Development](#15-local-development)
+    - [1.5. Development](#15-development)
+    - [1.5.1 Operator Install](#151-operator-install)
+      - [1.5.1.1 Local Mode](#1511-local-mode)
+      - [1.5.1.2 Cluster Mode](#1512-cluster-mode)
   - [2. The Custom Resources](#2-the-custom-resources)
     - [2.1. AccountPool CR](#21-accountpool-cr)
     - [2.2. Account CR](#22-account-cr)
     - [2.3. AccountClaim CR](#23-accountclaim-cr)
+    - [2.4. AWSFederatedRole CR](#24-awsfederatedrole-cr)
```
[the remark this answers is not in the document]
2uasimojo: :)

Makefile
```diff
@@ -255,8 +255,21 @@ deploy-aws-account-operator:
 # Create zero size account pool
 	@oc apply -f hack/files/aws_v1alpha1_zero_size_accountpool.yaml
 
-.PHONY: deploy
-deploy: check-aws-credentials deploy-aws-account-operator-credentials deploy-aws-account-operator
+.PHONY: deploy-local
```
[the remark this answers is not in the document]
2uasimojo: Since the sub-targets are using `oc apply`, those steps are idempotent, so running them each time is a no-op -- *unless* you've changed your secrets or those resource defs, which is also a thing you might be doing while developing. So I'd definitely like to keep umbrella targets that do both, if possible.

If you really want to keep a way to skip the setup steps, I split it out again. Naming-wise, I feel like a target with `deploy` in the name should actually wind up with the operator deployed. So I could
- rename the existing `deploy` (`prepare`? `predeploy`? `your-suggestion-here`?)
- (optionally) create new targets (`up-local`/`up-cluster`? `run-local`/`run-cluster`? modern operator-sdk CLI replaced `up` with `run` so the latter might be more forward-looking) that *only* run the actual operator
- keep these `deploy-local` and `deploy-cluster` targets doing the same thing they're doing now, i.e. {default `FORCE_DEV_MODE`, predeploy, run}, by subsuming whatever targets we end up with above.

Thoughts?

hack/scripts/edit_operator_yaml_for_dev.py
2uasimojo: Will do. (I actually meant to run "black" on this, but forgot.)

Makefile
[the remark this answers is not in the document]
iamkirkbater: I guess the only real problem I'm going to have is that if the aws credentials are not stored in your environment then the deploy will fail, forcing you to keep the specific AWS credentials in your env instead of just a set-and-forget mentality because that's what the `check-aws-credentials` make target is doing.

However, with that said we could do something like modify the `check-aws-credentials` and the `deploy-aws-account-operator-credentials` to no-op if the secret is already present.  Or if we want to consider utilizing these to change the workflow it could be something like this:

* `check-aws-credentials` will exit cleanly if the aws credentials are present OR the secret is set.
* `deploy-aws-account-operator-credentials` will no-op if the secret is set AND the env vars are not set, otherwise it will replace the secret if the secret is present and the env vars are set, OR it will create the secret if it's not present and the env vars are set.

And then it will no-op instead of erroring for each run.

And honestly, even if we _don't_ do all of that work, at the end of the day you can just run `make deploy-local` and then continue with running `operator-sdk up local` the same way everyone's doing it now anyway.

Makefile
```diff
@@ -255,8 +255,21 @@ deploy-aws-account-operator:
...
+deploy-local: FORCE_DEV_MODE?=local
+deploy-local: check-aws-credentials deploy-aws-account-operator-credentials deploy-aws-account-operator
+	@operator-sdk up local --namespace=$(OPERATOR_NAMESPACE)
+
+.PHONY: deploy-cluster
+deploy-cluster: FORCE_DEV_MODE?=cluster
+deploy-cluster: isclean
```
2uasimojo, line 265: Oops, this needs to pull in the same things as L260

Makefile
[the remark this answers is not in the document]
2uasimojo: Okay, I think I made this happy without having a zillion targets doing slightly different things. In the latest commit:
- `make predeploy` does what `deploy` did before. It's the (only) thing that needs the creds env vars.
- `make deploy-local` and `make deploy-cluster` only do the bringing-up-the-actual-operator bits.
So the expected workflow is now:
```
make predeploy
make deploy-$whichever

while not done:
    ... develop ...
    make deploy-$whichever
```

Makefile
2uasimojo: Actually ended up doing it the other way: neither `deploy-*` target pulls in the predeploy steps.

hack/scripts/edit_operator_yaml_for_dev.py
2uasimojo: Done.

Makefile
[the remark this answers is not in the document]
iamkirkbater: Nice.  I'm not going to hold this up anymore but I would personally prefer `make install` instead of `make pre-deploy`. I'll ultimately leave that decision up to you, but this works for me :)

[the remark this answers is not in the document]
2uasimojo: Thinking it through, I would like to keep the $verb the same for all the targets, to make it clear how they're related. So either `preinstall` `install-cluster` `install-local` or `predeploy` `deploy-local` `deploy-cluster`.

Unless there's some nuanced definition of "install" vs "deploy" that I'm missing?

2uasimojo:
Nothing has changed here.

> @2uasimojo 2uasimojo force-pushed the 2uasimojo:dev-modes branch from 0289057 to 9512747

:arrow_up: Rebase only.

> @2uasimojo 2uasimojo force-pushed the 2uasimojo:dev-modes branch from 9512747 to 5ecb5bf

:arrow_up: Squash.

2uasimojo:
Sorry, weekend-rebase-while-distracted was bad. Re-rebased, pulling it off of the other two commits. No actual changes since last week.

rogbas:
@fahlmant @rogbas can you please give it a run locally?

README.md
```diff
@@ -128,26 +129,65 @@ The below commands can be used to test payer account credentials where we create
...
+
+It is recommended to let the operator know when you're running it for testing purposes.
+This has benefits such as skipping AWS support case creation.
+This is done by setting `FORCE_DEV_MODE` in the operator's environment.
+This is handled for you if you use one of the `make deploy-*` targets described below.
 
 ### 1.5.1 Operator Install
 
-The operator can be installed locally on a Minishift or Code-Ready-Containers (CRC) cluster (or a private OpenShift cluster).  On a new local cluster, running the Makefile `deploy` target (passing the Access Key ID and Secret Access Key) will install the operator, and create all the necessary OpenShift CRDs and secrets for it to work locally.
+The operator can be installed into various cluster and pseudo-cluster environments. Depending which you choose, you can run in "local" mode or in "cluster" mode. The former is known to work in a Minishift or Code-Ready-Containers (CRC) cluster, or a private OpenShift cluster. The latter is known to work in a real OSD cluster. You can try to mix and match; it might work.
```
fahlmant, line 141: What is a `private OpenShift cluster` in this context?

pkg/controller/utils/utils.go
```diff
@@ -19,10 +19,21 @@ const (
...
+	// * Running locally: `FORCE_DEV_MODE=local operator-sdk up local`
+	// * Running in a cluster:
+	//   . Edit deploy/operator.yaml. Under .spec.template.spec.env, add:
+	//         - name: FORCE_DEV_MODE
+	//           value: cluster
+	//   . `oc apply` all the YAML files in deploy/, including the updated operator.yaml.
+	envDevMode = "FORCE_DEV_MODE"
+
+	// MetricsPort is the port at which we'll serve metrics
+	MetricsPort = "8080"
```
fahlmant, line 34: Are these ever going to be used anywhere but cmd/manager/main.go?

README.md
2uasimojo: No idea, just kept that from the previous version.

pkg/controller/utils/utils.go
2uasimojo: No; I only moved them here so they would be near the [comment](https://github.com/openshift/aws-account-operator/pull/407/files/9922a4fb3b2e578e868e7469ab32ce921e79014f#diff-c9c1114922f55187dc190facb2951ff9R56-R59) that talks about them. I can move it back if you don't like that (and reword the comment to refer to main.go, which I thought was slightly ickier at the time, but meh).

fahlmant: Since they're specific to main, I think it makes sense to keep them there, and keep utils cleaner as well. But I won't block the PR on it

2uasimojo: Sure, no big deal, I moved 'em back. Makes the diff smaller too :)

PR #410, opened 2020-07-15
Metrics for API Calls (1 of 2: kubeclient)

2uasimojo:
A) Register a new metric, `aws_account_operator_api_request_duration_seconds`, which times API calls made by the operator. Since we're using Histogram, we get a `_count` as well. The metric has the following dimensions:
- `controller`: The name of the controller making the API call.
- `method`: The HTTP method (`GET`, `PUT`, etc.).
- `resource`: A normalized interpretation of the target of the API call. For example, unique resource names are scrubbed. This is so we don't end up with an explosion of (not very useful) dimension values.
- `status`: The HTTP response status (`200 OK`, `404 Not Found`, etc.).

B) Instrument the controller-runtime Client used by each controller to publish this metric. This gives us durations and counts for calls to kubeclient APIs.

Note: we're only doing this for true controllers, not for secretwatcher or totalaccountwatcher.

Note: This commit only handles kubeclient API calls. A subsequent commit will instrument the AWS client in a similar fashion.

Jira: [OSD-4390](https://issues.redhat.com/browse/OSD-4390)


pkg/controller/utils/clientwrapper.go
```diff
@@ -0,0 +1,82 @@
+package utils
```
2uasimojo, line 1: This file largely cribbed from [hive](https://github.com/openshift/hive/blob/68fa75297dbfe1a3911bb3e880a9ae0a746d8c75/pkg/controller/utils/clientwrapper.go).

pkg/localmetrics/localmetrics.go
```diff
@@ -134,6 +139,11 @@ func NewMetricsCollector(store cache.Cache) *MetricsCollector {
 			ConstLabels: prometheus.Labels{"name": operatorName},
 			Buckets:     []float64{0.001, 0.01, 0.1, 1, 10, 30, 60, 120, 240},
 		}, []string{"controller"}),
+
+		apiCallCount: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "aws_account_operator_api_requests_total",
```
2uasimojo, line 144: Note that this is called "api requests" and not "kubeclient requests" because we also intend to [accumulate AWS API calls](https://issues.redhat.com/browse/OSD-4507) in the same metric, as noted in the commit message.

pkg/localmetrics/localmetrics.go
```diff
@@ -270,3 +282,71 @@ func (c *MetricsCollector) AddAccountReuseCleanupFailure() {
...
+
+// resourceFrom normalizes an API request URL, including removing individual namespace and
+// resource names, to yield a string of the form:
+//     $group/$version/$kind[/{NAME}[/...]]
+// or
+//     $group/$version/namespaces/{NAMESPACE}/$kind[/{NAME}[/...]]
+// ...where $foo is variable, {FOO} is actually {FOO}, and [foo] is optional.
+// This is so we can use it as a dimension for the apiCallCount metric, without ending up
+// with separate labels for each {namespace x name}.
+func resourceFrom(url *neturl.URL) (resource string) {
```
2uasimojo, line 303: Generously adapted from [hive](https://github.com/openshift/hive/blob/68fa75297dbfe1a3911bb3e880a9ae0a746d8c75/pkg/controller/utils/clientwrapper.go#L106-L131)

pkg/localmetrics/localmetrics.go
jharrington22: @2uasimojo looking at the dimensions we'll be able to determine if its a kube API request by the `resource` ?

pkg/localmetrics/localmetrics.go
```diff
@@ -270,3 +282,71 @@ func (c *MetricsCollector) AddAccountReuseCleanupFailure() {
...
+		// Core resources: /api/$version/...
+		// => core/$version/...
+		tokens[0] = "core"
+	case "apis":
+		// Extensions: /apis/$group/$version/...
+		// => $group/$version/...
+		tokens = tokens[1:]
+	default:
+		// Something else. Punt.
+		panic(1)
```
jharrington22, line 328: This will be a problem when we instrument AWS API calls no?

pkg/localmetrics/localmetrics.go
jharrington22: This is nice.

pkg/localmetrics/localmetrics.go
2uasimojo: Yes. More below.

pkg/localmetrics/localmetrics.go
2uasimojo: :)

As noted, stole the concept from hive. I used a slightly different format. But my main contribution was documenting the thing :P

pkg/localmetrics/localmetrics.go
2uasimojo: When we do that, we'll have to beef up this method to generate a usable resource string. I haven't fully formulated what that should look like, but it for sure doesn't need to have the same format as the kubeclient calls. We can make the prefix `aws/$service/` (e.g. `aws/ec2/`) and go from there. If we ever need to generate graphs for just kube or just AWS, prom lets you use regexes for label matching. Will still need to figure out how to scrub unique names/IDs out.

(BTW, anticipating the need to get the AWS service name is why I made the parameter to this func the whole URL rather than just the Path part as hive did -- IIUC the service name is in the domain, not the path.)

pkg/localmetrics/localmetrics.go
jharrington22: Saw it after :+1:

pkg/localmetrics/localmetrics.go
jharrington22: Yeah we are not instrumenting the AWS API calls now but good to know you thought of it for when the time comes. And  yes for AWS the service is in the domain. ec2.<region>.amazonaws.com exceptions are global services like route53 where there is no region but we can deal with that when the time comes.

jharrington22:
Held I'm happy with this but would like a review from v1alpha.

pkg/localmetrics/localmetrics.go
```diff
@@ -270,3 +282,71 @@ func (c *MetricsCollector) AddAccountReuseCleanupFailure() {
...
 	c.reconcileDuration.WithLabelValues(controller).Observe(duration)
 }
+
+func (c *MetricsCollector) AddAPICall(controller string, req *http.Request, resp *http.Response) {
+	c.apiCallCount.With(prometheus.Labels{
+		"controller": controller,
+		"method":     req.Method,
+		"resource":   resourceFrom(req.URL),
+		"status":     resp.Status,
+	}).Inc()
```
yeya24, line 292: This can be simplified to `c.apiCallCount.WithLabelValues(controller, req.Method,resourceFrom(req.URL), resp.Status)`

pkg/localmetrics/localmetrics.go
```diff
@@ -134,6 +139,11 @@ func NewMetricsCollector(store cache.Cache) *MetricsCollector {
 			ConstLabels: prometheus.Labels{"name": operatorName},
 			Buckets:     []float64{0.001, 0.01, 0.1, 1, 10, 30, 60, 120, 240},
 		}, []string{"controller"}),
+
+		apiCallCount: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "aws_account_operator_api_requests_total",
+			Help: "Counter incremented for each API request.",
+		}, []string{"controller", "method", "resource", "status"}),
```
yeya24, line 146: Please add the const label `ConstLabels: prometheus.Labels{"name": operatorName},`

pkg/localmetrics/localmetrics.go
yeya24: @2uasimojo Do we want to use this metric for both k8s and AWS API calls? Can I ask why is the design? IMO it is better to separate the metric for AWS and kube.

2uasimojo: @yeya24 Yes, the intent will be to use the same metric for both.

Scrape-/graph-wise, it's [sixes](https://en.wiktionary.org/wiki/six_of_one,_half_a_dozen_of_the_other): As written, we can filter by resource regex if we want to split them; or if we split them here we can always aggregate them on the graph side.

So it came down to how it's coded. I preferred the idea of funneling everything through a single common method. Also, the metrics for both types of call would be identical in structure, and I would prefer not to repeat.

pkg/localmetrics/localmetrics.go
2uasimojo: Ah, thanks, will do.

pkg/localmetrics/localmetrics.go
2uasimojo: I don't see that as "simplified". I like the explicitness of listing the label names, rather than relying on the order they happen to be listed on L146.

pkg/localmetrics/localmetrics.go
2uasimojo: Done.

pkg/localmetrics/localmetrics.go
yeya24: Umm, make sense!

pkg/localmetrics/localmetrics.go
yeya24: @2uasimojo For AWS API calls, how to get `method`, `status` and `resource`? Do you want to instrument the AWS client?

2uasimojo: I looked into it *briefly*, and the AWS clients allow you to pass in [handlers](https://github.com/aws/aws-sdk-go/blob/master/aws/request/handlers.go) as part of the config. AFAICT a "handler" contains a [callback function](https://github.com/aws/aws-sdk-go/blob/b868519ad5e70e7b66b3ef4753590fa752996c99/aws/request/handlers.go#L140) to which the [Request](https://github.com/aws/aws-sdk-go/blob/b868519ad5e70e7b66b3ef4753590fa752996c99/aws/request/request.go#L46) is passed, and the Request contains the [http.Request and Response](https://github.com/aws/aws-sdk-go/blob/b868519ad5e70e7b66b3ef4753590fa752996c99/aws/request/request.go#L55-L56).

Hopefully that's as simple as it looks, and we just have to decide a) where to stuff our handler, and b) what we want the `resource` to look like for AWS calls, so we can slice and dice the Request accordingly.

2uasimojo:
Smashed #415 into this.

pkg/controller/utils/clientwrapper.go
```diff
@@ -0,0 +1,84 @@
...
+	cfg := rest.CopyConfig(mgr.GetConfig())
+	AddControllerMetricsTransportWrapper(cfg, controller)
+
+	options := client.Options{
+		Scheme: mgr.GetScheme(),
+		Mapper: mgr.GetRESTMapper(),
+	}
+	c, err := client.New(cfg, options)
+	if err != nil {
+		log.Error(err, "Unable to initialize metrics-wrapped client")
```
arjunrn, line 29: `log.Fatal` and drop the `os.Exit`

pkg/controller/utils/clientwrapper.go
```diff
@@ -0,0 +1,84 @@
...
+// and register the call with our apiCallCount metric.
+func (cmt *ControllerMetricsTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	start := time.Now()
+	// Call the nested RoundTripper.
+	resp, err := cmt.RoundTripper.RoundTrip(req)
+
+	// Count this call, if it worked (where "worked" includes HTTP errors).
+	if err == nil {
+		localmetrics.Collector.AddAPICall(cmt.Controller, req, resp, time.Since(start).Seconds())
+	}
```
arjunrn, line 81: Do we also want to count failures? It could be useful when the failures are due to timeouts.

pkg/localmetrics/localmetrics.go
```diff
@@ -269,3 +286,75 @@ func (c *MetricsCollector) AddAccountReuseCleanupFailure() {
...
+// resourceFrom normalizes an API request URL, including removing individual namespace and
+// resource names, to yield a string of the form:
+//     $group/$version/$kind[/{NAME}[/...]]
+// or
+//     $group/$version/namespaces/{NAMESPACE}/$kind[/{NAME}[/...]]
+// ...where $foo is variable, {FOO} is actually {FOO}, and [foo] is optional.
+// This is so we can use it as a dimension for the apiCallCount metric, without ending up
+// with separate labels for each {namespace x name}.
+func resourceFrom(url *neturl.URL) (resource string) {
+	defer func() {
```
arjunrn, line 313: Can we avoid the `panic()/recover()` control flow? It might end up swallowing errors from other parts of the code later in life.

pkg/controller/utils/clientwrapper.go
2uasimojo: `logr.Logger` doesn't have a `Fatal`, and I didn't want to break with the rest of the repo by using a different logger interface.

pkg/controller/utils/clientwrapper.go
2uasimojo: Just making sure I understand what you're asking:

As written, this will still capture "failures" like 4xx/5xx. But you're talking about situations where there isn't even a Response, which presumably we recognize via `err != nil`?

Yes, we could capture those. We would need to decide what to put in the `status` field. Suggestions?

Or are you suggesting we count them in a separate metric?

pkg/localmetrics/localmetrics.go
2uasimojo: Since it's limited to this function, it won't swallow anything higher in the call stack. And we very much want to swallow anything unexpected that happens here and below. If we don't, an unexpected error (think malformed URL resulting in [array index out of bounds](https://github.com/openshift/aws-account-operator/pull/410/files/c06c3833e04cc709adb3bc2f8d03260abeebda18#diff-6454382f68b967b06cd3ef5504f90bc1R77-R79)) would percolate up and `panic` the caller, which could be very bad.

Do you agree?

pkg/controller/utils/clientwrapper.go
arjunrn: Yes, stuff like connection and read timeouts could be covered with separate metric. But it may be an overkill because we probably already have monitoring for stuff like that.

pkg/localmetrics/localmetrics.go
arjunrn: Maybe we would want there to be a panic in some cases. Assume in the future if there is a new path `/api-x/openshift/` all the objects under it would get counted under the same metric and there would be no way to find out when it happens. But in the end it's a stylistic choice for me(I don't like jumps in execution). So feel free to ignore it.

pkg/controller/utils/clientwrapper.go
2uasimojo: We can always iterate on this. Whether we want to fold it into this metric or introduce a separate one, let's have the discussion around a future PR, agree?

pkg/localmetrics/localmetrics.go
2uasimojo: The catch-all will result in those calls being counted in the `{OTHER}` bucket. As noted in the comment, I'm trying not to solve a problem we don't know we have yet -- if we start seeing things in that bucket, we can start digging in.

pkg/controller/utils/clientwrapper.go
arjunrn: Sounds good.

pkg/localmetrics/localmetrics.go
arjunrn: That works as well

2uasimojo:
/hold cancel

@rogbas has reviewed

PR #411, opened 2020-07-16
Edge case fix for ConcurrentModificationException OU error

ArielLima:
**Card**: https://issues.redhat.com/browse/OSD-4494
**Test**: None, hard to test and there's no need to IMO

This is a fix to the ConcurrentModificationException (caused by race a condition) error that we've been seeing due to two stage environments.

### DIFF
--- a/pkg/controller/accountclaim/accountclaim_controller.go
+++ b/pkg/controller/accountclaim/accountclaim_controller.go
@@ -324,6 +324,9 @@ func (r *ReconcileAccountClaim) Reconcile(request reconcile.Request) (reconcile.
 	if accountClaim.Spec.AccountOU == "" || accountClaim.Spec.AccountOU == "ROOT" {
 		err = MoveAccountToOU(r, reqLogger, accountClaim, unclaimedAccount)
 		if err != nil {
+			if err == awsv1alpha1.ErrAccMoveRaceCondition {
+				return reconcile.Result{Requeue: true}, nil
+			}
 			return reconcile.Result{}, err
 		}
 	}
--- a/pkg/controller/accountclaim/accountclaim_controller.go
+++ b/pkg/controller/accountclaim/accountclaim_controller.go
@@ -325,6 +325,7 @@ func (r *ReconcileAccountClaim) Reconcile(request reconcile.Request) (reconcile.
 		err = MoveAccountToOU(r, reqLogger, accountClaim, unclaimedAccount)
 		if err != nil {
 			if err == awsv1alpha1.ErrAccMoveRaceCondition {
+				// Due to a race condition, we need to requeue the reconcile to ensure that the account was correctly moved into the correct OU
 				return reconcile.Result{Requeue: true}, nil
 			}
 			return reconcile.Result{}, err
--- a/pkg/controller/accountclaim/organizational_units.go
+++ b/pkg/controller/accountclaim/organizational_units.go
@@ -3,6 +3,8 @@ package accountclaim
 import (
 	"context"
 	"fmt"
+	"math/rand"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/organizations"
@@ -121,8 +123,15 @@ func MoveAccount(reqLogger logr.Logger, client awsclient.Client, account *awsv1a
 				if check {
 					return awsv1alpha1.ErrAccAlreadyInOU
 				}
+			case "ConcurrentModificationException":
+				// if we encounter a race condition we simply try again after a random sleep interval
+				r := rand.Intn(5)
+				ConcurrentModificationExceptionMsg := fmt.Sprintf("CreateOrganizationalUnit:ConcurrentModificationException: Race condition while attempting to move Account: %s to OU: %s, trying again in %d seconds", account.Spec.AwsAccountID, OUID, r)
+				reqLogger.Info(ConcurrentModificationExceptionMsg)
+				time.Sleep(time.Duration(r) * time.Second)
+				MoveAccount(reqLogger, client, account, OUID, parentID)
 			default:
-				unexpectedErrorMsg := fmt.Sprintf("CreateOrganizationalUnit: Unexpected AWS Error when attempting to move AWS Account: %s to OU: %s", account.Spec.AwsAccountID, aerr.Code())
+				unexpectedErrorMsg := fmt.Sprintf("CreateOrganizationalUnit: Unexpected AWS Error when attempting to move AWS Account: %s to OU: %s, Error: %s", account.Spec.AwsAccountID, OUID, aerr.Code())
 				reqLogger.Info(unexpectedErrorMsg)
 			}
 		}
--- a/pkg/controller/accountclaim/organizational_units.go
+++ b/pkg/controller/accountclaim/organizational_units.go
@@ -52,15 +52,20 @@ func MoveAccountToOU(r *ReconcileAccountClaim, reqLogger logr.Logger, accountCla
 	err = MoveAccount(reqLogger, awsClient, account, OUID, rootID)
 	if err != nil {
 		// If error was cause by the account already being inside the OU, simply update the accountclaim cr and returns
-		if err == awsv1alpha1.ErrAccAlreadyInOU {
+		switch err {
+		case awsv1alpha1.ErrAccAlreadyInOU:
 			// Log account already in desired location
 			accountMovedMsg := fmt.Sprintf("OU: Account %s was already in the desired OU %s", account.Name, account.Spec.LegalEntity.ID)
 			reqLogger.Info(accountMovedMsg)
 			// Update accountclaim spec
 			accountClaim.Spec.AccountOU = OUID
 			return r.specUpdate(reqLogger, accountClaim)
+		case awsv1alpha1.ErrAccMoveRaceCondition:
+			// Simply return the error since we want a requeue
+			return awsv1alpha1.ErrAccMoveRaceCondition
+		default:
+			return err
 		}
-		return err
 	}
 
 	// Log account moved successfully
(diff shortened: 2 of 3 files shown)


pkg/controller/accountclaim/organizational_units.go
```diff
@@ -121,8 +123,15 @@ func MoveAccount(reqLogger logr.Logger, client awsclient.Client, account *awsv1a
 				if check {
 					return awsv1alpha1.ErrAccAlreadyInOU
 				}
+			case "ConcurrentModificationException":
+				// if we encounter a race condition we simply try again after a random sleep interval
+				r := rand.Intn(5)
+				ConcurrentModificationExceptionMsg := fmt.Sprintf("CreateOrganizationalUnit:ConcurrentModificationException: Race condition while attempting to move Account: %s to OU: %s, trying again in %d seconds", account.Spec.AwsAccountID, OUID, r)
+				reqLogger.Info(ConcurrentModificationExceptionMsg)
+				time.Sleep(time.Duration(r) * time.Second)
+				MoveAccount(reqLogger, client, account, OUID, parentID)
```
fahlmant, line 132: Slight concern (very unlikely) is if ConcurrentModificationException is returned on consecutive runs then we get recursion. Eventually it should all return but there is a possibility of it happening.

ArielLima: @fahlmant I mean we will get recursion any time we run into the error but I wasn't be concerned since the chances of it taking say 30 seconds are 1/15625

2uasimojo: I don't think we should be retrying here. The "best practice" would be to have the main reconcile requeue (possibly with a delay) so the retry happens on the next run through.

Obv that would rely on the code path through this point being nicely idempotent, which I didn't dig into. But in theory, that next reconcile wouldn't reach this point because whatever operation this bounced off of would have already done the right thing. If we could count on that being the case, we could even just ignore this error and carry on.

ArielLima: > But in theory, that next reconcile wouldn't reach this point because whatever operation this bounced off of would have already done the right thing.

@2uasimojo Yeah, that was the one of the things I was wondering about while writing this. I need to check that it is actually the case. @fahlmant I am wondering though, are there no other places where these errors could be occurring without our knowledge?

fahlmant: Which errors? We log every AWS error we hit. If we know of one we should be handling different, then we should add it to this case statement. Otherwise, they go to default so they are still logged.

2uasimojo: And in the default case, the error percolates [up|https://github.com/openshift/aws-account-operator/pull/411/files#diff-90b216857326e9f99796eb1d9cefbc91R65] and will [requeue|https://github.com/openshift/aws-account-operator/blob/d849220c9a47ef7f2c32732838a87433f7957b6b/pkg/controller/accountclaim/accountclaim_controller.go#L336]. Which, assuming we're properly idempotent, is what we should be doing anyway.

Was there an actual problem caused by this error as observed? If it was just a case of, "hey, we saw this message and it made us worry," then we may not want to bother doing anything. But if it did cause a problem, IMO we should be focusing on idempotence, not special-casing.

ArielLima: @2uasimojo @fahlmant I did some research and it is indeed the case that accounts get moved even when the error occurs so I am updating the PR to reflect that. I am, however, still catching this error in order to stop these messages from cluttering the logs:

`
{"level":"error","ts":1594987296.3907142,"logger":"kubebuilder.controller","msg":"Reconciler error","controller":"accountclaim-controller","request":"uhc-staging-1efka7k6ue5idru7t9u7ftd0i0c4tua3/ci-cluster-4-4","error":"ConcurrentModificationException: AWS Organizations can't complete your request because it conflicts with another attempt to modify the same entity. Try again later.\n\tstatus code: 400, request id: 1359eaf3-3acd-406f-80bf-8b7555d4ccf7","stacktrace":"github.com/openshift/aws-account-operator/vendor/github.com/go-logr/zapr.(*zapLogger).Error\n\tsrc/github.com/openshift/aws-account-operator/vendor/github.com/go-logr/zapr/zapr.go:128\ngithub.com/openshift/aws-account-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\tsrc/github.com/openshift/aws-account-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:217\ngithub.com/openshift/aws-account-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func1\n\tsrc/github.com/openshift/aws-account-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:158\ngithub.com/openshift/aws-account-operator/vendor/k8s.io/apimachinery/pkg/util/wait.JitterUntil.func1\n\tsrc/github.com/openshift/aws-account-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:133\ngithub.com/openshift/aws-account-operator/vendor/k8s.io/apimachinery/pkg/util/wait.JitterUntil\n\tsrc/github.com/openshift/aws-account-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:134\ngithub.com/openshift/aws-account-operator/vendor/k8s.io/apimachinery/pkg/util/wait.Until\n\tsrc/github.com/openshift/aws-account-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:88"}
`

pkg/controller/accountclaim/organizational_units.go
```diff
@@ -124,12 +122,10 @@ func MoveAccount(reqLogger logr.Logger, client awsclient.Client, account *awsv1a
...
 			case "ConcurrentModificationException":
-				// if we encounter a race condition we simply try again after a random sleep interval
-				r := rand.Intn(5)
-				ConcurrentModificationExceptionMsg := fmt.Sprintf("CreateOrganizationalUnit:ConcurrentModificationException: Race condition while attempting to move Account: %s to OU: %s, trying again in %d seconds", account.Spec.AwsAccountID, OUID, r)
+				// if we encounter a race condition we simply log the condition and return
+				ConcurrentModificationExceptionMsg := fmt.Sprintf("OU:CreateOrganizationalUnit:ConcurrentModificationException: Race condition while attempting to move Account: %s to OU: %s", account.Spec.AwsAccountID, OUID)
 				reqLogger.Info(ConcurrentModificationExceptionMsg)
-				time.Sleep(time.Duration(r) * time.Second)
-				MoveAccount(reqLogger, client, account, OUID, parentID)
+				return nil
```
fahlmant, line 128: why return nil and not an error?

2uasimojo: Because this error indicates that the thing we wanted to do is already done. Which could be made clearer in the comment and/or log message.

However, that's only the case because we happen to know this is the only thing we're doing to this resource. It would be safer and more future-proof to requeue (either by returning the error or return with `Requeue: true`) and let the subsequent reconcile make sure the thing is in the right state.

fahlmant: Got it, guess I missed that this error still means the $stuff got done. Something more explicit would be nice here

pkg/controller/accountclaim/organizational_units.go
```diff
@@ -52,15 +52,20 @@ func MoveAccountToOU(r *ReconcileAccountClaim, reqLogger logr.Logger, accountCla
...
 			accountMovedMsg := fmt.Sprintf("OU: Account %s was already in the desired OU %s", account.Name, account.Spec.LegalEntity.ID)
 			reqLogger.Info(accountMovedMsg)
 			// Update accountclaim spec
 			accountClaim.Spec.AccountOU = OUID
 			return r.specUpdate(reqLogger, accountClaim)
+		case awsv1alpha1.ErrAccMoveRaceCondition:
+			// Simply return the error since we want a requeue
+			return awsv1alpha1.ErrAccMoveRaceCondition
+		default:
+			return err
```
2uasimojo, line 67: This delta is unnecessary

pkg/controller/accountclaim/accountclaim_controller.go
```diff
@@ -324,6 +324,9 @@ func (r *ReconcileAccountClaim) Reconcile(request reconcile.Request) (reconcile.
 	if accountClaim.Spec.AccountOU == "" || accountClaim.Spec.AccountOU == "ROOT" {
 		err = MoveAccountToOU(r, reqLogger, accountClaim, unclaimedAccount)
 		if err != nil {
+			if err == awsv1alpha1.ErrAccMoveRaceCondition {
+				return reconcile.Result{Requeue: true}, nil
```
2uasimojo, line 328: This would be a good spot for a comment like the one [below](https://github.com/openshift/aws-account-operator/pull/411/files#diff-90b216857326e9f99796eb1d9cefbc91R130).

pkg/controller/accountclaim/organizational_units.go
ArielLima: +1

pkg/controller/accountclaim/accountclaim_controller.go
2uasimojo: ++

pkg/controller/accountclaim/organizational_units.go
```diff
@@ -52,13 +52,17 @@ func MoveAccountToOU(r *ReconcileAccountClaim, reqLogger logr.Logger, accountCla
...
+		case awsv1alpha1.ErrAccAlreadyInOU:
 			// Log account already in desired location
 			accountMovedMsg := fmt.Sprintf("OU: Account %s was already in the desired OU %s", account.Name, account.Spec.LegalEntity.ID)
 			reqLogger.Info(accountMovedMsg)
 			// Update accountclaim spec
 			accountClaim.Spec.AccountOU = OUID
 			return r.specUpdate(reqLogger, accountClaim)
+		case awsv1alpha1.ErrAccMoveRaceCondition:
+			// Simply return the error since we want a requeue
+			return awsv1alpha1.ErrAccMoveRaceCondition
```
2uasimojo, line 65: Sorry, what I meant here is that this whole thing could have been left as is. The race condition case is the same as the fall-through, simply returning the `err`. So the only special thing that needs to be done here is if `ErrAccAlreadiInOU`, which is how it was written before.

Not a blocking issue, as the logic is (still) correct. Just an opportunity to clean up a little bit :)

ArielLima: You are right, I'll update it once again

rogbas:
@fahlmant please doublecheck the requeue approach and LGTM

PR #412, opened 2020-07-16
Use BYOCAdminAccess-uid role for secret rotation

fahlmant:
During the refactor from BYOCAdminAccess to BYOCAdminAccess-uid I missed this bit.

### DIFF
--- a/pkg/controller/account/credentials_rotator.go
+++ b/pkg/controller/account/credentials_rotator.go
@@ -25,9 +25,10 @@ func (r *ReconcileAccount) RotateCredentials(reqLogger logr.Logger, awsSetupClie
 
 	//var awsAssumedRoleClient awsclient.Client
 	var roleToAssume string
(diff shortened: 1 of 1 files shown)


pkg/controller/account/credentials_rotator.go
```diff
@@ -25,9 +25,10 @@ func (r *ReconcileAccount) RotateCredentials(reqLogger logr.Logger, awsSetupClie
 
 	//var awsAssumedRoleClient awsclient.Client
 	var roleToAssume string
-
+	accountIAMID := account.Labels[fmt.Sprintf("%s", awsv1alpha1.IAMUserIDLabel)]
```
iamkirkbater, line 28: Why the `fmt.Sprintf` here?  Is there a chance this comes back as not a string? Not blocking, just curious.

fahlmant: @iamkirkbater This is how its done in the account controller, so I copied it

PR #413, opened 2020-07-16
Adjust histogram buckets and rename reconcile metric

yeya24:
Signed-off-by: Ben Ye <yb532204897@gmail.com>

Adjust the buckets based on the observations on our dashboard.

### DIFF
--- a/pkg/localmetrics/localmetrics.go
+++ b/pkg/localmetrics/localmetrics.go
@@ -51,8 +51,6 @@ type MetricsCollector struct {
 }
 
 func NewMetricsCollector(store cache.Cache) *MetricsCollector {
-	// representing in minutes [1 3 5 10 20 30 60 120 240 300 480 600]
-	accountDurationBuckets := []float64{60, 180, 300, 600, 1200, 1800, 3600, 7200, 14400, 18000, 28800, 36000}
 	return &MetricsCollector{
 		store: store,
 		awsAccounts: prometheus.NewGauge(prometheus.GaugeOpts{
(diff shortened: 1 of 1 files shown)


pkg/localmetrics/localmetrics.go
```diff
@@ -129,10 +128,10 @@ func NewMetricsCollector(store cache.Cache) *MetricsCollector {
 			ConstLabels: prometheus.Labels{"name": operatorName},
 		}),
 		reconcileDuration: prometheus.NewHistogramVec(prometheus.HistogramOpts{
-			Name:        "aws_account_operator_reconcile_duration",
+			Name:        "aws_account_operator_reconcile_duration_seconds",
```
2uasimojo, line 131: ++

pkg/localmetrics/localmetrics.go
```diff
@@ -107,7 +106,7 @@ func NewMetricsCollector(store cache.Cache) *MetricsCollector {
 			Name:        "aws_account_operator_account_claim_ready_duration_seconds",
 			Help:        "The duration for account claim cr to get claimed",
 			ConstLabels: prometheus.Labels{"name": operatorName},
-			Buckets:     accountDurationBuckets,
+			Buckets:     []float64{1, 5, 10, 20, 30, 45, 60},
```
rogbas, line 109: Considering this is in seconds, I wonder if we should add another bucket for the extreme cases where it takes >1min.

Something like `120`.

yeya24: That makes sense, will update it.

yeya24:
PTAL again @rogbas @2uasimojo

2uasimojo:
does
/lgtm
need to be on its own line?

rogbas:
/lgtm
Let's merge this on Monday

PR #420, opened 2020-07-21
Strictify AWS mocks for BYOC and OU tests

2uasimojo:
These tests were lax in two ways:
- There was no `.Finish()`, so mock calls weren't being validated.
- The mock `EXPECT()`s were using `.AnyTimes()`, which means that (even once the `.Finish()` was added) the expected calls could be missed and the test would still pass.

(The mocks were still partially effective in that they were setting up appropriate `Return()` values.)

This commit adds the `.Finish()`es and removes the `.AnyTimes()`s. This enforces that each `EXPECT()`ed call is made exactly once (the default if you don't use any `*Times()`), and that no un`EXPECT()`ed calls are made.


rogbas:
@rogbas please run locally and cmmnt

PR #419, opened 2020-07-20
Mock AWS client in accountclaim_controller tests

2uasimojo:
Previously these tests were making real AWS calls (but not failing because the errors weren't percolating up).

This commit just mocks the AWS client so we're no longer making those calls. A future commit ought to make proper use of the mocks to `EXPECT()` calls appropriately, etc.

PR #423, opened 2020-07-21
Consistent, mockable AWS Clients

2uasimojo:
Building on [OSD-2839](https://issues.redhat.com/browse/OSD-2839), this commit refactors AWS Client generating such that it is done the same way by all controllers, and is easily mocked for testing purposes.

With this commit, the `awsclient` package exposes:
- A `Builder` interface with a `GetClient` method whose contract is identical to the (now gone) GetAWSClient func (except it now needs a receiver).
- A `RealBuilder` empty concrete implementation of `Builder`, whose `GetClient` method is exactly the artist formerly known as `GetAWSClient`.
- A `MockBuilder` concrete implementation of `Builder`. This struct has a field for a `gomock.Controller`, and its `GetClient` method returns a singleton instance of the mocked AWS client (from generated code in the `mock` package).
- A `GetMockClient` convenience method to facilitate access to an `EXPECT()`able `MockClient` from tests.
 
Controllers should use this as follows:
- Include a field of type `awsclient.Builder` in their Reconciler impl struct.
- Initialize that field to a pointer to an instance of `awsclient.RealBuilder`.
- Invoke the Reconciler's `Builder`'s `GetClient()` method to grab a new AWS Client.

Tests should do one of two things, depending on their scope:
- When testing `Reconcile` or other methods with a Reconciler receiver, initialize the receiver's `Builder` field to a pointer to an instance of `MockBuilder`. To `EXPECT()` things, get the mocked Client in your test via the `GetMockClient` helper.

- When testing a func that accepts a `Client`, that `Client` can still be created via `mock.NewMockClient()` as before.

This commit makes every effort **not** to change any logic. There are places where clients could potentially be cached/reused, places (such as totalaccountwatcher) that don't follow the above pattern perfectly, etc. Those are left for future work.

Of particular note, accountclaim_controller_test.go was previously making real AWS calls implicitly. **It still is**, but now it's explicit in that the Reconciler it's using is being initialized with a `RealBuilder`. A subsequent commit will mock this properly.

Jira: [OSD-4575](https://issues.redhat.com/browse/OSD-4575)

### DIFF
--- a/pkg/awsclient/client.go
+++ b/pkg/awsclient/client.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/service/route53/route53iface"
 	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+	"github.com/golang/mock/gomock"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
@@ -36,6 +37,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/sts/stsiface"
 	"github.com/aws/aws-sdk-go/service/support"
 	"github.com/aws/aws-sdk-go/service/support/supportiface"
+	"github.com/openshift/aws-account-operator/pkg/awsclient/mock"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	kubeclientpkg "sigs.k8s.io/controller-runtime/pkg/client"
@@ -354,8 +356,8 @@ func (c *awsClient) ChangeResourceRecordSets(input *route53.ChangeResourceRecord
 	return c.route53client.ChangeResourceRecordSets(input)
 }
 
-// NewClient creates our client wrapper object for the actual AWS clients we use.
-func NewClient(awsAccessID, awsAccessSecret, token, region string) (Client, error) {
+// newClient creates our client wrapper object for the actual AWS clients we use.
+func newClient(awsAccessID, awsAccessSecret, token, region string) (Client, error) {
 	awsConfig := &aws.Config{Region: aws.String(region)}
 	awsConfig.Credentials = credentials.NewStaticCredentials(
 		awsAccessID, awsAccessSecret, token)
@@ -376,12 +378,21 @@ func NewClient(awsAccessID, awsAccessSecret, token, region string) (Client, erro
 	}, nil
 }
 
-// GetAWSClient generates an awsclient
+// Builder implementations know how to produce a Client.
+type Builder interface {
+	GetClient(kubeClient kubeclientpkg.Client, input NewAwsClientInput) (Client, error)
+}
+
+// RealBuilder is a Builder implementation that knows how to produce a real AWS Client (i.e. one
+// that really talks to the AWS APIs).
+type RealBuilder struct{}
+
+// GetClient generates a real awsclient
 // function must include region
 // Pass in token if sessions requires a token
 // if it includes a secretName and nameSpace it will create credentials from that secret data
 // If it includes awsCredsSecretIDKey and awsCredsSecretAccessKey it will build credentials from those
-func GetAWSClient(kubeClient kubeclientpkg.Client, input NewAwsClientInput) (Client, error) {
+func (rp *RealBuilder) GetClient(kubeClient kubeclientpkg.Client, input NewAwsClientInput) (Client, error) {
 
 	// error if region is not included
 	if input.AwsRegion == "" {
@@ -410,7 +421,7 @@ func GetAWSClient(kubeClient kubeclientpkg.Client, input NewAwsClientInput) (Cli
 				input.SecretName, awsCredsSecretAccessKey)
 		}
 
-		awsClient, err := NewClient(string(accessKeyID), string(secretAccessKey), input.AwsToken, input.AwsRegion)
+		awsClient, err := newClient(string(accessKeyID), string(secretAccessKey), input.AwsToken, input.AwsRegion)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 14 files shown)


pkg/awsclient/client.go
```diff
@@ -354,8 +356,8 @@ func (c *awsClient) ChangeResourceRecordSets(input *route53.ChangeResourceRecord
 	return c.route53client.ChangeResourceRecordSets(input)
 }
 
-// NewClient creates our client wrapper object for the actual AWS clients we use.
-func NewClient(awsAccessID, awsAccessSecret, token, region string) (Client, error) {
+// newClient creates our client wrapper object for the actual AWS clients we use.
+func newClient(awsAccessID, awsAccessSecret, token, region string) (Client, error) {
```
2uasimojo, line 360: Renamed to make private, because only GetClient (below) should be using this directly.

pkg/awsclient/client.go
```diff
@@ -376,12 +378,21 @@ func NewClient(awsAccessID, awsAccessSecret, token, region string) (Client, erro
 	}, nil
 }
 
-// GetAWSClient generates an awsclient
+// Builder implementations know how to produce a Client.
+type Builder interface {
```
2uasimojo, line 382: Lest you think this name is odd, I originally had it as AWSClientProvider, but my IDE made an excellent point that it would be used as `awsclient.AWSClientProvider`, which "stutters" (it actually used that word).

pkg/controller/accountclaim/accountclaim_controller_test.go
```diff
@@ -62,7 +63,12 @@ var _ = Describe("AccountClaim", func() {
 
 	Context("Reconcile", func() {
 		It("should reconcile correctly", func() {
-			r = &ReconcileAccountClaim{client: fakeClient, scheme: scheme.Scheme}
+			r = &ReconcileAccountClaim{
+				client: fakeClient,
+				scheme: scheme.Scheme,
+				// TODO(efried): Mock this!
```
2uasimojo, line 69: To be done in #419, coming soon.

pkg/controller/accountclaim/accountclaim_controller_test.go
```diff
@@ -102,7 +108,12 @@ var _ = Describe("AccountClaim", func() {
 				fakeClient,
 				true,
 			}
-			r = &ReconcileAccountClaim{client: cl, scheme: scheme.Scheme}
+			r = &ReconcileAccountClaim{
+				client: cl,
+				scheme: scheme.Scheme,
+				// TODO(efried): Mock this!
```
2uasimojo, line 114: Ditto.

pkg/totalaccountwatcher/totalaccountwatcher_test.go
```diff
@@ -114,7 +114,6 @@ func TestTotalAwsAccounts(t *testing.T) {
 
 			// Act
 			TotalAccountWatcher = NewTotalAccountWatcher(mocks.fakeKubeClient, mocks.mockAWSClient, 10)
-			TotalAccountWatcher.AwsClient = mocks.mockAWSClient
```
2uasimojo, line 117: Unrelated change, but this was redundant since `NewTotalAccountWatcher` does exactly this via the second arg.

test/integration/federated_access_cr_test.go
```diff
@@ -191,7 +191,7 @@ func getSecretCredentials(t *testing.T, secret *awsUserSecret) {
 }
 
 // Gets AWS Client using passed in credentials struct
-func getAWSClient(t *testing.T, awsCreds awsUserSecret) (*iam.IAM, error) {
+func getAWSIAMClient(t *testing.T, awsCreds awsUserSecret) (*iam.IAM, error) {
```
2uasimojo, line 194: Almost-unrelated change, but I wanted to rename this so future me wasn't confused about what kind of "client" this was returning.

rogbas:
@rogbas can you please review/test locally?

pkg/awsclient/client.go
```diff
@@ -376,12 +378,21 @@ func NewClient(awsAccessID, awsAccessSecret, token, region string) (Client, erro
...
+	GetClient(kubeClient kubeclientpkg.Client, input NewAwsClientInput) (Client, error)
+}
+
+// RealBuilder is a Builder implementation that knows how to produce a real AWS Client (i.e. one
+// that really talks to the AWS APIs).
+type RealBuilder struct{}
```
rafael-azevedo, line 388: The only comment I have on this is the RealBuilder struct. I am not in love with the naming convention here. Not sure if there is anything a better we could use.

rafael-azevedo: Outside of this implementation looks solid and hopefully will make testing easier on these reconciles.

2uasimojo: > The only comment I have on this is the RealBuilder struct. I am not in love with the naming convention here. Not sure if there is anything a better we could use.

Yeah, I agree it's not ideal. I played with a couple of other ideas (e.g. calling the interface `IBuilder` so I could call the real builder `Builder`) but ended up hating this one the least. Happy to take suggestions.

rafael-azevedo: I do like `IBuilder `, `Builder `better maybe even `BuilderInt `and `Builder`

rogbas: `BuilderInt` is confusing IMO, could be associated with integer. I vote for the `IBuilder` for the interface and Builder for the real obj

2uasimojo: Roger(io) that.

I'm going to do it in a followon PR, since this one is already reviewed for content, and to avoid having to un-conflict the two PRs that are riding on top of this one. Cool?

rogbas:
Overall the PR LGTM

pkg/awsclient/client.go
rogbas: awesome @2uasimojo, thanks

PR #424, opened 2020-07-21
Fixes bug preventing BYOCAdminAccess role creating failure from causing an immediate reconcile; condition update for accountClaim

clcollins:
This fixes a bug that prevents a failure to create the BYOCAdminAccess role from immediately returning, logging and restarting the reconcile loop for CCS accounts. Previously, if the creation of the role failed, the CCS initialization would just continue on.

This also fixes condition messages for the CCS accountClaims, so that they are updated as errors occur, as opposed to keeping the same error message no matter what.  This prevented OCM/users from seeing the actual error message causing their CCS account to fail.

This is a direct impact on https://issues.redhat.com/browse/OSD-3978 "report more granular status for OCM", allowing the failure to surface, and repeated errors to be reported (updating the condition, as opposed to leaving it as failed with the original message).

### DIFF
--- a/pkg/controller/account/byoc.go
+++ b/pkg/controller/account/byoc.go
@@ -16,7 +16,7 @@ import (
 
 	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
 	"github.com/openshift/aws-account-operator/pkg/awsclient"
-	controllerutils "github.com/openshift/aws-account-operator/pkg/controller/utils"
+	"github.com/openshift/aws-account-operator/pkg/controller/utils"
 )
 
 const (
@@ -84,6 +84,7 @@ func (r *ReconcileAccount) initializeNewBYOCAccount(reqLogger logr.Logger, curre
 	if !accountHasState(currentAcctInstance) {
 		tags := awsclient.AWSTags.BuildTags(currentAcctInstance).GetIAMTags()
 		roleID, err = createBYOCAdminAccessRole(reqLogger, awsSetupClient, client, adminAccessArn, currentAccInstanceID, tags)
+
 		if err != nil {
 			r.accountClaimBYOCError(reqLogger, accountClaim, err)
 			return "", err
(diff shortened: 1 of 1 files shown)


clcollins:
/hold

unsure why the prow tests are failing on the `aws` command missing.  Will check it out in the am

pkg/controller/account/byoc.go
```diff
@@ -373,3 +372,8 @@ func validateBYOCClaim(accountClaim *awsv1alpha1.AccountClaim) error {
 
 	return nil
 }
+
+// Print a human readable error in the operator logs for debugging
+func printErr(location string, message error) {
+	fmt.Printf("\n+++++ In %s: %s +++++\n\n", location, message)
+}
```
rogbas, line 379: @clcollins can you please remove this debug function?

clcollins: @rogbas Removed. Sorry for looking over it the first time.

pkg/controller/account/byoc.go
```diff
@@ -141,10 +142,9 @@ func createBYOCAdminAccessRole(reqLogger logr.Logger, awsSetupClient awsclient.C
 
 	// Create the base role
 	roleID, croleErr := CreateRole(reqLogger, byocInstanceIDRole, principalARN, byocAWSClient, tags)
-	if err != nil {
+	if croleErr != nil {
 		return roleID, croleErr
 	}
-	reqLogger.Info(fmt.Sprintf("New RoleID: %s", roleID))
```
rogbas, line 147: Are we losing this log msg on purpose? I don't see anywhere else where we output the RoleID of the created role

2uasimojo:
I have looked at this. Other than @rogbas's question, it lgtm.

pkg/controller/account/byoc.go
clcollins: It did not seem helpful to me in the context of how our logging works within the pod.  It seemed to me to be more related to debugging than logging helpful information.

That said, I will put that line back in.

clcollins: Reverted the removal of this log message.

clcollins:
@rogbas @2uasimojo Thanks for your reviews.  I reverted the removal of that log message (did not feel strongly about the change).

I also noticed right after that a spot where we are not handling a potential error, and added a handler to that.

https://github.com/openshift/aws-account-operator/pull/424/files#diff-aa743eaa1e82c1f0f3ed118da91e6173R152-R159

I would appreciate your final review on this when you get a chance.  Thanks!

PR #425, opened 2020-07-22
OSD 3978 granular status reports

clcollins:
This includes a number of changes to facilitate surfacing of underlying account failure messages to OCM for CCS accounts, as well as more details error messages for standard accounts upon failure.  It also renames some "BYOC" items to "CCS", and consolidates some code used between both standard and ccs accounts into one place.

This incorporates the bugfixes from #424, and must be merged after.

The complete changelog:

- Adds function to report failed accounts to account claim, retrieve account claim to the account controller
- Removes byoc-specific function for the above, so both types share the same code
- Renamed "SetAccountStatus" to "SetAccountCondition" to match actual workings; renamed "BYOCAccountFailedClaim" to "CCSAccountFailedClaim"
- Renames "setStatusFailed" to "setStateFailed"
- expand use of "statusUpdate" not previously utilized widely in account controller
- creates utils/status.go and moves "SetAccountStatus" to it
- Changes BYOC to CCS in "accountClaimBYOCError" message
- Utilize new constant ClaimedAccountFailed for BYOC account fail message; replace existing "AccountFailed" string with constant
- adds new conditions for accountClaims whose underlying account has failed
- handles missing accountClaims (for standard accounts) and doesn't try to update the accountClaim condition for them

Fixes (or improves toward) https://issues.redhat.com/browse/OSD-3978 "report more granular status for OCM"

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -18,6 +18,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	kubeclientpkg "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -27,6 +28,7 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	"github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
 	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
 	"github.com/openshift/aws-account-operator/pkg/awsclient"
 	totalaccountwatcher "github.com/openshift/aws-account-operator/pkg/totalaccountwatcher"
@@ -53,7 +55,8 @@ const (
 	AccountReady = "Ready"
 	// AccountPendingVerification indicates verification (of AWS limits and Enterprise Support) is pending
 	AccountPendingVerification = "PendingVerification"
-	byocRole                   = "BYOCAdminAccess"
+
+	byocRole = "BYOCAdminAccess"
 
 	adminAccessArn = "arn:aws:iam::aws:policy/AdministratorAccess"
 	iamUserNameUHC = "osdManagedAdmin"
@@ -121,6 +124,7 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 	// Fetch the Account instance
 	currentAcctInstance := &awsv1alpha1.Account{}
 	err := r.Client.Get(context.TODO(), request.NamespacedName, currentAcctInstance)
+
 	if err != nil {
 		if k8serr.IsNotFound(err) {
 			return reconcile.Result{}, nil
@@ -161,6 +165,7 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 	if newBYOCAccount(currentAcctInstance) {
 		byocRoleID, err = r.initializeNewBYOCAccount(reqLogger, currentAcctInstance, awsSetupClient, adminAccessArn)
 		if err != nil || byocRoleID == "" {
+			r.setStateFailed(reqLogger, currentAcctInstance, err.Error())
 			reqLogger.Error(err, "Failed setting up new BYOC account")
 			return reconcile.Result{}, err
 		}
@@ -182,7 +187,7 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 
 					// Update supportCaseId in CR
 					currentAcctInstance.Status.SupportCaseID = caseID
-					SetAccountStatus(reqLogger, currentAcctInstance, "Account pending verification in AWS", awsv1alpha1.AccountPendingVerification, AccountPendingVerification)
+					utils.SetAccountStatus(currentAcctInstance, "Account pending verification in AWS", awsv1alpha1.AccountPendingVerification, AccountPendingVerification)
 					err = r.statusUpdate(reqLogger, currentAcctInstance)
 					if err != nil {
 						return reconcile.Result{}, err
@@ -219,7 +224,7 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 			if resolved {
 				reqLogger.Info(fmt.Sprintf("Case %s resolved", currentAcctInstance.Status.SupportCaseID))
 
-				SetAccountStatus(reqLogger, currentAcctInstance, "Account ready to be claimed", awsv1alpha1.AccountReady, AccountReady)
+				utils.SetAccountStatus(currentAcctInstance, "Account ready to be claimed", awsv1alpha1.AccountReady, AccountReady)
 				err = r.statusUpdate(reqLogger, currentAcctInstance)
 				if err != nil {
 					return reconcile.Result{}, err
@@ -243,7 +248,8 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 
 		// see if in creating for longer then default wait time
 		if accountCreatingTooLong(currentAcctInstance) {
-			r.setStatusFailed(reqLogger, currentAcctInstance, fmt.Sprintf("Creation pending for longer then %d minutes", utils.WaitTime))
+			errMsg := fmt.Sprintf("Creation pending for longer then %d minutes", utils.WaitTime)
+			r.setStateFailed(reqLogger, currentAcctInstance, errMsg)
 		}
 
 		if accountIsUnclaimedAndHasNoState(currentAcctInstance) {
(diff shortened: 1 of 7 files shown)


rogbas:
@rogbas @fahlmant @iamkirkbater @ArielLima please review/cmmnt

pkg/controller/account/account_controller.go
```diff
@@ -631,6 +625,81 @@ func (r *ReconcileAccount) statusUpdate(reqLogger logr.Logger, account *awsv1alp
...
 }
 
+func (r *ReconcileAccount) setStateFailed(reqLogger logr.Logger, currentAcctInstance *awsv1alpha1.Account, msg string) error {
+	reqLogger.Info(msg)
+	// Upodate the account status (state and condition) to Failed
+	utils.SetAccountStatus(reqLogger, currentAcctInstance, msg, v1alpha1.AccountFailed, AccountFailed)
+	err := r.statusUpdate(reqLogger, currentAcctInstance)
+	if err != nil {
+		return err
+	}
```
2uasimojo, line 635: Redundant with [L643-5 below](https://github.com/openshift/aws-account-operator/pull/425/files#diff-32f12988d73ded1726ea7c67c8cb1f90R643-R645). Remove one or the other.

pkg/controller/account/account_controller.go
```diff
@@ -631,6 +625,81 @@ func (r *ReconcileAccount) statusUpdate(reqLogger logr.Logger, account *awsv1alp
...
+
+	// Set a failure condition in the accountClaim
+	err = r.setAccountClaimError(reqLogger, currentAcctInstance, msg)
+	if err != nil {
+		return err
+	}
+
+	// Apply the update
+	err = r.statusUpdate(reqLogger, currentAcctInstance)
+	return err
```
2uasimojo, line 645: Redundant with [L632-5 above](https://github.com/openshift/aws-account-operator/pull/425/files#diff-32f12988d73ded1726ea7c67c8cb1f90R643-R645). Remove one or the other.

pkg/controller/account/account_controller.go
```diff
@@ -631,6 +625,81 @@ func (r *ReconcileAccount) statusUpdate(reqLogger logr.Logger, account *awsv1alp
...
+	err := r.statusUpdate(reqLogger, currentAcctInstance)
+	if err != nil {
+		return err
+	}
+
+	// Set a failure condition in the accountClaim
+	err = r.setAccountClaimError(reqLogger, currentAcctInstance, msg)
+	if err != nil {
+		return err
+	}
```
2uasimojo, line 641: I don't think we should be doing this here at all.

To the extent possible, $foo_controller ought to be exclusively responsible for updating $foo CRs; otherwise it's hard to understand which one is supposed to "win".

And in this case, when you update the Account's Status, it'll trigger a reconcile in the accountclaim_controller, which will see the Account's updated status and make the same status update anyway.

[Later] Okay, I see this is latent, account_controller was doing `accountClaimBYOCError` before. So this might not be the PR to make that change.

pkg/controller/account/account_controller.go
```diff
@@ -631,6 +625,81 @@ func (r *ReconcileAccount) statusUpdate(reqLogger logr.Logger, account *awsv1alp
...
+	accountClaim := &awsv1alpha1.AccountClaim{}
+	err := r.Client.Get(context.TODO(), types.NamespacedName{
+		Name: account.Spec.ClaimLink, Namespace: account.Spec.ClaimLinkNamespace}, accountClaim)
+	if err != nil {
+		return nil, err
+	}
+	return accountClaim, nil
+}
+
+func (r *ReconcileAccount) setAccountClaimError(reqLogger logr.Logger, currentAccountInstance *awsv1alpha1.Account, message string) error {
```
2uasimojo, line 658: As noted above, IMO this should just go away. But if you were going to keep it, why wouldn't it call the corresponding func from the accountclaim package? Circular import? (Which may be another hint that we shouldn't be updating the same object from both places.)

pkg/controller/accountclaim/reuse.go
```diff
@@ -141,7 +140,7 @@ func (r *ReconcileAccountClaim) resetAccountSpecStatus(reqLogger logr.Logger, re
 	reusedAccount.Status.Claimed = false
 	reusedAccount.Status.Reused = true
 	conditionMsg := fmt.Sprintf("Account Reuse - %s", conditionStatus)
-	account.SetAccountStatus(reqLogger, reusedAccount, conditionMsg, accountState, conditionStatus)
+	utils.SetAccountStatus(reqLogger, reusedAccount, conditionMsg, accountState, conditionStatus)
```
2uasimojo, line 143: Similar comment here: it's worrisome that accountclaim is updating Account CRs. But in this case, there would be nothing to trigger the account controller to do this status update.

Again, this is latent, but something that should probably be looked at for the future. It's possible to make the Account controller watch for AccountClaim changes as well (would have to use a map func, since the ownership is going the wrong way). Not sure if that helps here - i.e. if there's an AccountClaim change that the Account controller could recognize in order to set this status.

pkg/controller/utils/status.go
```diff
@@ -0,0 +1,21 @@
+package utils
+
+import (
+	"github.com/go-logr/logr"
+	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// SetAccountStatus sets the condition and state of an account
+func SetAccountStatus(reqLogger logr.Logger, awsAccount *awsv1alpha1.Account, message string, ctype awsv1alpha1.AccountConditionType, state string) {
```
2uasimojo, line 10: I get why you moved this, but IMHO it's more appropriate to keep it in the account package to highlight the fact that the account package ought to own changes to Account CRs.

pkg/controller/account/account_controller.go
```diff
@@ -164,6 +170,9 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 	if newBYOCAccount(currentAcctInstance) {
 		byocRoleID, err = r.initializeNewBYOCAccount(reqLogger, currentAcctInstance, awsSetupClient, adminAccessArn)
 		if err != nil || byocRoleID == "" {
+			errMsg := fmt.Sprint(err)
```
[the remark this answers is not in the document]
clcollins: Thanks, I'm still learning the specifics of Go, and didn't realize this.  I will make these changes.

pkg/controller/account/account_controller.go
```diff
@@ -631,6 +625,81 @@ func (r *ReconcileAccount) statusUpdate(reqLogger logr.Logger, account *awsv1alp
...
+		}
+		reqLogger.Error(err, fmt.Sprintf("Unable to get accountClaim for %s", currentAccountInstance.Name))
+		return err
+	}
+
+	var reason string
+	var conditionType awsv1alpha1.AccountClaimConditionType
+
+	if accountIsBYOC(currentAccountInstance) {
+		message = fmt.Sprintf("CCS Account Failed: %-v", message)
```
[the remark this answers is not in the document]
clcollins: This was copied from the way it had been done by someone else in the code, but I agree: `%s` is better.  I will update that as well.

pkg/controller/account/byoc.go
```diff
@@ -60,20 +59,20 @@ func (r *ReconcileAccount) initializeNewBYOCAccount(reqLogger logr.Logger, curre
 	client, accountClaim, err := r.getBYOCClient(currentAcctInstance)
 	if err != nil {
 		if accountClaim != nil {
-			r.accountClaimBYOCError(reqLogger, accountClaim, err)
+			r.setAccountClaimError(reqLogger, currentAcctInstance, fmt.Sprint(err))
```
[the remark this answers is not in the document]
clcollins: :+1:

pkg/controller/account/account_controller.go
```diff
@@ -54,7 +55,12 @@ const (
 	AccountReady = "Ready"
 	// AccountPendingVerification indicates verification (of AWS limits and Enterprise Support) is pending
 	AccountPendingVerification = "PendingVerification"
-	byocRole                   = "BYOCAdminAccess"
+	// AccountFailedClaim indicates that the claimed account underlying an accoutnClaim has failed to become ready
+	AccountFailedClaim = "ClaimedAccountFailed"
+	// CCSAccountFailedClaim indicates a failure to claim a CCS/BYOC account in the accountClaim status
+	CCSAccountFailedClaim = "ClaimedCCSAccountFailed"
```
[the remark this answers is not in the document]
clcollins: Good point, thank you.  I think that was leftover from some early workings.

pkg/controller/account/account_controller.go
clcollins: I agreed with your point - I am loathe to have one controller make changes to another, but this was the suggested way forward by the team for resolving this.  I've got some notes for future improvements that I'd like make cards for, and this is one of them.

pkg/controller/accountclaim/reuse.go
clcollins: :+1:

pkg/controller/account/account_controller.go
clcollins: Yeah, I was struggling with circular imports.  I will flag this for consideration in a future PR to address across-controller resource handling.

pkg/controller/account/account_controller.go
clcollins: Done!

pkg/controller/account/account_controller.go
clcollins: Done!

pkg/controller/account/account_controller.go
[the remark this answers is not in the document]
2uasimojo: Missed this one :)

clcollins:
Rebased.

/retest

pkg/controller/account/byoc.go
```diff
@@ -289,9 +288,9 @@ func DetachPolicyFromRole(reqLogger logr.Logger, policy *iam.AttachedPolicy, byo
 			default:
 				reqLogger.Error(
 					aerr,
-					fmt.Sprintf(aerr.Error()),
+					aerr.Error(),
 				)
-				reqLogger.Error(err, fmt.Sprintf("%v", err))
+				reqLogger.Error(err, err.Error())
```
2uasimojo, line 293: nit: this will end up printing the same error twice. This could just be

```suggestion
				reqLogger.Error(err, "")
```

...to the same effect.

(Here and below.)

clcollins: Thanks!  I'm refactoring how the errors are handles in the byoc.go for another PR, and will make those changes there.

rogbas:
@rogbas @fahlmant please check

2uasimojo:
[Rebase diff](https://github.com/openshift/aws-account-operator/compare/cb8a40af2a450fe7f325707e139bda45d5e7cc24..5872f8dc4262134e156d3d9749998b5283403df2) is ugh, but clean.

/lgtm

PR #426, opened 2020-07-22
Adds a default value for current accounts.

iamkirkbater:
Adds a very high value for the intial account count so that we stop creating accounts on initialization failures.

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -253,6 +253,10 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 
 			if !accountHasAwsAccountID(currentAcctInstance) {
 				// before doing anything make sure we are not over the limit if we are just error
+				if totalaccountwatcher.TotalAccountWatcher.Total == 0 {
+					reqLogger.Error(awsv1alpha1.ErrAccountWatcherNoTotal, "AccountWatcher has not run successfully yet")
+					return reconcile.Result{}, awsv1alpha1.ErrAccountWatcherNoTotal
+				}
 				if ok, _ := checkAWSAccountsLimitReached(r, reqLogger, totalaccountwatcher.TotalAccountWatcher.Total); ok {
 					reqLogger.Error(awsv1alpha1.ErrAwsAccountLimitExceeded, "AWS Account limit reached", "Account Total", totalaccountwatcher.TotalAccountWatcher.Total)
 					return reconcile.Result{}, awsv1alpha1.ErrAwsAccountLimitExceeded
--- a/pkg/totalaccountwatcher/totalaccountwatcher.go
+++ b/pkg/totalaccountwatcher/totalaccountwatcher.go
@@ -88,6 +88,7 @@ func (s *totalAccountWatcher) UpdateTotalAccounts(log logr.Logger) error {
 	accountTotal, err := TotalAwsAccounts()
 	if err != nil {
 		log.Error(err, "Failed to get account list with error code")
+		return nil
 	}
 	localmetrics.Collector.SetTotalAWSAccounts(accountTotal)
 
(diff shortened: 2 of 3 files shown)


pkg/totalaccountwatcher/totalaccountwatcher.go
```diff
@@ -88,6 +89,7 @@ func (s *totalAccountWatcher) UpdateTotalAccounts(log logr.Logger) error {
 	accountTotal, err := TotalAwsAccounts()
 	if err != nil {
 		log.Error(err, "Failed to get account list with error code")
+		return nil
```
2uasimojo, line 92: This is goodness regardless. IIUC previously an error would cause L94 to reset the total number of accounts back down to zero :astonished:

pkg/totalaccountwatcher/totalaccountwatcher.go
```diff
@@ -48,6 +48,7 @@ func Initialize(client client.Client, watchInterval time.Duration) {
 	}
 
 	TotalAccountWatcher = NewTotalAccountWatcher(client, AwsClient, watchInterval)
+	TotalAccountWatcher.Total = 99999 // Init this with a stupid high value to account for multiple aws calls on multiple account watchers starting at the same time.
```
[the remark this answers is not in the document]
iamkirkbater: Yeah, it's definitely a low-LOE fix.

I recently proposed making the TotalAccountWatcher have the methodology which gates account creation based on the number of accounts.

Basic Idea: We have a new var inside of the TotalAccountWatcher called (naming is hard so this is just the idea) `TotalAccountWatcher.AccountsCanBeCreated` which is a bool.  Then account_controller doesn't do any of the logic to get the total number of accounts and compare it to the limit, that's handled within account_controller and that bool is updated.  That bool would default to false and accounts cannot be created until the initial run of the AccountWatcher is completed.

There would also need to be an update hook here so that we could tell the Total AccountWatcher that we've created an account so it can temporarily update the amount of accounts for the edge-cases when we're nearing the limit of accounts we have set.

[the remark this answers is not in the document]
rogbas: @2uasimojo yeah, makes sense, it's a cleaner approach and quick to implement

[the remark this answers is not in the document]
iamkirkbater: I have to check for a zero-value here because int cannot be compared to nil, but we just have to keep this in mind if we ever run this on a fresh account, but I'm not even sure that would come in at 0 because it would at least have itself in the account?  Not sure, but we should just keep it in the back of our collective mind in case it ever comes up.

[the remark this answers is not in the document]
2uasimojo: Not comparing the total to `nil`, comparing the whole object to `nil`. Let's go over this tomorrow.

pkg/totalaccountwatcher/totalaccountwatcher.go
2uasimojo: Wait, I take it back, this should be

```suggestion
		return err
```

so that the log on L76 happens.

pkg/controller/account/account_controller.go
```diff
@@ -253,6 +253,10 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 
 			if !accountHasAwsAccountID(currentAcctInstance) {
 				// before doing anything make sure we are not over the limit if we are just error
+				if totalaccountwatcher.TotalAccountWatcher.Total == 0 {
```
2uasimojo, line 256: I don't really feel like 0 is a whole lot better than 99999.

My suggestion about leaving the `TotalAccountWatcher` global `nil` until it's successfully initialized would require a bigger refactor, but a simpler version meets you halfway to your `AccountsCanBeCreated` field: Add an `Initialized` bool field that you only set to `true` once `UpdateTotalAccounts` has succeeded, and check that here instead of the `Total`.

2uasimojo: Demonstrated [here](https://github.com/iamkirkbater/aws-account-operator/pull/1)

rogbas: +1 to the suggestion with the `Initialized` bool, it prevents a bug where we're using the Payer Account for the first time and it'll have actually ZERO accounts in it

PR #428, opened 2020-07-24
WIP/PoC: Use reconcile timer from operator-custom-metrics

2uasimojo:
Replace in-repo X with X from operator-custom-metrics for X in:
- Definition of the `aws_account_operator_reconcile_duration_seconds` metric.
- Observation of that metric in `Reconcile`s

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -15,6 +15,7 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/openshift/aws-account-operator/pkg/controller/utils"
 	"github.com/openshift/aws-account-operator/pkg/localmetrics"
+	"github.com/openshift/operator-custom-metrics/pkg/wrappers"
 	corev1 "k8s.io/api/core/v1"
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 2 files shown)


2uasimojo:
PoC for now, showing how we would cut over to using https://github.com/openshift/operator-custom-metrics/pull/29 once it's released. Obviously not going to work until then.

pkg/controller/account/account_controller.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/openshift/aws-account-operator/pkg/controller/utils"
 	"github.com/openshift/aws-account-operator/pkg/localmetrics"
+	"github.com/openshift/operator-custom-metrics/pkg/wrappers"
```
2uasimojo, line 18: Ultimately, this file's delta would need to be applied similarly to all the controllers.

pkg/controller/account/account_controller.go
```diff
@@ -98,6 +104,10 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 var _ reconcile.Reconciler = &ReconcileAccount{}
 
 // ReconcileAccount reconciles a Account object
+type reconcileAccount struct {
+	wrappers.TimedReconciler
+}
+
```
2uasimojo, line 110: whoops, whack this.
```suggestion
```

pkg/controller/account/account_controller.go
```diff
@@ -71,10 +72,15 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
-	return &ReconcileAccount{
-		Client:           utils.NewClientWithMetricsOrDie(log, mgr, controllerName),
-		scheme:           mgr.GetScheme(),
-		awsClientBuilder: awsclient.GetAWSClient,
+	return &wrappers.TimedReconciler{
```
rafael-azevedo, line 75: I'm hesitant to wrap the boiler plate code with a dependency in another library. If this is functionality that needs to be default for all the operators maybe it should be contributed upstream in the SDK?

2uasimojo: > I'm hesitant to wrap the boiler plate code

I'm open to suggestions on a different pattern to use here, but this isn't boilerplate code. We're already instrumenting the controller-runtime client and adding an AWS client to it. This is effectively the same thing as the former.

> maybe it should be contributed upstream in the SDK?

We've talked about that, and it's a good possibility, but it would be quite a long time before we would be able to use that here: It has to get proposed, reviewed, iterated on, merged, and released; and *then* a-a-o has to upgrade to whatever release of operator-sdk it winds up in.

PR #434, opened 2020-07-31
Update account state to reflect generic AWS support case pending

jharrington22:
This PR updates the account CR state when waiting for AWS support cases to be resolved to be more generic. We no longer request quota increases and only enable enterprise support, this message suits the enterprise support status and any other future tickets we might add.

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -52,9 +52,9 @@ const (
 	AccountFailed = "Failed"
 	// AccountReady indicates account creation is ready
 	AccountReady = "Ready"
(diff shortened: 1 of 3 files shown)


jharrington22:
/hold

To test.

2uasimojo:
> @jharrington22: The following test **failed**, say `/retest` to rerun all failed tests:

Test fails are because the [API const](https://github.com/openshift/aws-account-operator/blob/master/pkg/apis/aws/v1alpha1/account_types.go#L98-L99) needs to be changed as well.

rogbas:
@rogbas will resubmit this PR fixing existing issues.

PR #436, opened 2020-08-03
Removes error on byoc secret finalizer

iamkirkbater:


pkg/controller/accountclaim/reuse.go
```diff
@@ -99,6 +100,10 @@ func (r *ReconcileAccountClaim) finalizeAccountClaim(reqLogger logr.Logger, acco
 		// Cleanup BYOC secret
 		err = r.removeBYOCSecretFinalizer(accountClaim)
 		if err != nil {
+			// If the secret can't be found, don't error, just return
+			if k8serr.IsNotFound(err) {
+				return nil
```
2uasimojo, line 105: Do we want to log here?

2uasimojo: Never mind, what confused me was the comment, which should really say "If the finalizer can't be found..."

fahlmant: If the secret can't be found, then we don't want to return an error.

2uasimojo: Sorry, I get it now. If the secret doesn't exist, it doesn't have a finalizer to remove, so this is n/a.

So then, shouldn't we push this check into `removeBYOCSecretFinalizer` itself? Otherwise, we need to do this same check in the two other places it's called (in reuse.go).

fahlmant: Good point

iamkirkbater: Updated.

PR #439, opened 2020-08-06
Refactors Account Struct to have methods

iamkirkbater:
Moves the helper methods out of the Account Controller and BYOC files in
order to put them directly on the struct where they make more sense.

### DIFF
--- a/pkg/apis/aws/v1alpha1/account_types.go
+++ b/pkg/apis/aws/v1alpha1/account_types.go
@@ -1,6 +1,8 @@
 package v1alpha1
 
 import (
+	"time"
+
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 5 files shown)


pkg/apis/aws/v1alpha1/account_types.go
```diff
@@ -131,3 +131,109 @@ type AccountList struct {
 func init() {
 	SchemeBuilder.Register(&Account{}, &AccountList{})
 }
+
```
2uasimojo, line 134: Minor suggestion: Could put these into a separate file in the same package. But meh.

pkg/apis/aws/v1alpha1/account_types.go
```diff
@@ -131,3 +131,109 @@ type AccountList struct {
...
+// Helper Functions
+
+//IsFailed returns true if an account is in a failed state
+func (a *Account) IsFailed() bool {
+	return a.Status.State == string(AccountFailed)
```
2uasimojo, line 139: This has technically changed, and continues to work only because the `AccountFailed` `AccountConditionType` in this package has the same string value as the `AccountFailed` const in `pkg/controller/account` defined in account_controller.go. Similar comment applies to the other methods below.

I've mentioned [before](https://github.com/openshift/aws-account-operator/pull/425#discussion_r459042894) that it's awkward/weird to have those strings duplicated for `Reason` and `Status`. I'm not sure if this is the right place to address that, though.

pkg/controller/account/account_controller.go
```diff
@@ -558,105 +558,11 @@ func matchSubstring(roleID, role string) (bool, error) {
...
-
-func accountHasClaimLink(currentAcctInstance *awsv1alpha1.Account) bool {
-	return currentAcctInstance.Spec.ClaimLink != ""
-}
-
 func accountCreatingTooLong(currentAcctInstance *awsv1alpha1.Account) bool {
-	return accountIsCreating(currentAcctInstance) &&
+	return currentAcctInstance.IsCreating() &&
 		time.Now().Sub(currentAcctInstance.GetCreationTimestamp().Time) > createPendTime
 }
```
2uasimojo, line 564: Why not this one?

pkg/controller/account/account_controller.go
```diff
@@ -558,105 +558,11 @@ func matchSubstring(roleID, role string) (bool, error) {
...
-
-// Return true if account state is AccountCreating and has not been claimed
-func accountIsUnclaimedAndIsCreating(currentAcctInstance *awsv1alpha1.Account) bool {
-	return accountIsCreating(currentAcctInstance) &&
-		!accountIsClaimed(currentAcctInstance)
-}
-
 type accountInstance interface {
 	accountIsUnclaimedAndIsCreating() bool
 }
```
2uasimojo, line 568: I think we can trash this now.

pkg/controller/account/account_controller.go
iamkirkbater: Specifically because `createPendTime` is defined here in the controller.

I probably _could_ modify the function to take a duration and then compare against that, so then we don't have to define the pending time within the account CR (which I don't think would be a good place for it).  I might just do that 🤔

2uasimojo:
Thanks for the updates @iamkirkbater.

/lgtm

iamkirkbater:
/hold

There's a bit of work still to be done with some of the recent changes that have gone in with InitRegions.

rogbas:
@rogbas please test locally and lgtm

pkg/controller/utils/conditions.go
```diff
@@ -9,6 +9,9 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// =====
+// TODO This entire UpdateConditionCheck Block could probably be refactored into the structs by way of it's own package.
+
```
iamkirkbater, line 14: Most of this file looks like it is a great candidate to be refactored into the structs themselves, but I felt like that is too much for this PR, and each of the structs would probably be it's own card.

iamkirkbater:
/hold cancel

All remaining work has been completed, this should be ready for primetime now :)

pkg/controller/account/account_controller.go
```diff
@@ -167,13 +167,13 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
 		reqLogger.Info(fmt.Sprintf("Account %s is failed. Ignoring.", currentAcctInstance.Name))
 		return reconcile.Result{}, nil
 	}
 
 	// Detect accounts for which we kicked off asynchronous region initialization
-	if accountIsInitializingRegions(currentAcctInstance) {
+	if currentAcctInstance.IsInitializingRegions() {
 		// Detect whether we set the InitializingRegions condition in *this* invocation of the
 		// operator or a previous one.
 		if regionInitStale(currentAcctInstance) {
```
2uasimojo, line 179: This (`regionInitStale`) could also be method-ified.

2uasimojo (commented):
Just a note about the squash label: it'll make the commit message (the one that shows up in the git history) kind of ugly. So probably best to do the squash locally.

pkg/controller/account/account_controller.go
iamkirkbater: At first pass I tried that as well, but the problem lies with trying to import `utils` to get the operator start time, because `utils` imports `awsv1alpha1` and when you try to get the operator start time from utils it causes a cyclical dependency.  If you have a suggestion as to how to get around that without another relatively major refactor, then I'd be happy to try it.

iamkirkbater:
Oh boo :(  That's what I get for trying to do this the easy way :( [re: squash-commit message]

pkg/controller/account/account_controller.go
2uasimojo: Bleh, yeah, that sucks. I played with it a little bit too, but there's no easy answer. It would be nice to add a TODO to the method to figure it out for the future, but meh.

2uasimojo (commented):
Noticed one more thing, fix if you wish (are you respinning to squash anyway?) but nonblocking.

/lgtm

pkg/controller/account/account_controller.go
iamkirkbater: We _could_ pull out the operator startup thing into its own package, which should work.  Or if you add that to the bootstrap package you're working on we could use that instead?  Or we could propose that to the upstream maybe?

2uasimojo: I suspect it's too hacky for upstream, but maybe worth a shot.

An isolated package could work, though again kind of hacky.

We're *far* away from having actual go code in boilerplate.

2uasimojo (commented):
still
/lgtm

rogbas:
@iamkirkbater thanks for the refactor, it's much clearer to read/follow.

Reviewed and tested locally.

/lgtm

PR #440, opened 2020-08-06
Factor out Reconciler metrics

2uasimojo:
We're about to add a success/error dimension to the Reconciler duration metric. Doing so inline would entail creating a wrapper func for each controller's Reconcile. Each such wrapper func would have the same code in it.

Instead, prepare by factoring the existing reporting into a utility. This will allow the subsequent commit to add the new dimension in one place rather than having to touch all the controllers.

The utility exposes a wrapper factory, `NewReconcilerWithMetrics`, that accepts and returns a Reconciler. Its job is to
- Log the start & completion messages.
- Time the wrapped Reconciler's Reconcile func and report it to the appropriate prometheus Collector.

Now controllers no longer need to do the above. Instead they simply wrap their locally-defined Reconciler instance via the new util func.

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -14,7 +14,6 @@ import (
 	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/go-logr/logr"
 	"github.com/openshift/aws-account-operator/pkg/controller/utils"
-	"github.com/openshift/aws-account-operator/pkg/localmetrics"
 	corev1 "k8s.io/api/core/v1"
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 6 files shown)


iamkirkbater:
Overall this lgtm to me, but I still have to test it.

I wonder if something like this would be a good candidate to also contribute upstream to the operator-sdk as well?

2uasimojo:
> I wonder if something like this would be a good candidate to also contribute upstream to the operator-sdk as well?

Definitely. We've been talking about how to strategerizify that.

2uasimojo:
Responding to this, asked in #441:

> Other question, would this also satisfy #428, or are these separate issues I'm getting mixed up?

Yes, good eye. The objective is exactly the same: factor out common code so we don't have to repeat it in every controller. This PR does it in a slightly lighter way than the PoC proposed at https://github.com/openshift/operator-custom-metrics/pull/29. Of particular note, the metric isn't part of the wrapper struct; it's hardcoded into the Reconcile impl. But part of the goal of this change is to make the lift easier if we do wind up with this stuff in a common place.

iamkirkbater:
`make test-all` worked without any problems for me.

PR #442, opened 2020-08-06
Deploy accountpool cr cicd

ArielLima:
Add AccountPool CR to the template used by CICD, having parameters per environment to keep the accountLimit
https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/7578

### DIFF
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -62,6 +62,6 @@ objects:
   kind: AccountPool
   metadata:
     name: aws-account-operator-accountpool
-    namespace: aws-account-operator-accountpool-namespace
+    namespace: aws-account-operator
   spec:
     poolSize: "${ACCOUNT_POOL_SIZE}"
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -16,6 +16,10 @@ parameters:
   required: true
 - name: BASE_OU_ID
   required: true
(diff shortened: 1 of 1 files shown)


hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -55,3 +57,11 @@ objects:
     account-limit: "${ACCOUNT_LIMIT}"
     root: ${ROOT_OU_ID}
     base: ${BASE_OU_ID}
+
+- apiVersion: v1
+  kind: AccountPool
+  metadata:
+    name: aws-account-operator-accountpool
+    namespace: aws-account-operator-accountpool-namespace
```
fahlmant, line 65: I think this can be changed to `aws-account-operator` or be removed

ArielLima:
https://issues.redhat.com/browse/OSD-4602

hack/olm-registry/olm-artifacts-template.yaml
ArielLima: oops that was a mistake on my part, nice catch

iamkirkbater:
So if I'm understanding this correctly, this olm-artifacts file will tell OLM what to deploy?

This structure looks correct and good to me, but I'll let someone who knows what they're really looking at review as well.

fahlmant:
Maybe we should seperate the template out to have OLM artifacts and other (configmap, account pool, federated roles)

ArielLima:
@fahlmant When we were doing this last time with the configmap @maorfr thought it would be best if all templates remained in one place -> https://github.com/openshift/aws-account-operator/pull/406

PR #441, opened 2020-08-06
Reconciler metrics: count errors

2uasimojo:
Add a new `error` label to the `aws_account_operator_reconcile_duration_seconds` metric. This gets the value `true` or `false` according to whether the wrapped `Reconcile` returned a non-nill `error`.

PR #443, opened 2020-08-06
Consolidate duplicate pkg/controller/utils imports

2uasimojo:
There were a few files where pkg/controller/utils was being imported twice with different aliases. Consolidate.

Note that in these cases I chose to retain the `controllerutils` alias, since `utils` is so generic. However, I didn't go back through and change all the other files where that alias isn't used.


2uasimojo:
/hold

I put this on top of #441 to avoid a messier rebase, and wanting to merge that one first since it's more important. Rebase this once #441 merges. Meanwhile, [review just this change](https://github.com/openshift/aws-account-operator/pull/443/commits/0f0153606442add5d79ab8449b3eb0309fd2db45).

clcollins (approved):
/lgtm

@2uasimojo looks like #441 has merged.  I think the HOLD can be removed?

2uasimojo:
Thanks @clcollins. Rebased.

/hold cancel

fahlmant (approved):
/lgtm

PR #447, opened 2020-08-07
Trivial: typos

2uasimojo:
One of these will actually make it easier to trace logs. The other two I
just happened to noticed along the way.


ArielLima:
+1

PR #448, opened 2020-08-07
WIP: Fix AWS API call metrics

2uasimojo:
Use a better key and a threadsafe tracker for AWS API call timers.

Jira: [OSD-4763](https://issues.redhat.com/browse/OSD-4763)


Makefile
```diff
@@ -284,7 +284,7 @@ deploy-aws-account-operator-credentials: check-aws-credentials
 .PHONY: predeploy-aws-account-operator
 predeploy-aws-account-operator:
 # Create aws-account-operator namespace
-	@oc get namespace ${NAMESPACE} || oc new-project ${NAMESPACE}
+	@oc get namespace ${NAMESPACE} && oc project ${NAMESPACE} || oc new-project ${NAMESPACE}
```
2uasimojo, line 287: unrelated change, can split

pkg/awsclient/client.go
```diff
@@ -376,15 +407,22 @@ func newClient(controllerName, awsAccessID, awsAccessSecret, token, region strin
...
 		// Complete is the last phase of a request. Stop the timer as late as possible (as the last handler there).
 		s.Handlers.Complete.PushBack(func(r *request.Request) {
-			defer func() { delete(timers, r) }()
-			localmetrics.Collector.AddAPICall(controllerName, r.HTTPRequest, r.HTTPResponse, time.Since(timers[r]).Seconds())
+			dur := timers.end(r.RequestID)
+			if dur > 0 {
+				localmetrics.Collector.AddAPICall(controllerName, r.HTTPRequest, r.HTTPResponse, dur)
+			} else {
+				localmetrics.Collector.AddAPICall(controllerName, r.HTTPRequest, r.HTTPResponse, 123456)
+
```
2uasimojo, line 424: Not actually going to do this, just using it as an eyecatcher so I can tell whether the RequestID is working (it isn't). We should really have logging here, but that'll take some additional work.

pkg/awsclient/client.go
```diff
@@ -376,15 +407,22 @@ func newClient(controllerName, awsAccessID, awsAccessSecret, token, region strin
...
 		// NOTE(efried): We can't count on the returned client being used by a single thread, so we
 		// create a separate timer for each request.
-		timers := make(map[*request.Request]time.Time)
+		timers := &timerMap{
+			theMap: make(map[string]time.Time),
+		}
 		// Validate is the first phase of a request. Start the timer as early as possible (as the first handler there).
 		s.Handlers.Validate.PushFront(func(r *request.Request) {
-			timers[r] = time.Now()
+			timers.start(r.RequestID)
```
2uasimojo, line 415: Initial testing shows that this RequestID isn't set at this point. I am sad. I'll try using a later-stage Handler. If that doesn't work, I may go back to using the Request's memory address -- that may be okay if I just add the thread safety -- but I sure would like to find something better.

PR #454, opened 2020-08-17
Make region init fully asynchronous

2uasimojo:
TL;DR: free up the Account controller to carry on servicing events for
other Accounts while regions are being initialized.

Background: Initializing supported regions can take a nontrivial amount
of time.  Previously the regions were initialized in parallel, but the
parent account controller Reconcile would wait for all of those threads
to complete.  Thus the controller's ability to service the next event
was constrained to however long the longest region init took. As
currently written, this is a theoretical maximum of 50 minutes if we hit
all the timeouts, but is always at least four seconds plus four AWS API
calls.

With this commit, we set a new state, InitializingRegions, and decouple
the parent of these threads from the main account controller Reconcile,
spawning it in its own thread and immediately returning with no requeue.
The new parent thread is responsible for the final state transition once
all the per-region threads complete. (This state transition would be
what triggers subsequent Reconciles, e.g. to continue watching for
verification of non-CCS Accounts.) The main Reconcile ignores an Account
in the InitializingRegions state, allowing the thread to work, unless it
stays in that state for way too long, in which case we time it out and
transition it to Failed. (Note that if, by some chance, the thread is
still running, its final state transition will bounce because its
`resourceVersion` will be stale due to the Failed transition from the
main controller.)

Jira: [OSD-4766](https://issues.redhat.com/browse/OSD-4766)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ create-account: check-aws-account-id-env
 # Delete account
 .PHONY: delete-account
 delete-account:
-	# Create Account
+	# Delete Account
 	test/integration/api/delete_account.sh
 	# Delete Secrets
 	test/integration/api/delete_account_secrets.sh
--- a/pkg/apis/aws/v1alpha1/account_types.go
+++ b/pkg/apis/aws/v1alpha1/account_types.go
@@ -99,6 +99,9 @@ const (
 	AccountPendingVerification AccountConditionType = "PendingVerification"
 	// AccountReused is set when account is reused
 	AccountReused AccountConditionType = "Reused"
+	// AccountInitializingRegions indicates we've kicked off the process of creating and terminating
+	// instances in all supported regions
+	AccountInitializingRegions = "InitializingRegions"
 )
 
 // +genclient
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -43,7 +43,26 @@ const (
 	awsCredsSecretAccessKey = "aws_secret_access_key"
 	awsAMI                  = "ami-000db10762d0c4c05"
 	awsInstanceType         = "t2.micro"
-	createPendTime          = utils.WaitTime * time.Minute
+	// createPendTime is the maximum time we allow an Account to sit in Creating state before we
+	// time out and set it to Failed.
+	createPendTime = utils.WaitTime * time.Minute
+	// initPendTime is the maximum time we allow between Account creation and the end of region
+	// initialization.
+	// NOTE(efried): We use quite a long timeout here. Within reason, we would rather go too
+	// long than too short.
+	// - We want to give region init every opportunity to finish. We ignore the Account anyway,
+	//   regardless of whether it's failed or initializing; but if we set it to Failed too soon,
+	//   and the goroutine actually finishes, it will (*should*) bounce off of the
+	//   resourceVersion.
+	// - The async region init takes a theoretical maximum of WaitTime * 2 minutes plus a handful
+	//   of AWS API calls. See asyncRegionInit.
+	// - Ideally it would be nice to start timing from right before we kicked off the region
+	//   init, but for simplicity we just use the Account's creation time. So we add the
+	//   maximum time it could take in the Creating state, which is WaitTime minutes (see
+	//   accountCreatingTooLong).
+	// - We add an extra WaitTime for other miscellaneous steps that are otherwise not
+	//   accounted for, hopefully with plenty of buffer.
+	initPendTime = utils.WaitTime * time.Minute * time.Duration(4)
 
 	// AccountPending indicates an account is pending
 	AccountPending = "Pending"
@@ -51,6 +70,9 @@ const (
 	AccountCreating = "Creating"
 	// AccountFailed indicates account creation has failed
 	AccountFailed = "Failed"
+	// AccountInitializingRegions indicates we've kicked off the process of creating and terminating
+	// instances in all supported regions
+	AccountInitializingRegions = "InitializingRegions"
 	// AccountReady indicates account creation is ready
 	AccountReady = "Ready"
 	// AccountPendingVerification indicates verification (of AWS limits and Enterprise Support) is pending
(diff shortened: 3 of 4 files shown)


pkg/controller/account/ec2.go
```diff
@@ -19,7 +19,9 @@ import (
 
 // InitializeSupportedRegions concurrently calls InitializeRegion to create instances in all supported regions
 // This should ensure we don't see any AWS API "PendingVerification" errors when launching instances
-func (r *ReconcileAccount) InitializeSupportedRegions(reqLogger logr.Logger, account *awsv1alpha1.Account, regions map[string]map[string]string, creds *sts.AssumeRoleOutput) error {
+// NOTE: This function does not have any returns. In particular, error conditions from the
+// goroutines are logged, but do not result in a failure up the stack.
+func (r *ReconcileAccount) InitializeSupportedRegions(reqLogger logr.Logger, account *awsv1alpha1.Account, regions map[string]map[string]string, creds *sts.AssumeRoleOutput) {
```
2uasimojo, line 24: Note to reviewers: Taking the return out of the prototype here is a logical no-op, since as written this could only ever return `nil`. I did it to be explicit, and so future me isn't confused by the redundant error check in the caller.

2uasimojo:
/hold

In my sleep last night I thought of a thing that should get further attention before we try to land this:

The code path up to the point where we kick off the new goroutine can include status updates, which will trigger our Watch and cause this Account to be queued for reconciliation, and that Reconcile is able to run as soon as we return, while the goroutine is still running. So we need to make sure that the whole rest of the Reconcile ignores the Account during that time. Rather than trying to suss out all the possible states and conditions as they exist today (there's a lot of 'em) I think it would be better to:
- Introduce a new state, like AccountInitializingRegions, that we set just before spawning the goroutine.
- Add a check early in the Reconcile for that state that simply bails.
- But also take into account the goroutine failing/dying and include a check for `accountInitializingTooLong` similar to the existing one for `accountCreatingTooLong`.

iamkirkbater:
Agreed with the new state.  The big thing that I noticed when doing that I want to just call out here is that updating anything on a CR will trigger a reconcile.  So if we are setting the status to InitializingRegions or whatever, we should be returning right after that update, and having another fork in the main process to kick off the goroutine.

2uasimojo:
> So if we are setting the status to InitializingRegions or whatever, we should be returning right after that update, and having another fork in the main process to kick off the goroutine.

So yeah, while I agree in principle, I'm totally not going to do that for this PR. There are two dozen other places we're updating statuses etc. but continuing on without requeueing. It would take a total redo to fix that. Meanwhile, to make the logic work for this case, I would need a *separate* state that tells the controller we're *ready* to start the init, because part of this PR is making the controller _ignore_ the Account if it's in the InitializingRegions state.

iamkirkbater:
Yeah, I guess that makes sense in this case.  I must have missed the part where you originally said that you were going to ignore the account if it's in the InitializingRegions state.  :+1: to me then!

2uasimojo:
Still needs rebase, which is going to be tricky. Working...

Makefile
```diff
@@ -49,7 +49,7 @@ create-account: check-aws-account-id-env
 # Delete account
 .PHONY: delete-account
 delete-account:
-	# Create Account
+	# Delete Account
```
2uasimojo, line 52: Unrelated typo fix because it was annoying me.

pkg/apis/aws/v1alpha1/account_types.go
```diff
@@ -99,6 +99,9 @@ const (
 	AccountPendingVerification AccountConditionType = "PendingVerification"
 	// AccountReused is set when account is reused
 	AccountReused AccountConditionType = "Reused"
+	// AccountInitializingRegions indicates we've kicked off the process of creating and terminating
+	// instances in all supported regions
+	AccountInitializingRegions = "InitializingRegions"
```
2uasimojo, line 104: Still miffed that I have to duplicate this in two places.

pkg/controller/account/account_controller.go
```diff
@@ -145,7 +167,18 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 
 	// Log accounts that have failed and don't attempt to reconcile them
 	if accountIsFailed(currentAcctInstance) {
-		reqLogger.Info(fmt.Sprintf("Account %s is failed ignoring", currentAcctInstance.Name))
+		reqLogger.Info(fmt.Sprintf("Account %s is failed. Ignoring.", currentAcctInstance.Name))
```
2uasimojo, line 170: Unrelated message cleanup because it was annoying me.

pkg/controller/account/account_controller.go
```diff
@@ -383,35 +416,53 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
 
-		// Initialize all supported regions by creating and terminating an instance in each
-		err = r.InitializeSupportedRegions(reqLogger, currentAcctInstance, awsv1alpha1.CoveredRegions, creds)
-		if err != nil {
-			errMsg := fmt.Sprintf("Unable to build or destroy ec2 instances: %s", err)
-			r.setStateFailed(reqLogger, currentAcctInstance, errMsg)
+		// We're about to kick off region init in a goroutine. This status makes subsequent
+		// Reconciles ignore the Account (unless it stays in this state for too long).
+		utils.SetAccountStatus(currentAcctInstance, "Initializing Regions", awsv1alpha1.AccountInitializingRegions, AccountInitializingRegions)
+		if err := r.statusUpdate(reqLogger, currentAcctInstance); err != nil {
```
2uasimojo, line 422: @clcollins Would you please vet that I've done these things according to the new pattern? The PR originated before yours landed.

2uasimojo:
/hold cancel

This is tested (including the ignores and the new timeout path) and works like a charm. Ready for review/merge.

pkg/controller/account/account_controller.go
clcollins: This looks fine here!

PR #455, opened 2020-08-24
OSD-4944 - Fix deploy-local make target

rogbas:
Make sure the FORCE_DEV_MODE variable is passed to operator-sdk

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -294,9 +294,8 @@ predeploy-aws-account-operator:
 predeploy: predeploy-aws-account-operator deploy-aws-account-operator-credentials
 
 .PHONY: deploy-local
-deploy-local: FORCE_DEV_MODE?=local
 deploy-local:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -294,9 +294,8 @@ predeploy-aws-account-operator:
 predeploy: predeploy-aws-account-operator deploy-aws-account-operator-credentials
 
 .PHONY: deploy-local
-deploy-local: FORCE_DEV_MODE?=local
 deploy-local:
-	@operator-sdk up local --namespace=$(OPERATOR_NAMESPACE)
+	@FORCE_DEV_MODE=local operator-sdk up local --namespace=$(OPERATOR_NAMESPACE)
```
2uasimojo, line 298: Okay.

The way this was before, it *should* have worked the same if you *didn't* set the variable, but should have allowed you to override it in your env. I can't think of a decent use case for the latter, so...

/lgtm

PR #459, opened 2020-08-27
Detect and recover stale region initialization

2uasimojo:
Commit 801903c1 (PR #454) made region initialization happen in the
background. To prevent subsequent reconciles of the same Account from
thrashing, it introduced a new InitializingRegions state which was
ignored by the controller; unless a really long time elapsed, whereupon
we assumed the goroutine died and we set the account to Failed.

This caused a regression: Previously, Accounts stayed in Creating while
regions were being initialized. If the operator died, it would come back
up and churn through to the region initialization code path again. With
the new InitializingRegions state, that wasn't happening.

With this commit, we restore that behavior by detecting whether the
InitializingRegions condition was set earlier than the start time of the
operator. If so, we assume it's dead, and we set the Account's state
back to Creating to give the controller another crack at it.

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -171,14 +171,45 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 		return reconcile.Result{}, nil
 	}
 
-	// Ignore accounts for which we're (asynchronously) initializing regions -- unless they've been
-	// in that state for too long.
+	// Detect accounts for which we kicked off asynchronous region initialization
 	if accountIsInitializingRegions(currentAcctInstance) {
+		// Detect whether we set the InitializingRegions condition in *this* invocation of the
+		// operator or a previous one.
+		if regionInitStale(currentAcctInstance) {
+			// This means the region init goroutine(s) for this account were still running when an
+			// earlier invocation of the operator died. We want to recover those, so set them back
+			// to Creating, which should cause us to hit the region init code path again.
+			msg := "Recovering from stale region initialization."
+			// We're no longer InitializingRegions
+			utils.SetAccountCondition(
+				currentAcctInstance.Status.Conditions,
+				awsv1alpha1.AccountInitializingRegions,
+				// Switch the Condition off
+				corev1.ConditionFalse,
+				AccountInitializingRegions,
+				msg,
+				// Make sure the existing condition is updated
+				utils.UpdateConditionAlways,
+				currentAcctInstance.Spec.BYOC)
+			// TODO(efried): This doesn't change the lastTransitionTime, which it really should.
+			// In fact, since the Creating condition is guaranteed to already be present, this
+			// is currently not doing anything more than
+			//    currentAcctInstance.Status.State = AccountCreating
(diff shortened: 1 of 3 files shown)


pkg/controller/account/account_controller.go
```diff
@@ -171,14 +171,30 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
+		// operator or a previous one.
+		if regionInitStale(currentAcctInstance) {
+			// This means the region init goroutine(s) for this account were still running when an
+			// earlier invocation of the operator died. We want to recover those, so set them back
+			// to Creating, which should cause us to hit the region init code path again.
+			msg := "Recovering from stale region initialization."
+			utils.SetAccountStatus(currentAcctInstance, msg, v1alpha1.AccountCreating, AccountCreating)
+			// The status update will trigger another Reconcile, but be explicit. The requests get
+			// collapsed anyway.
+			return reconcile.Result{Requeue: true}, nil
```
2uasimojo, line 186: It'll help if I actually update the status.

pkg/controller/account/account_controller.go
```diff
@@ -171,14 +171,30 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 		return reconcile.Result{}, nil
 	}
 
-	// Ignore accounts for which we're (asynchronously) initializing regions -- unless they've been
-	// in that state for too long.
+	// Detect accounts for which we kicked off asynchronous region initialization
```
2uasimojo, line 174: We need to do this before `accountCreatingTooLong`. Otherwise we miss recovering accounts that were creating+initializing for more than 25 minutes but less than 100.

pkg/controller/account/account_controller.go
```diff
@@ -171,14 +171,30 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
 	if accountIsInitializingRegions(currentAcctInstance) {
+		// Detect whether we set the InitializingRegions condition in *this* invocation of the
+		// operator or a previous one.
+		if regionInitStale(currentAcctInstance) {
+			// This means the region init goroutine(s) for this account were still running when an
+			// earlier invocation of the operator died. We want to recover those, so set them back
+			// to Creating, which should cause us to hit the region init code path again.
+			msg := "Recovering from stale region initialization."
+			utils.SetAccountStatus(currentAcctInstance, msg, v1alpha1.AccountCreating, AccountCreating)
```
2uasimojo, line 183: We should reset the timestamp on this. Otherwise if we've been creating+initializing for more than 25 minutes but less than 100, we'll time out on `accountCreatingTooLong` and fail the account.

2uasimojo: Yeah, turns out `accountCreatingTooLong` uses the CR's creation time, not the `Creating` Condition's `lastTransitionTime`.

pkg/controller/account/account_controller.go
2uasimojo: Done.

2uasimojo:
Tested and working. Ready for review.

/hold cancel

rogbas:
I did a local test by interrupting the operator and restarting it while an account was in `InitializingRegions`, the process was picked up fine when the operator was brought back up.
/lgtm

PR #467, opened 2020-09-09
Deploy configmap in make

ArielLima:
card -> https://issues.redhat.com/browse/OSD-5095

Add step to the predeploy Makefile target, making sure the a-a-o configuration configMap (aws-account-operator-configmap) gets created.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -291,7 +291,7 @@ predeploy-aws-account-operator:
 	@oc apply -f hack/files/aws_v1alpha1_zero_size_accountpool.yaml
 
 .PHONY: predeploy
-predeploy: predeploy-aws-account-operator deploy-aws-account-operator-credentials
+predeploy: predeploy-aws-account-operator deploy-aws-account-operator-credentials create-ou-map
 
 .PHONY: deploy-local
 deploy-local:
@@ -320,8 +320,8 @@ endif
 # Test apply ou map cr
 .PHONY: create-ou-map
 create-ou-map:
-	# Create OU map
-	@oc process -p ROOT=${OSD_STAGING_1_OU_ROOT_ID} -p BASE=${OSD_STAGING_1_OU_BASE_ID} -f hack/templates/aws_v1alpha1_configmap.tmpl | oc apply -f -
+# Create OU map
+	@oc process -p ROOT=${OSD_STAGING_1_OU_ROOT_ID} -p BASE=${OSD_STAGING_1_OU_BASE_ID} -p ACCOUNTLIMIT="0" -f hack/templates/aws_v1alpha1_configmap.tmpl | oc apply -f -
 
 # Test delete ou map cr
 .PHONY: delete-ou-map
--- a/hack/templates/aws_v1alpha1_configmap.tmpl
+++ b/hack/templates/aws_v1alpha1_configmap.tmpl
@@ -3,6 +3,7 @@ kind: Template
 parameters:
 - name: ROOT 
 - name: BASE
+- name: ACCOUNTLIMIT
 metadata:
   name: test-aws-ou-mapping-configmap-template
 objects:
@@ -14,3 +15,4 @@ objects:
   data:
     root: ${ROOT}
     base: ${BASE}
+    account-limit: ${ACCOUNTLIMIT}


Makefile
```diff
@@ -281,8 +281,14 @@ deploy-aws-account-operator-credentials: check-aws-credentials
 # Create the aws-account-operator-credentials secret
 	@oc process -p OPERATOR_ACCESS_KEY_ID=${ID} -p OPERATOR_SECRET_ACCESS_KEY=${KEY} -p OPERATOR_NAMESPACE=aws-account-operator -f hack/templates/aws_v1alpha1_aws_account_operator_credentials.tmpl | oc apply -f -
 
+#Deploy the operator configmap
+.PHONY: deploy-operator-configmap
+deploy-operator-configmap:
+# Deploy configmap with stage base and root OUs and account-limit to 1
+	@@oc process -p ROOT=r-0wd6 -p BASE=ou-0wd6-tmsbvahq -p ACCOUNTLIMIT="1" -f hack/templates/aws_v1alpha1_configmap.tmpl | oc apply -f -
```
2uasimojo, line 288: This looks very similar to the existing `create-ou-map` target.

If the idea is to hardcode values for `ROOT` and `BASE`, perhaps this can simply set those values and run that target. However, I kinda suspect that's not the point. So I'll ask another way: is there a reason not to simply reuse that target?

hack/templates/aws_v1alpha1_configmap.tmpl
```diff
@@ -3,6 +3,7 @@ kind: Template
 parameters:
 - name: ROOT 
 - name: BASE
+- name: ACCOUNTLIMIT
```
2uasimojo, line 6: On a similar note, won't this will break the existing `create-ou-map` target, which doesn't set `ACCOUNTLIMIT`?

If we leave this out, is the effect that the configmap gets the "nil" value of zero for the account limit? And does that behave improperly for test environments?

2uasimojo (commented):
TL;DR: I wonder if we could just chain the `create-ou-map` target as a dep on `predeploy-aws-account-operator` and be done.

hack/templates/aws_v1alpha1_configmap.tmpl
fahlmant: I'd say change `create-ou-map` to include this and make the target required for `predeploy`.

2uasimojo (commented):
Yeah, this is nice, thanks @ArielLima

/lgtm

Makefile
```diff
@@ -320,8 +320,8 @@ endif
 # Test apply ou map cr
 .PHONY: create-ou-map
 create-ou-map:
-	# Create OU map
-	@oc process -p ROOT=${OSD_STAGING_1_OU_ROOT_ID} -p BASE=${OSD_STAGING_1_OU_BASE_ID} -f hack/templates/aws_v1alpha1_configmap.tmpl | oc apply -f -
+# Create OU map
+	@oc process -p ROOT=${OSD_STAGING_1_OU_ROOT_ID} -p BASE=${OSD_STAGING_1_OU_BASE_ID} -p ACCOUNTLIMIT="1" -f hack/templates/aws_v1alpha1_configmap.tmpl | oc apply -f -
```
fahlmant, line 324: Why is this 1?

ArielLima: I don't know. @rogbas could you clarify? https://issues.redhat.com/browse/OSD-5095

drpaneas: AFAIK this number _used to be_ important in the past, but with the current implementation it doesn't matter if it's `0` or `1` anymore. That has been told, I would prefer to go with `0` since it's the typical initialisation value in programming.

fahlmant:
@rogbas Thoughts on 0 vs 1 for the cm? Otherwise, generally looks good

2uasimojo:
For reference, I commented thusly in the [card](https://issues.redhat.com/browse/OSD-5095):

> I've just been in this neighborhood for #473 and I observed that what the account-limit is being compared against is the number of accounts AWS thinks you have. Under normal circumstances, we're testing with our child accounts, but the total seems to be based on the parent. For me, today, this was 5782 and growing. So initializing this to zero or one or a thousand will make no difference.
>
> Are there cases where we'll be testing with virgin accounts? If so, zero probably makes more sense, to align with the zero-size accountpool. That is, out of the box, prevent you from doing anything, and you have to update it explicitly for the case you're interested in.

fahlmant:
We may be getting a new payer level account soon, so yes we should move down to 0 imo. Also for anyone else out in the void using this operator, 1 could create an account unexpectedly. So +1 to defaulting to 0.

I see 1 as a req in the ticket though, just curious if that's a hard req and if so why

Makefile
rogbas: We should set it to 0. The 1 value was a reminiscence from the previous solution

rogbas:
@ArielLima can you please set account-limit to `0`? Other than that, it LGTM

drpaneas (approved):
/lgtm

fahlmant:
LGTM, mind rebasing?

fahlmant (approved):

PR #469, opened 2020-09-11
Add role for ccs access

iamkirkbater:
Fulfills OSD-4893.  Prerequisite to https://github.com/openshift/osd-utils-cli/pull/68

This PR is a pre-requisite to being able to assume the BYOCAdminAccess role in CCS accounts.  This PR adds a second principal to the BYOCAdminAccess role.

For SREs to access that role they first assume the RH-SRE-CCS-Access role in the payer account that the account was created in and then using the STS credentials for that role they should be able to further access the BYOCAdminAccess role in the customer's account.

To test:

* `make create-ccs`
* `oc get accountclaim --all-namespaces` and note the hash at the end of the account CR name
* `unset AWS_SESSION_TOKEN AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY` to ensure that your env vars are clear
* `aws sts assume-role --role-arn arn:aws:iam::[staging2 account number]:role/RH-SRE-CCS-Access --role-session-name role-chain-test`
* `aws sts assume-role --role-arn arn:aws:iam::[your staging2 account number]:role/BYOCAdminAccess-[hash from step2]`
* `aws sts get-caller-identity` and confirm that you are now accessing your CCS account

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -9,8 +9,12 @@ import (
 	"runtime"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/iam"
 	routev1 "github.com/openshift/api/route/v1"
 	"github.com/openshift/aws-account-operator/pkg/apis"
+	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
+	"github.com/openshift/aws-account-operator/pkg/awsclient"
 	"github.com/openshift/aws-account-operator/pkg/controller"
 	"github.com/openshift/aws-account-operator/pkg/controller/utils"
 	"github.com/openshift/aws-account-operator/pkg/localmetrics"
@@ -19,6 +23,8 @@ import (
 	"github.com/operator-framework/operator-sdk/pkg/leader"
 	"github.com/operator-framework/operator-sdk/pkg/log/zap"
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
@@ -158,12 +164,17 @@ func main() {
 	// work
 	stopCh := signals.SetupSignalHandler()
 
-	// Initialize the TotalAccountWatcher
-	accountWatcherClient, err := client.New(cfg, client.Options{})
+	// Define an awsClient for any processes that need to run during operator startup or independent routines to use
+	awsClient, err := client.New(cfg, client.Options{})
 	if err != nil {
 		log.Error(err, "")
 	}
-	totalaccountwatcher.Initialize(accountWatcherClient, totalWatcherInterval)
+
+	// Initialize our ConfigMap with default values if necessary.
+	initOperatorConfigMapVars(awsClient)
+
+	// Initialize the TotalAccountWatcher
+	totalaccountwatcher.Initialize(awsClient, totalWatcherInterval)
 	go totalaccountwatcher.TotalAccountWatcher.Start(log, stopCh)
 
 	log.Info("Starting the Cmd.")
@@ -172,5 +183,39 @@ func main() {
 	if err := mgr.Start(stopCh); err != nil {
 		log.Error(err, "Manager exited non-zero")
 		os.Exit(1)
+	}
+}
+
+func initOperatorConfigMapVars(kubeClient client.Client) {
+	builder := &awsclient.Builder{}
+	awsClient, err := builder.GetClient("", kubeClient, awsclient.NewAwsClientInput{
+		SecretName: utils.AwsSecretName,
+		NameSpace:  awsv1alpha1.AccountCrNamespace,
+		AwsRegion:  "us-east-1",
+	})
+
+	// Check if config map exists.
+	cm := &corev1.ConfigMap{}
+	err = kubeClient.Get(context.TODO(), types.NamespacedName{Namespace: awsv1alpha1.AccountCrNamespace, Name: awsv1alpha1.DefaultConfigMap}, cm)
+	if err != nil {
+		log.Error(err, "There was an error getting the default configmap.")
+		return
+	}
+
+	// Get the SRE Admin Access role for CCS Accounts and populate the role name into the configmap
+	role, err := awsClient.GetRole(&iam.GetRoleInput{
+		RoleName: aws.String(awsv1alpha1.SREAccessRoleName),
+	})
+	if err != nil {
+		log.Error(err, "There was an error getting the SRE CCS Access Role")
+		return
+	}
+	cm.Data["CCS-Access-Arn"] = *role.Role.Arn
+
+	// Apply the changes to the ConfigMap
+	err = kubeClient.Update(context.TODO(), cm)
+	if err != nil {
+		log.Error(err, "There was an error updating the configmap.")
+		return
 	}
 }
--- a/pkg/apis/aws/v1alpha1/account_types.go
+++ b/pkg/apis/aws/v1alpha1/account_types.go
@@ -28,8 +28,10 @@ const (
 	accountStatusPendingVerification AccountStateStatus = "PendingVerification"
 	// AccountCrNamespace namespace where AWS accounts will be created
 	AccountCrNamespace = "aws-account-operator"
-	// IAM Role name for IAM user creating resources in account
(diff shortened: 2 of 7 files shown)


cmd/manager/main.go
```diff
@@ -158,6 +165,12 @@ func main() {
 	// work
 	stopCh := signals.SetupSignalHandler()
 
+	accountInitClient, err := client.New(cfg, client.Options{})
```
fahlmant, line 168: Could this use the same client as the account watcher? Not sure if it matters but just for less memory usgae

pkg/controller/account/byoc.go
```diff
@@ -85,12 +88,21 @@ func (r *ReconcileAccount) initializeNewCCSAccount(reqLogger logr.Logger, accoun
...
 	accountID := account.Labels[fmt.Sprintf("%s", awsv1alpha1.IAMUserIDLabel)]
 
+	// Get SRE Access ARN from configmap
+	cm := &corev1.ConfigMap{}
+	cmErr := r.Client.Get(context.TODO(), types.NamespacedName{Namespace: awsv1alpha1.AccountCrNamespace, Name: awsv1alpha1.DefaultConfigMap}, cm)
+	if cmErr != nil {
+		reqLogger.Error(cmErr, "There was an error getting the ConfigMap to get the SRE Access Role")
+		return "", reconcile.Result{}, cmErr
+	}
+	SREAccessARN := cm.Data["CCS-Access-Arn"]
```
fahlmant, line 98: Thoughts on adding validation here? Maybe someone changed or removed from the CM after opreator start up

pkg/controller/account/byoc.go
```diff
@@ -190,7 +202,7 @@ func CreateRole(reqLogger logr.Logger, byocRole string, principalARN string, byo
 			Effect: "Allow",
 			Action: []string{"sts:AssumeRole"},
 			Principal: &awsv1alpha1.Principal{
-				AWS: principalARN,
+				AWS: []string{principalARN, SREAccessARN},
```
fahlmant, line 205: Any reason to pass this all the way through instead of use the constant directly here?

cmd/manager/main.go
iamkirkbater: Yeah, I was wondering what the best play would be here.  I have no problem doing that, but I'd want to rename it to be a generic client that we pass into any other startup process there.

pkg/controller/account/byoc.go
iamkirkbater: Good idea, I'll do that.

pkg/controller/account/byoc.go
iamkirkbater: Mainly for when we add tests to this. If we try to pull the constant directly we'll have a hard time trying to build the tests if we need to change that constant for the test.

fahlmant: The tests will have it in the configmap too right?

pkg/controller/account/byoc.go
iamkirkbater: Updated.  If you want to test this specifically, 

* `operator-sdk up local --namespace aws-account-operator
* in another terminal: `oc edit configmap -n aws-account-operator aws-account-operator-configmap` and remove the CCS-Access-Arn line.
* `make test-ccs` should now fail with the new error.

cmd/manager/main.go
iamkirkbater: Created a generic client.

pkg/controller/account/byoc.go
iamkirkbater: They could, but that's an extra mock you have to do.  This way instead of having to mock out a configmap and add it to the mock-api for each test you'd just pass in either one.

I mean, I think I understand the concern, there are a lot of parameters to these functions to begin with, and I'm adding a new one.

What if instead of passing in two separate strings for the principal we change it so that we pass in a single `[]string` for principals.  Then we keep it to one single parameter for the principals but also have the ease of just flipping a variable in a test?

fahlmant: Oh, I like that idea.

pkg/apis/aws/v1alpha1/account_types.go
```diff
@@ -28,8 +28,10 @@ const (
 	accountStatusPendingVerification AccountStateStatus = "PendingVerification"
 	// AccountCrNamespace namespace where AWS accounts will be created
 	AccountCrNamespace = "aws-account-operator"
-	// IAM Role name for IAM user creating resources in account
+	// AccountOperatorIAMRole name for IAM user creating resources in account
 	AccountOperatorIAMRole = "OrganizationAccountAccessRole"
+	// SREAccessRoleName for CCS Account Access
+	SREAccessRoleName = "RH-SRE-CCS-Access"
```
iamkirkbater, line 34: Adding this here so that I can use it in `osdctl` as well.

pkg/apis/aws/v1alpha1/account_types.go
```diff
@@ -28,8 +28,10 @@ const (
 	accountStatusPendingVerification AccountStateStatus = "PendingVerification"
 	// AccountCrNamespace namespace where AWS accounts will be created
 	AccountCrNamespace = "aws-account-operator"
-	// IAM Role name for IAM user creating resources in account
+	// AccountOperatorIAMRole name for IAM user creating resources in account
 	AccountOperatorIAMRole = "OrganizationAccountAccessRole"
```
iamkirkbater, line 32: Fixes doc lint warning

pkg/controller/account/byoc.go
```diff
@@ -85,12 +88,26 @@ func (r *ReconcileAccount) initializeNewCCSAccount(reqLogger logr.Logger, accoun
...
+	cm := &corev1.ConfigMap{}
+	cmErr := r.Client.Get(context.TODO(), types.NamespacedName{Namespace: awsv1alpha1.AccountCrNamespace, Name: awsv1alpha1.DefaultConfigMap}, cm)
+	if cmErr != nil {
+		reqLogger.Error(cmErr, "There was an error getting the ConfigMap to get the SRE Access Role")
+		return "", reconcile.Result{}, cmErr
+	}
+
+	SREAccessARN := cm.Data["CCS-Access-Arn"]
+	if SREAccessARN == "" {
+		reqLogger.Error(fmt.Errorf("CCS-Access-Arn Key missing from ConfigMap %s/%s", awsv1alpha1.AccountCrNamespace, awsv1alpha1.DefaultConfigMap), "The key is missing from the ConfigMap")
```
clcollins, line 101: We have an ["awsv1alpha1.ErrInvalidConfigMap](https://github.com/openshift/aws-account-operator/blob/bbda1c91242b7d0cd78456e3cda55201dc9188f0/pkg/apis/aws/v1alpha1/shared_types.go#L97-L98)" type, and I would argue missing an expected/required value makes the configMap invalid.  Would it be worth using here?

```suggestion
		reqLogger.Error(awsv1alpha1.InvalidConfigMap, "configmap key missing", "keyName", "CCS-Access-Arn")
```
The `"keyName", "CCS-Access-Arn"` will append a field to the logline so that it's indexable/searchable in a log aggregator.

The ReqLogger already [includes the NameSpace as one of these extra appended fields](https://github.com/openshift/aws-account-operator/blob/bbda1c91242b7d0cd78456e3cda55201dc9188f0/pkg/controller/account/account_controller.go#L136), so that can be omitted.

Including the configMap name is fine, or not.  I don't have any opinion on that.

cmd/manager/main.go
```diff
@@ -174,3 +185,38 @@ func main() {
...
+		SecretName: utils.AwsSecretName,
+		NameSpace:  awsv1alpha1.AccountCrNamespace,
+		AwsRegion:  "us-east-1",
+	})
+
+	// Check if config map exists.
+	cm := &corev1.ConfigMap{}
+	err = kubeClient.Get(context.TODO(), types.NamespacedName{Namespace: awsv1alpha1.AccountCrNamespace, Name: awsv1alpha1.DefaultConfigMap}, cm)
+	if err != nil {
+		// TODO: Should we just create the configmap here if it doesn't exist?
```
clcollins, line 201: My two cents is no.  I would argue the configMap data should be predictable and configuration data should be managed by a human.  If we create one, what is the point of the configMap in general?  We either store the config data in code, or we dont.

clcollins (commented):
Other than the logging note, this LGTM!

rogbas:
@iamkirkbater the PR LGTM, but there seems to be a build error, could you please check?

cmd/manager/main.go
iamkirkbater: TODO Removed :)

Thanks @clcollins

PR #460, opened 2020-08-31
Upgrades aws-sdk-go to 1.34.13

clcollins:
This commit upgrades the aws-sdk-go vendor package to 1.34.13, adding, among other things, support for serviceQuota clients and quotaIncreaseRequests.

This passes the Go and Makefile tests, but could use many additional eyes to test, please.

REF: [https://issues.redhat.com/browse/OSD-4821](https://issues.redhat.com/browse/OSD-4821)

Signed-off-by: Christopher Collins <collins.christopher@gmail.com>

PR #470, opened 2020-09-11
Validates vCPU quota for region on account initialization

clcollins:
Adds vCPU quota validation for new accounts
    
Some AWS accounts have been created below the standard vCPU quota request limit, causing failures when installing OpenShift clusters.

This change:
    
 * Retrieves the desired configuration (vCPU quota included) from the operator configmap ("vcpu.quota")
 * Checks the desired quota against the existing quota for each region
 * If the existing quota is too low, checks to see if a request to increase quota has already been submitted
 * If a request has not yet been submitted, requests an increase in quota

**Note: A new value will need to be added to the configMap for the operator - `vcpu.quota` - with the desired quota value.** This change will have no effect if it is not added, defaulting to 0 and believing that no quota increase is required.

This should also be included in the configMap created by #467, though it is only necessary for testing locally.
    
REF: [https://issues.redhat.com/browse/OSD-4821](https://issues.redhat.com/browse/OSD-4821)
    

clcollins:
/hold

Requires/Incorporates #460

pkg/controller/account/account_controller.go
```diff
@@ -706,6 +709,56 @@ func matchSubstring(roleID, role string) (bool, error) {
...
+}
+
+func accountIsCreating(currentAcctInstance *awsv1alpha1.Account) bool {
+	return currentAcctInstance.Status.State == AccountCreating
+}
+
+func accountIsInitializingRegions(currentAcctInstance *awsv1alpha1.Account) bool {
+	return currentAcctInstance.Status.State == AccountInitializingRegions
+}
+
```
2uasimojo, line 761: Did you start this before #439 merged?

pkg/controller/account/ec2.go
```diff
@@ -58,14 +81,59 @@ func (r *ReconcileAccount) InitializeRegion(reqLogger logr.Logger, account *awsv
...
+		// If the caseID is set, a quota increase was requested, either just now or previously. Log it.
+		// Can't update account conditions from within the asyncRegionInit goroutine, because
+		// the account is being updated elsewhere and will conflict.
+		if caseID != "" {
+			reqLogger.Info("quota increase request submitted successfully", "region", region, "caseID", caseID)
+		}
+	}
+
+	// Unset err to be explicit for use below
+	err = nil
```
2uasimojo, line 135: This is weird. Wouldn't L137 overwrite it no matter what?

2uasimojo (commented):
I didn't look over this carefully yet, but I did notice what looks like an incomplete rebase.

pkg/controller/account/account_controller.go
```diff
@@ -706,6 +709,56 @@ func matchSubstring(roleID, role string) (bool, error) {
...
+// getConfigDataByKey gets the desired key from the default configmap or returns an empty string and error
+func getConfigDataByKey(kubeClient client.Client, key string) (string, error) {
+	configMap := &corev1.ConfigMap{}
+	err := kubeClient.Get(
+		context.TODO(),
+		types.NamespacedName{Namespace: awsv1alpha1.AccountCrNamespace,
+			Name: awsv1alpha1.DefaultConfigMap}, configMap)
+	if err != nil {
+		return "", err
+	}
```
fahlmant, line 721: @iamkirkbater has a GET similar to this in his PR and the total account watcher has one too. Thoughts on consolidating it in a function?

pkg/controller/account/ec2.go
clcollins: Just trying to be clear, but yes, you are right.  Removed.

pkg/controller/account/account_controller.go
clcollins: Good idea!  See https://github.com/openshift/aws-account-operator/pull/470/commits/afe82221a3668cf96250187c42ee4fba31f1124f for refactor into a function that will be more universally usable.

Should follow up for implementation with the code @iamkirkbater was working on, and the totalaccountwatcher.

pkg/controller/account/ec2.go
```diff
@@ -58,14 +81,56 @@ func (r *ReconcileAccount) InitializeRegion(reqLogger logr.Logger, account *awsv
...
+		if err != nil {
+			reqLogger.Error(err, "failed retriving current vCPU quota from AWS")
+		}
+	}
+
+	if quotaIncreaseRequired {
+		reqLogger.Info("vCPU quota increase required", "region", region)
+		caseID, err = checkQuotaRequestHistory(awsClient, vCPUQuota)
+		if err != nil {
+			reqLogger.Error(err, "failed retriving quota change history")
```
fahlmant, line 107: Is there any error here that we want to check and possibly return on?

pkg/controller/account/ec2.go
```diff
@@ -258,3 +323,266 @@ func ListEC2InstanceStatus(reqLogger logr.Logger, client awsclient.Client) (*ec2
...
+	v, ok := configMap.Data["quota.vcpu"]
+	if !ok {
+		err = awsv1alpha1.ErrInvalidConfigMap
+	}
+	if err != nil {
+		reqLogger.Error(err, "failed getting desired vCPU quota from configmap data")
+		return vCPUQuota, err
+	}
+
+	vCPUQuota, err = strconv.ParseFloat(v, 64)
```
fahlmant, line 342: Why parse float and not parse int?

fahlmant (commented):
Overall looks good to me. I really like the way you did the retries. Managed to handle both AWS propagation AND rate limits. Only one or two questions. This is dependent on the AWS SDK bump, right?

pkg/controller/account/ec2.go
clcollins: The vCPUQuota request below requires a float64 type, so I parsed it as such here to avoid parsing from string to int and then to float later.

fahlmant: ah got it. Also, cmon AWS, why do you need a float64 for a whole number???

pkg/controller/account/ec2.go
clcollins: I've tried to handle all the error that could reasonably be handled within that function using the retry and conditional exceptions.

With regard to returning, I think it's more appropriate to log a quota request failure, but allow the initialization to continue, as not having the default quota is not necessarily a killer, and I was concerned about bombing out the entire goroutine.

I suppose with Eric's recent changes to add the initializing state, that might not be critical.  I'm happy to take suggestions on whether we should do so or not.

fahlmant: I think its fine as it if you don't see any obvious error to return on. Just want to ensure that all paths were considered since the most that ever happens with this specific err is printing it.

clcollins:
@fahlmant: Rebased!

pkg/controller/account/ec2.go
2uasimojo: small brain dump (apply that adjective however you like...)

It doesn't make sense for a thing that's going to be run in a goroutine to return an error (or anything at all). So that just needs to be removed from the prototype and flows. That's something I left behind when I did #454 because prior to that the error conditions were being ignored anyway.

Errors in the goroutine can only be dealt with via logging and CR status. If we deem an error condition to be fatal, set the CR to `Failed`. (Optimization: cancel any other sub-goroutines.)

pkg/awsclient/client.go
```diff
@@ -383,13 +408,14 @@ func newClient(controllerName, awsAccessID, awsAccessSecret, token, region strin
...
-		s3Client:      s3.New(s),
-		route53client: route53.New(s),
+		ec2Client:           ec2.New(s),
+		iamClient:           iam.New(s),
+		orgClient:           organizations.New(s),
+		route53client:       route53.New(s),
+		s3Client:            s3.New(s),
+		stsClient:           sts.New(s),
+		supportClient:       support.New(s),
+		serviceQuotasClient: servicequotas.New(s, aws.NewConfig().WithRegion("us-west-2")),
```
jharrington22, line 418: Why are you specifying a specific region here?

rogbas:
@rogbas to test and comment

pkg/awsclient/client.go
clcollins: That was an oversight on my part.  I was testing whether or not the quotas were region-based and forgot to remove that.

pkg/controller/account/ec2.go
```diff
@@ -257,3 +322,277 @@ func ListEC2InstanceStatus(reqLogger logr.Logger, client awsclient.Client) (*ec2
...
+	if err != nil {
+		reqLogger.Error(err, "failed converting vCPU quota from configmap string to float64")
+		return vCPUQuota, err
+	}
+
+	return vCPUQuota, nil
+}
+
+// getVCPUQUota returns the current set vCPU quota for the region
+func getVCPUQuota(client awsclient.Client, desiredQuota float64) (bool, error) {
```
iamkirkbater, line 351: If we want to leave the name here as `getVCPUQuota` I think we should be returning an int (or whatever the value that's returned is.  Otherwise I think we should call this something like `validateQuotaIsWithinLimits` or something along those lines.

pkg/controller/utils/utils.go
```diff
@@ -196,3 +200,13 @@ func AccountCRHasIAMUserIDLabel(accountCR *awsv1alpha1.Account) bool {
 
 	return false
 }
+
+// GetOperatorConfigMap retrieves the default configMap data for the AWS Account Operator from Kubernetes
+func GetOperatorConfigMap(kubeClient client.Client) (*corev1.ConfigMap, error) {
```
iamkirkbater, line 205: Nice!

iamkirkbater (commented):
Just a few comments.  Overall looks great, though.

pkg/controller/account/ec2.go
clcollins: Good call.  This is an old name that needs to be updated.  Will do this tomorrow.

pkg/controller/account/account_controller.go
```diff
@@ -37,7 +37,19 @@ import (
 var log = logf.Log.WithName("controller_account")
 
 const (
-	awsInstanceType = "t2.micro"
+	// AwsLimit tracks the hard limit to the number of accounts; exported for use in cmd/manager/main.go
```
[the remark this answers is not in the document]
clcollins: I don't know where this came from.  Ugly rebase, I guess.  Will fix tomorrow.

pkg/controller/account/account_controller.go
```diff
@@ -748,3 +756,13 @@ func matchSubstring(roleID, role string) (bool, error) {
 	matched, err := regexp.MatchString(roleID, role)
 	return matched, err
 }
+
+func regionInitStale(currentAcctInstance *awsv1alpha1.Account) bool {
```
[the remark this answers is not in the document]
clcollins: Ditto. Will fix tomorrow.

pkg/controller/account/ec2.go
```diff
@@ -257,3 +322,277 @@ func ListEC2InstanceStatus(reqLogger logr.Logger, client awsclient.Client) (*ec2
...
+
+// getDesiredVCPUValue retrieves the desired quota information from the operator configmap and converts it to a float64
+func (r *ReconcileAccount) getDesiredVCPUValue(reqLogger logr.Logger) (float64, error) {
+	var err error
+	var vCPUQuota float64
+
+	configMap, err := controllerutils.GetOperatorConfigMap(r.Client)
+	v, ok := configMap.Data["quota.vcpu"]
+	if !ok {
+		err = awsv1alpha1.ErrInvalidConfigMap
```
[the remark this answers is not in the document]
clcollins: :+1:

pkg/controller/utils/utils.go
clcollins: :+1:

pkg/controller/account/ec2.go
clcollins: Renamed to `vCPUQuotaNeedsIncrease`, as it will return "True" if it does need an increase, to better match the behavior of the function.

rogbas:
Overall this looks good, we can merge next Monday Oct 5th.

/hold

clcollins:
Rebased off master (Go Modules support)

PR #471, opened 2020-09-14
Clarifies environment variable setup instructions and fixes some linter issues for README.md

clcollins:
Signed-off-by: Christopher Collins <collins.christopher@gmail.com>

PR #472, opened 2020-09-14
Adds various fixes for Go linter standards

clcollins:
This PR includes a number of fixes to conform to Go linter standards from golangci-linter.  Mostly this includes adding or fixing comments for exported functions, constants and variables.  In a few cases, like the `totalaccountwatcher`, exported-but-unused-outside-the-package functions were changed to be unexported.

Also adds default SREP boilerplate .golangci.yml file provided by @2uasimojo.

Commits have not been squashed so they may be addressed as is, if desired.

Note: `pkg/controller/testutils/testutils.go` has a few linter notices that I am unsure how to fix.  Otherwise, with #471, this handles all the linter issues I could find.

### DIFF
--- a/pkg/apis/aws/v1alpha1/account_types.go
+++ b/pkg/apis/aws/v1alpha1/account_types.go
@@ -18,17 +18,9 @@ const (
 	AccountStatusRequested AccountStateStatus = "Requested"
 	// AccountStatusClaimed const for Claimed status
 	AccountStatusClaimed AccountStateStatus = "Claimed"
-	// AccountStatusTransfering const for Transfering status
-	accountStatusTransfering AccountStateStatus = "Transfering"
-	// AccountStatusTransfered const for Transfering status
-	accountStatusTransfered AccountStateStatus = "Transfered"
-	// AccountStatusDeleting const for Deleting status
-	accountStatusDeleting AccountStateStatus = "Deleting"
-	// AccountStatusPendingVerification const for Pending Verification status
-	accountStatusPendingVerification AccountStateStatus = "PendingVerification"
 	// AccountCrNamespace namespace where AWS accounts will be created
 	AccountCrNamespace = "aws-account-operator"
-	// AccountOperatorIAMRole name for IAM user creating resources in account
+	// AccountOperatorIAMRole is the name for IAM user creating resources in account
 	AccountOperatorIAMRole = "OrganizationAccountAccessRole"
 	// SREAccessRoleName for CCS Account Access
 	SREAccessRoleName = "RH-SRE-CCS-Access"
@@ -216,7 +208,7 @@ func (a *Account) HasAwsv1alpha1Finalizer() bool {
 //IsOlderThan takes a parameter of a time and returns true if the creation timestamp is longer than
 //the passed in time.
 func (a *Account) IsOlderThan(maxDuration time.Duration) bool {
-	return time.Now().Sub(a.GetCreationTimestamp().Time) > maxDuration
+	return time.Since(a.GetCreationTimestamp().Time) > maxDuration
 }
 
 //IsBYOCPendingDeletionWithFinalizer returns true if account is a BYOC Account,
@@ -256,7 +248,7 @@ func (a *Account) IsInitializingRegions() bool {
 	return a.Status.State == AccountInitializingRegions
 }
 
-// FindCondition finds in the condition that has the
+// GetCondition finds the condition that has the
 // specified condition type in the given list. If none exists, then returns nil.
 func (a *Account) GetCondition(conditionType AccountConditionType) *AccountCondition {
 	for i, condition := range a.Status.Conditions {
--- a/pkg/apis/aws/v1alpha1/awsfederatedrole_types.go
+++ b/pkg/apis/aws/v1alpha1/awsfederatedrole_types.go
@@ -68,6 +68,7 @@ type AWSFederatedRoleStatus struct {
 	Conditions []AWSFederatedRoleCondition `json:"conditions"`
 }
 
+// AWSFederatedRoleCondition is a Kubernetes condition type for tracking AWS Federated Role status changes
 type AWSFederatedRoleCondition struct {
 	// Type is the type of the condition.
 	Type AWSFederatedRoleConditionType `json:"type"`
--- a/pkg/awsclient/iam.go
+++ b/pkg/awsclient/iam.go
@@ -12,6 +12,7 @@ import (
 	"github.com/openshift/aws-account-operator/pkg/controller/utils"
 )
 
+// ListIAMUserTags returns a list of the tags assigned to an IAM user in AWS
 func ListIAMUserTags(reqLogger logr.Logger, client Client, userName string) (*iam.ListUserTagsOutput, error) {
 	input := &iam.ListUserTagsInput{
 		UserName: aws.String(userName),
@@ -39,6 +40,7 @@ func ListIAMUserTags(reqLogger logr.Logger, client Client, userName string) (*ia
 	return result, nil
 }
 
+// ListIAMUsers returns an *iam.User list of users from the current account
 func ListIAMUsers(reqLogger logr.Logger, client Client) ([]*iam.User, error) {
 	input := &iam.ListUsersInput{}
 	// List of IAM users to return
@@ -76,7 +78,7 @@ func ListIAMUsers(reqLogger logr.Logger, client Client) ([]*iam.User, error) {
 
 }
 
-// checkIAMUserExists checks if a given IAM user exists within an account
+// CheckIAMUserExists checks if a given IAM user exists within an account
 // Takes a logger, an AWS client for the target account, and a target IAM username
 func CheckIAMUserExists(reqLogger logr.Logger, client Client, userName string) (bool, *iam.GetUserOutput, error) {
 	// Retry when getting IAM user information
@@ -174,6 +176,7 @@ func CreateIAMUser(reqLogger logr.Logger, client Client, account *awsv1alpha1.Ac
 	return createUserOutput, err
 }
 
+// ListIAMRoles returns an *iam.Role list of roles in the AWS account
 func ListIAMRoles(reqLogger logr.Logger, client Client) ([]*iam.Role, error) {
 
 	// List of IAM roles to return
@@ -186,9 +189,7 @@ func ListIAMRoles(reqLogger logr.Logger, client Client) ([]*iam.Role, error) {
 			return nil, err
 		}
 
-		for _, role := range output.Roles {
-			iamRoleList = append(iamRoleList, role)
-		}
+		iamRoleList = append(iamRoleList, output.Roles...)
 
 		if *output.IsTruncated {
 			marker = output.Marker
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -36,12 +36,7 @@ import (
 var log = logf.Log.WithName("controller_account")
 
 const (
-	// AwsLimit tracks the hard limit to the number of accounts; exported for use in cmd/manager/main.go
-	awsCredsUserName        = "aws_user_name"
-	awsCredsSecretIDKey     = "aws_access_key_id"
-	awsCredsSecretAccessKey = "aws_secret_access_key"
-	awsAMI                  = "ami-000db10762d0c4c05"
-	awsInstanceType         = "t2.micro"
+	awsInstanceType = "t2.micro"
 	// createPendTime is the maximum time we allow an Account to sit in Creating state before we
 	// time out and set it to Failed.
 	createPendTime = utils.WaitTime * time.Minute
@@ -225,7 +220,10 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 		ccsRoleID, result, initErr = r.initializeNewCCSAccount(reqLogger, currentAcctInstance, awsSetupClient, adminAccessArn)
 		if initErr != nil {
 			// TODO: If we have recoverable results from above, how do we allow them to requeue if state is failed
-			r.setStateFailed(reqLogger, currentAcctInstance, initErr.Error())
+			stateErr := r.setStateFailed(reqLogger, currentAcctInstance, initErr.Error())
+			if stateErr != nil {
+				reqLogger.Error(stateErr, "failed setting account state", "desiredState", "Failed")
+			}
 			reqLogger.Error(initErr, "failed initializing new CCS account")
 			return result, initErr
 		}
@@ -265,6 +263,10 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 					// After creating the support case requeue the request. To avoid flooding and being blacklisted by AWS when
 					// starting the operator with a large AccountPool, add a randomInterval (between 0 and 30 secs) to the regular wait time
 					randomInterval, err := strconv.Atoi(currentAcctInstance.Spec.AwsAccountID)
+					if err != nil {
+						reqLogger.Error(err, "failed converting AwsAccountID string to int")
+						return reconcile.Result{}, err
+					}
 					randomInterval %= 30
 
 					// This will requeue verification for between 30 and 60 (30+30) seconds, depending on the account
@@ -315,7 +317,10 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 		// see if in creating for longer then default wait time
 		if currentAcctInstance.IsCreating() && currentAcctInstance.IsOlderThan(createPendTime) {
 			errMsg := fmt.Sprintf("Creation pending for longer then %d minutes", utils.WaitTime)
-			r.setStateFailed(reqLogger, currentAcctInstance, errMsg)
+			stateErr := r.setStateFailed(reqLogger, currentAcctInstance, errMsg)
+			if stateErr != nil {
+				reqLogger.Error(stateErr, "failed setting account state", "desiredState", "Failed")
+			}
 		}
 
 		if currentAcctInstance.IsUnclaimedAndHasNoState() {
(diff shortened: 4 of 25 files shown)


pkg/controller/utils/utils.go
```diff
@@ -140,6 +143,7 @@ func LogAwsError(logger logr.Logger, errMsg string, customError error, err error
 	}
 }
 
+// Contains returns true a list of strings includes a specific string
```
2uasimojo, line 146:
```suggestion
// Contains returns true if a list of strings includes a specific string
```

2uasimojo (commented):
This is great, @clcollins.

Whose linting standards were you following here? VSCode's defaults? That might be worth calling out in the PR message.

Note that integrating with boilerplate will [bring in](https://github.com/openshift/boilerplate/blob/a54a406c306b250529929dbdc8db633ab1aa9136/boilerplate/openshift/golang-osd-operator/standard.mk#L82-L85) a hard requirement to conform to golangci-lint, so it may be worth heading in that direction sooner rather than later.

clcollins:
@2uasimojo Thanks!  I was using the default in vscode.  I just added `golangci-lint` and will begin working on those.

/hold wip

2uasimojo:
> @2uasimojo Thanks! I was using the default in vscode. I just added `golangci-lint` and will begin working on those.

Cool. In case it wasn't clear from the other link, here's the [config](https://github.com/openshift/boilerplate/blob/a54a406c306b250529929dbdc8db633ab1aa9136/boilerplate/openshift/golang-osd-operator/golangci.yml) as it currently stands. (It uses the defaults for everything that matters to you.) Note also the [specific version](https://github.com/openshift/boilerplate/blob/a54a406c306b250529929dbdc8db633ab1aa9136/boilerplate/_lib/ensure.sh#L4) which may affect how those defaults behave.

We reserve the right to change that config and/or version, of course, but this ought to get you really close.

Let me know if you need help addressing any of the issues.

pkg/totalaccountwatcher/totalaccountwatcher_test.go
```diff
@@ -47,7 +47,7 @@ func TestAccountWatcherCreation(t *testing.T) {
 		// after mocks is defined
 		defer mocks.mockCtrl.Finish()
 
-		totalAccountWatcher := NewTotalAccountWatcher(mocks.fakeKubeClient, mocks.mockAWSClient, 10)
+		totalAccountWatcher := newTotalAccountWatcher(mocks.fakeKubeClient, mocks.mockAWSClient, 10)
```
fahlmant, line 50: Does this still work now that it's not exported? I guess its in the same pkg

fahlmant (commented):
Looks good to me, mind squashing?

pkg/totalaccountwatcher/totalaccountwatcher_test.go
clcollins: Yeah, it appears to work to me, as you said, in the same package.

clcollins:
@fahlmant I am going to keep working on this with @2uasimojo's `golangci-lint` suggestion.  I'm pretty close with the work from this morning.

pkg/apis/aws/v1alpha1/account_types.go
```diff
@@ -19,16 +19,16 @@ const (
...
+	accountStatusTransfering AccountStateStatus = "Transfering" //nolint:varcheck,deadcode,unused // May be used in future
 	// AccountStatusTransfered const for Transfering status
-	accountStatusTransfered AccountStateStatus = "Transfered"
+	accountStatusTransfered AccountStateStatus = "Transfered" //nolint:varcheck,deadcode,unused // May be used in future
 	// AccountStatusDeleting const for Deleting status
-	accountStatusDeleting AccountStateStatus = "Deleting"
+	accountStatusDeleting AccountStateStatus = "Deleting" //nolint:varcheck,deadcode,unused // May be used in future
 	// AccountStatusPendingVerification const for Pending Verification status
-	accountStatusPendingVerification AccountStateStatus = "PendingVerification"
+	accountStatusPendingVerification AccountStateStatus = "PendingVerification" //nolint:varcheck,deadcode,unused // Maybe be used in future
```
2uasimojo, line 28: It's odd to preserve these as private+unused. If we ever were to use them in the future, we would export them to match `AccountStatusRequested` and `AccountStatusClaimed`, right?

IMO it would make sense to replace all of this with just a comment listing the `AccountStateStatus`es that we "might" use in the future.

But if we are going to keep them, we should make the case in the docstrings match.

clcollins:
@fahlmant & @2uasimojo: Finished linting to golangci-linter standards with the default .golangci.yml file provided by Eric.   If you'll give it a look over and a thumbs up, I'll squash all the commits before an LTGM.

pkg/apis/aws/v1alpha1/account_types.go
```diff
@@ -214,7 +214,7 @@ func (a *Account) HasAwsv1alpha1Finalizer() bool {
 //IsOlderThan takes a parameter of a time and returns true if the creation timestamp is longer than
 //the passed in time.
 func (a *Account) IsOlderThan(maxDuration time.Duration) bool {
-	return time.Now().Sub(a.GetCreationTimestamp().Time) > maxDuration
+	return time.Since(a.GetCreationTimestamp().Time) > maxDuration
```
2uasimojo, line 217: /me fixes #473...

pkg/apis/aws/v1alpha1/account_types.go
```diff
@@ -254,7 +254,7 @@ func (a *Account) IsInitializingRegions() bool {
 	return a.Status.State == AccountInitializingRegions
 }
 
-// FindCondition finds in the condition that has the
+// GetCondition finds in the condition that has the
```
2uasimojo, line 257: (latent)
```suggestion
// GetCondition finds the condition that has the
```

pkg/apis/aws/v1alpha1/awsfederatedrole_types.go
```diff
@@ -68,6 +68,7 @@ type AWSFederatedRoleStatus struct {
 	Conditions []AWSFederatedRoleCondition `json:"conditions"`
 }
 
+//AWSFederatedRoleCondition is a Kubernetes condition type for tracking AWS Federated Role status changes
```
2uasimojo, line 71: micronit
```suggestion
// AWSFederatedRoleCondition is a Kubernetes condition type for tracking AWS Federated Role status changes
```

pkg/controller/account/account_controller.go
```diff
@@ -266,6 +264,10 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 					// After creating the support case requeue the request. To avoid flooding and being blacklisted by AWS when
 					// starting the operator with a large AccountPool, add a randomInterval (between 0 and 30 secs) to the regular wait time
 					randomInterval, err := strconv.Atoi(currentAcctInstance.Spec.AwsAccountID)
+					if err != nil {
+						reqLogger.Error(err, "failed converting AwsAccountID string to int")
+						return reconcile.Result{}, err
```
2uasimojo, line 269: NTS: behavior change here (but a good one).

pkg/controller/account/account_controller.go
```diff
@@ -226,7 +221,10 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 		ccsRoleID, result, initErr = r.initializeNewCCSAccount(reqLogger, currentAcctInstance, awsSetupClient, adminAccessArn)
 		if initErr != nil {
 			// TODO: If we have recoverable results from above, how do we allow them to requeue if state is failed
-			r.setStateFailed(reqLogger, currentAcctInstance, initErr.Error())
+			stateErr := r.setStateFailed(reqLogger, currentAcctInstance, initErr.Error())
+			if stateErr != nil {
+				reqLogger.Error(stateErr, "failed setting account state", "desiredState", "Failed")
+			}
```
2uasimojo, line 227: Totally optional:

These changes (throughout) will result in a pretty chatty log, since `setStateFailed` already logs. So perhaps we should just ignore the error (with or without using `//nolint`) and add a comment to explain why. E.g.
```suggestion
			// Ignore error: setStateFailed logs, and we want to return initErr rather than the error from setStateFailed.
			_ = r.setStateFailed(reqLogger, currentAcctInstance, initErr.Error())
```

pkg/controller/account/account_controller.go
```diff
@@ -316,7 +318,10 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 		// see if in creating for longer then default wait time
 		if currentAcctInstance.IsCreating() && currentAcctInstance.IsOlderThan(createPendTime) {
 			errMsg := fmt.Sprintf("Creation pending for longer then %d minutes", utils.WaitTime)
-			r.setStateFailed(reqLogger, currentAcctInstance, errMsg)
+			stateErr := r.setStateFailed(reqLogger, currentAcctInstance, errMsg)
```
2uasimojo, line 321: (latent) This should return with no requeue.

pkg/controller/account/ec2.go
```diff
@@ -31,7 +31,7 @@ func (r *ReconcileAccount) InitializeSupportedRegions(reqLogger logr.Logger, acc
 
 	// Create go routines to initialize regions in parallel
 	for region := range regions {
-		go r.InitializeRegion(reqLogger, account, region, regions[region]["initializationAMI"], ec2Notifications, ec2Errors, creds)
+		go r.InitializeRegion(reqLogger, account, region, regions[region]["initializationAMI"], ec2Notifications, ec2Errors, creds) //nolint:errcheck // Unable to do anything with the returned error
```
2uasimojo, line 34: +1 with sad face

pkg/controller/account/iam.go
```diff
@@ -141,7 +136,7 @@ func deleteAllAccessKeys(client awsclient.Client, iamUser *iam.User) error {
 
 // checkIAMUserExists checks if a given IAM user exists within an account
 // Takes a logger, an AWS client for the target account, and a target IAM username
-func checkIAMUserExists(reqLogger logr.Logger, client awsclient.Client, userName string) (bool, *iam.GetUserOutput, error) {
+func checkIAMUserExists(reqLogger logr.Logger, client awsclient.Client, userName string) (bool, *iam.GetUserOutput, error) { //nolint:deadcode
```
2uasimojo, line 139: Why not whack it? We can always resurrect it from the git history if we need it later.

[Later] Appears to be redundant with `CheckIAMUserExists` in pkg/awsclient/iam.go.

pkg/controller/accountclaim/accountclaim_controller.go
```diff
@@ -140,7 +138,7 @@ func (r *ReconcileAccountClaim) Reconcile(request reconcile.Request) (reconcile.
 					// First we want to see if this was an update race condition where the credentials rotator will update the CR while the finalizer is trying to run.  If that's the case, we want to requeue and retry, before outright failing the account.
 					if errors.IsConflict(err) {
 						reqLogger.Info("Account CR Modified during CR reset.")
-						err = rawError.New(fmt.Sprintf("Account CR Modified during CR reset. %s", err.Error()))
+						err = fmt.Errorf("Account CR Modified during CR reset. %s", err.Error())
```
2uasimojo, line 141: `errors.Wrap` does pretty much the same thing, but includes more useful information.

pkg/controller/awsfederatedrole/awsfederatedrole_controller.go
```diff
@@ -139,12 +139,19 @@ func (r *ReconcileAWSFederatedRole) Reconcile(request reconcile.Request) (reconc
 		NameSpace:  awsv1alpha1.AccountCrNamespace,
 		AwsRegion:  "us-east-1",
 	})
+	if err != nil {
+		return reconcile.Result{}, err
```
2uasimojo, line 143: NTS: behavior change (but a good one)

pkg/controller/awsfederatedrole/awsfederatedrole_controller.go
```diff
@@ -139,12 +139,19 @@ func (r *ReconcileAWSFederatedRole) Reconcile(request reconcile.Request) (reconc
...
+	}
 
 	// Validates Custom IAM Policy
 
 	log.Info("Validating Custom Policies")
 	// Build custom policy in AWS-valid JSON and converts to string
 	jsonPolicy, err := utils.MarshalIAMPolicy(*instance)
+	if err != nil {
+		reqLogger.Error(err, "failed marshalling IAM Policy", "instanceRoleName", instance.Spec.RoleDisplayName)
+		return reconcile.Result{}, err
```
2uasimojo, line 153: ditto... though I'm wondering if this one will get you in a hard spin (if you couldn't marshal it before, what could have changed that you could marshal it the next time?)

2uasimojo (commented):
Probably nothing in here worth blocking over, but a couple of things to think about.

pkg/apis/aws/v1alpha1/account_types.go
clcollins: I struggled with that decision.  I didn't want to leave unused code, but felt like that made for a complete list of states. I'll remove them.  Thanks for the input.

pkg/controller/account/iam.go
clcollins: I wanted to have someone else on the team evaluate whether or not to remove it.  Usually I create a card for removing a whole function just to be safe.

Thanks for the second set of eyes.  Will remove this.

pkg/controller/awsfederatedrole/awsfederatedrole_controller.go
clcollins: @2uasimojo: Agreed.  It looks like the only error that would be returned here is one bubbled up from `json.Marshal(&policyDoc)` in the underlying function.  That would indeed just continue to cycle the reconcile loop over and over.

I'm unsure how to handle this, though. It appears that all the errors in this controller just requeue. We could avoid the loop by not returning the error to the reconcile loop, but then the CR updates are just ignored.  I don't believe there are states for this CR that can be marked as "failed" or what not.

Removing the return doesn't change anything, as this will just fail later at https://github.com/openshift/aws-account-operator/blob/0c9e7b4cb0d304012f2d28b61f68666700c0a8d9/pkg/controller/awsfederatedrole/awsfederatedrole_controller.go#L177, and be reconciled with an error, thus looping again.

pkg/controller/accountclaim/accountclaim_controller.go
clcollins: TIL, thanks!

pkg/controller/awsfederatedrole/awsfederatedrole_controller.go
2uasimojo: Yeah, I wouldn't worry about it for this PR. If you truly wanted to punt, you could remove the `return` and leave a comment similar to :arrow_up:, but meh.

pkg/controller/accountclaim/accountclaim_controller.go
```diff
@@ -136,10 +137,9 @@ func (r *ReconcileAccountClaim) Reconcile(request reconcile.Request) (reconcile.
 					// we will flag the account with the Failed Reuse condition, and with state = Failed
 
 					// First we want to see if this was an update race condition where the credentials rotator will update the CR while the finalizer is trying to run.  If that's the case, we want to requeue and retry, before outright failing the account.
-					if errors.IsConflict(err) {
+					if k8serr.IsConflict(err) {
 						reqLogger.Info("Account CR Modified during CR reset.")
-						err = fmt.Errorf("Account CR Modified during CR reset. %s", err.Error())
-						return reconcile.Result{}, err
+						return reconcile.Result{}, errors.Wrap(err, "account CR modified during reset")
```
2uasimojo, line 142: Looks like you need to update the expected string in the test (or change this one back).

2uasimojo (commented):
lgtm other than the test fail.

pkg/controller/accountclaim/accountclaim_controller.go
clcollins: I saw that, fixed.  Thanks!

2uasimojo (commented):
lgtm. Squash?

clcollins:
Squashed!  Thanks for the good review @2uasimojo, I appreciate it!

2uasimojo:
Clean squash.

/lgtm
/approve

pkg/controller/account/iam.go
```diff
@@ -139,59 +134,6 @@ func deleteAllAccessKeys(client awsclient.Client, iamUser *iam.User) error {
 	return nil
 }
 
-// checkIAMUserExists checks if a given IAM user exists within an account
-// Takes a logger, an AWS client for the target account, and a target IAM username
-func checkIAMUserExists(reqLogger logr.Logger, client awsclient.Client, userName string) (bool, *iam.GetUserOutput, error) {
```
fahlmant, line 144: Was this removed because its not called anywhere?

2uasimojo: See https://github.com/openshift/aws-account-operator/pull/472#discussion_r488721276

clcollins:
Rebased.

pkg/controller/account/account_controller.go
fahlmant: Updating the status will trigger a requeue anyway.

2uasimojo:
Ran golangci-lint locally on the rebase. It looks to have brought in some code that has a new linter error.

```
cmd/manager/main.go:189:13: ineffectual assignment to `err` (ineffassign)
	awsClient, err := builder.GetClient("", kubeClient, awsclient.NewAwsClientInput{
	           ^
```

2uasimojo:
To prevent drift, you could (here or in a followon) add a target that runs golangci-lint and put it in the way of prow. (Once aao is fully boilerplated, it'll do that for you.)

pkg/controller/account/account_controller.go
2uasimojo: Right, but
1. My main point was that this should return, not carry on. I think as it stands today we get lucky and the subsequent conditions prevent any further code from actually running.
2. We've talked about adding predicates to filter out status-update-only reconciles (cc @iamkirkbater). If we do that, it will be relevant whether our actual return statements request an explicit requeue. Of course, we would have to go curate all of them at that point anyway...

clcollins:
Thanks @2uasimojo, I didn't think to check that. Fixed the lint issue, and added a Makefile target for linting, and added that to the test-all target as well, so we should all end up linting our code immediately when running our tests.

I'll follow up with how to add this to prow.  I'm not sure of that process.

pkg/awsclient/iam.go
```diff
@@ -186,9 +189,7 @@ func ListIAMRoles(reqLogger logr.Logger, client Client) ([]*iam.Role, error) {
 			return nil, err
 		}
 
-		for _, role := range output.Roles {
-			iamRoleList = append(iamRoleList, role)
-		}
+		iamRoleList = append(iamRoleList, output.Roles...)
```
rogbas, line 192: That's an interesting replacement

2uasimojo:
Not sure `test-all` is the right place to add that, but it'll work for now.

I'll help with the prow setup once this merges.

/lgtm

pkg/controller/account/iam.go
rogbas: +1, since we moved that function to pkg/awsclient/iam.go

pkg/awsclient/iam.go
clcollins: Yeah!  I learned some things running through all the fixes. It's been great.

PR #473, opened 2020-09-14
Time InitializingRegions by Condition

2uasimojo:
Asynchronous region initialization (#454, #459) timed out 100 minutes after the Account CR was *created*. This was problematic in the following scenario:

- We're out of capacity in the config map's `account-limit`.
- The Account CR is created.
- Time passes (more than 100 minutes).
- We increase the `account-limit`.

The desired behavior is for the Account CR to be picked up and initialized. However, prior to this commit we would time it out as soon as we had transitioned it to the `InitializingRegions` state, because that immediately preceded backgrounding the region initialization, but the state transition would trigger a Reconcile.

With this commit, we:
- Measure the time the CR has been in the `InitializingRegions` state by looking at the `lastTransitionTime` of the corresponding Condition.
- Time that out after 51 minutes, because we no longer need to account for the Creating phase.

Note that there's still a corner case where we can timeout-fail a non-CCS Account if more than 25 minutes pass between when the CR was created and when we flip it from InitializingRegions back to Creating. This is left as a TODO.

Jira: [OSD-5009](https://issues.redhat.com/browse/OSD-5009)

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -45,23 +45,10 @@ const (
 	// createPendTime is the maximum time we allow an Account to sit in Creating state before we
 	// time out and set it to Failed.
 	createPendTime = utils.WaitTime * time.Minute
-	// initPendTime is the maximum time we allow between Account creation and the end of region
-	// initialization.
-	// NOTE(efried): We use quite a long timeout here. Within reason, we would rather go too
-	// long than too short.
-	// - We want to give region init every opportunity to finish. We ignore the Account anyway,
-	//   regardless of whether it's failed or initializing; but if we set it to Failed too soon,
-	//   and the goroutine actually finishes, it will (*should*) bounce off of the
-	//   resourceVersion.
-	// - The async region init takes a theoretical maximum of WaitTime * 2 minutes plus a handful
-	//   of AWS API calls. See asyncRegionInit.
-	// - Ideally it would be nice to start timing from right before we kicked off the region
-	//   init, but for simplicity we just use the Account's creation time. So we add the
-	//   maximum time it could take in the Creating state, which is WaitTime minutes (see
-	//   accountCreatingTooLong).
-	// - We add an extra WaitTime for other miscellaneous steps that are otherwise not
-	//   accounted for, hopefully with plenty of buffer.
-	initPendTime = utils.WaitTime * time.Minute * time.Duration(4)
+	// regionInitTime is the maximum time we allow an account CR to be in the InitializingRegions
+	// state. This is based on async region init taking a theoretical maximum of WaitTime * 2
+	// minutes plus a handful of AWS API calls (see asyncRegionInit).
+	regionInitTime = (utils.WaitTime * time.Duration(2)) + time.Minute
 
 	// AccountPending indicates an account is pending
 	AccountPending = "Pending"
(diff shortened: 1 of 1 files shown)


pkg/controller/account/account_controller.go
```diff
@@ -194,8 +188,8 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 			return reconcile.Result{Requeue: true}, r.statusUpdate(currentAcctInstance)
 		}
 		// The goroutines happened in this invocation. Time out if that has taken too long.
-		if currentAcctInstance.IsOlderThan(initPendTime) {
-			errMsg := fmt.Sprintf("Initializing regions for longer than %d seconds", initPendTime)
+		if time.Now().Sub(irCond.LastTransitionTime.Time) > regionInitTime {
```
clcollins, line 191: One of the golangci-linter rules suggests using `time.Since(...)` rather than `time.Now().Sub(...)`.

```suggestion
		if time.Since(irCond.LastTransitionTime.Time) > regionInitTime {
```

clcollins (commented):
I will echo again that it's suggested that CR Conditions are for human consumption only and should not be used programatically, but I do conceed that I cannot think of a better way to do this.  Labels and annotations are similarly designed for human consumption, so the only think I could think would be to add state, which the Conditions are a part of anyway.

clcollins:
Other than the suggested lint change, this LGTM.

pkg/controller/account/account_controller.go
2uasimojo: Done, thanks.

2uasimojo:
Thanks for the review @clcollins. I did remember you warning about using Conditions in this manner. Did we ever establish a technical _reason_ for that? But yeah, as you say, nothing significantly better springs to mind, and IMO this makes the code pretty clean :)

2uasimojo:
Tested the recovery code path, works. Regression testing the timeout now.

2uasimojo:
The timeout still has a small hole.

As it stands, if you kill the operator while an Account is in InitializingRegions state, and then start it up more than 25 minutes later, we'll correctly put the Account back into Creating -- but then we'll hit the `accountCreatingTooLong` condition, because that's still being calculated against the creation time of the CR.

A way to solve this would be to change `accountCreatingTooLong` to _also_ compare against the Condition's lastTransitionTime. Unclear what other repercussions this could have. Is this worth trying to fix? Here or in another PR?

2uasimojo:
Per @fahlmant ([slack](https://coreos.slack.com/archives/CNJPJSG3Z/p1600198756442200?thread_ts=1600092724.359800&cid=CNJPJSG3Z)):

> I see it as an edge case and not that important. As long as it only affects non-CCS, I'm fine letting it sit for now

I added a TODO. This is ready to go now.

rogbas:
> The timeout still has a small hole.
> 
> As it stands, if you kill the operator while an Account is in InitializingRegions state, and then start it up more than 25 minutes later, we'll correctly put the Account back into Creating -- but then we'll hit the `accountCreatingTooLong` condition, because that's still being calculated against the creation time of the CR.
> 
> A way to solve this would be to change `accountCreatingTooLong` to _also_ compare against the Condition's lastTransitionTime. Unclear what other repercussions this could have. Is this worth trying to fix? Here or in another PR?

Are we positive this hole will impact only non-CCS accounts?

2uasimojo:
> Are we positive this hole will impact only non-CCS accounts?

Until/unless we move [this code](https://github.com/openshift/aws-account-operator/blob/bbda1c91242b7d0cd78456e3cda55201dc9188f0/pkg/controller/account/account_controller.go#L316-L320), yes.

rogbas:
@2uasimojo thanks, changes validated. This timeout is also blocking us from resetting accounts in integration.

/lgtm

PR #474, opened 2020-09-14
Remove osd-operators-registry targets

2uasimojo:
[osd-operators-registry](https://github.com/openshift/osd-operators-registry) is obsolete; and aws-account-operator was never integrated into it anyway. Remove `make` targets that existed solely for its use.


clcollins:
/lgtm

Straightforward.

fahlmant (approved):
/approve

PR #475, opened 2020-09-15
Delay requeue when account limit reached

2uasimojo:
We refuse to create accounts when the count according to AWS exceeds the configmap's account-limit. Previously this condition triggered an immediate requeue, which would hot-loop the account controller, because
- The count according to AWS effectively never decreases.
- The configmap doesn't get updated frequently.

With this commit, we inject a 15-minute delay before requeueing to avoid this hot loop.

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -327,7 +327,8 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 				// before doing anything make sure we are not over the limit if we are just error
 				if !totalaccountwatcher.TotalAccountWatcher.AccountsCanBeCreated() {
 					reqLogger.Error(awsv1alpha1.ErrAwsAccountLimitExceeded, "AWS Account limit reached")
-					return reconcile.Result{}, awsv1alpha1.ErrAwsAccountLimitExceeded
+					// We don't expect the limit to change very frequently, so wait a while before requeueing to avoid hot lopping.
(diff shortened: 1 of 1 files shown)


2uasimojo:
Tested, works.

pkg/controller/account/account_controller.go
```diff
@@ -327,7 +327,8 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 				// before doing anything make sure we are not over the limit if we are just error
 				if !totalaccountwatcher.TotalAccountWatcher.AccountsCanBeCreated() {
 					reqLogger.Error(awsv1alpha1.ErrAwsAccountLimitExceeded, "AWS Account limit reached")
-					return reconcile.Result{}, awsv1alpha1.ErrAwsAccountLimitExceeded
+					// We don't expect the limit to change very frequently, so wait a while before requeueing to avoid hot lopping.
+					return reconcile.Result{Requeue: true, RequeueAfter: time.Duration(15) * time.Minute}, nil
```
fahlmant, line 331: This is a bit long. I'd prefer 5, but not going to block on that

fahlmant (approved):
/lgtm

pkg/controller/account/account_controller.go
2uasimojo: We did talk about this in [slack](https://coreos.slack.com/archives/CNJPJSG3Z/p1600192298435200) a bit yesterday. The 15 minutes is based on the default spin time of the total account watcher, acknowledging that that's only tangentially relevant. Easy to change if it becomes problematic.

PR #476, opened 2020-09-23
Try and assume BYOCAdminAcccess-uid role during AFAA cleanup

fahlmant:
During the AWS Federated Account Access clean up, the OrganizationAccounAccessRole is assumed. If this fails, then currently the CR will be stuck in the finalizer forever. This means any CCS cluster with an AWS Federated Account Access will fail to clean that role. Since there's no guarantee that the Account CR is around, it cannot be relied on to see if a given cluster is CCS or default. To work around this, if assuming the OrganizationAccounAccessRole fails, the cleanup now tries to assume BYOCAdminAccess-uid role. If that works, then the cleanup continues normally. If both fail, then its assumed the cluster is a misconfigured CCS cluster.


iamkirkbater (approved):
/lgtm

PR #477, opened 2020-09-23
Adds conditional retry for some specific errors, eg: InvalidClientTokenID

clcollins:
This adds a conditional, retry with exponential backoff to the CreateAccessKey request, to handle specific errors that may be recoverable with a retry, to address InvalidClientTokenID errors that appear under load in some cases.

It also removes a duplicate setting of account state:

The `RotateIAMAccessKeys` function called `utils.SetAccountStatus` to set the account to failed if creating a new IAM user access key failed, but if that function returns an error to `BuildIAMUser`, `BuildIAMUser` will return an error to the reconcile loop, which also attempts to set the state to failed.

This removes the duplicate from `RotateIAMAccess Keys`, and reformats the error log message that is generated to make use of the searchable/indexable fields.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -34,6 +34,14 @@
   version = "1.0.0"
 
 [[projects]]
+  digest = "1:c67a4f1ad6bb0a6d743a93078f932e2553290b0402efdf655eaf58634dd254be"
+  name = "github.com/avast/retry-go"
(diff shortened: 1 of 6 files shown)


rogbas:
@clcollins it LGTM, were you able to test it locally?

PR #478, opened 2020-09-24
Adds error code checking and retry for AccessDenied

iamkirkbater:
This adds the error check, but if anyone has any idea how to test this I'm all ears.


pkg/controller/account/iam.go
```diff
@@ -162,6 +162,11 @@ func CreateIAMUser(reqLogger logr.Logger, client awsclient.Client, userName stri
 					if attempt == 10 {
 						return &iam.CreateUserOutput{}, err
 					}
+				case "AccessDenied":
+					reqLogger.Info("Attempt to create user is Unauthorized. Trying Again due to AWS Eventual Consistency")
+					if attempt == 5 {
```
jharrington22, line 167: Any reason we're not making this `10` like the case above?

fahlmant (approved):
/lgtm

PR #479, opened 2020-09-25
osd-5105 - Convert totalaccountwacther initialization to async

drpaneas:
WDYT?

### DIFF
--- a/pkg/totalaccountwatcher/totalaccountwatcher.go
+++ b/pkg/totalaccountwatcher/totalaccountwatcher.go
@@ -24,7 +24,7 @@ import (
 var ErrAwsAccountLimitExceeded = errors.New("AccountLimitExceeded")
 
 // TotalAccountWatcher global var for TotalAccountWatcher
-var TotalAccountWatcher *totalAccountWatcher
+var TotalAccountWatcher = &totalAccountWatcher{}
 
 var log = logf.Log.WithName("aws-account-operator")
 
@@ -36,8 +36,8 @@ type totalAccountWatcher struct {
 	accountsCanBeCreated bool
 }
 
-// Initialize creates a global instance of the TotalAccountWatcher
-func Initialize(client client.Client, watchInterval time.Duration) {
+// initialize creates a global instance of the TotalAccountWatcher
+func initialize(client client.Client, watchInterval time.Duration) *totalAccountWatcher {
 	log.Info("Initializing the totalAccountWatcher")
 
 	// NOTE(efried): This is a snowflake use of awsclient.IBuilder. Everyone else puts the
@@ -52,7 +52,7 @@ func Initialize(client client.Client, watchInterval time.Duration) {
 
 	if err != nil {
 		log.Error(err, "Failed to get AwsClient")
-		return
+		return TotalAccountWatcher
 	}
 
 	TotalAccountWatcher = newTotalAccountWatcher(client, awsClient, watchInterval)
@@ -60,6 +60,7 @@ func Initialize(client client.Client, watchInterval time.Duration) {
 	if err != nil {
 		log.Error(err, "failed updating total accounts count")
 	}
+	return TotalAccountWatcher
 }
 
 // newTotalAccountWatcher returns a new instance of the TotalAccountWatcher interface
@@ -79,8 +80,9 @@ func newTotalAccountWatcher(
 
 // TotalAccountWatcher will trigger AwsLimitUpdate every `scanInternal` and only stop if the operator is killed or a
 // message is sent on the stopCh
-func (s *totalAccountWatcher) Start(log logr.Logger, stopCh <-chan struct{}) {
+func (s *totalAccountWatcher) Start(log logr.Logger, stopCh <-chan struct{}, client client.Client, watchInterval time.Duration) {
 	log.Info("Starting the totalAccountWatcher")
+	s = initialize(client, watchInterval)
 	for {
 		select {
 		case <-time.After(s.watchInterval):
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -47,7 +47,7 @@ var (
 	customMetricsPort string = "8080"
 	customMetricsPath string = "/metrics"
 
-	totalWatcherInterval = time.Duration(15) * time.Minute
+	totalWatcherInterval = time.Duration(5) * time.Minute
 )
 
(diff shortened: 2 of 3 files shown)


pkg/totalaccountwatcher/totalaccountwatcher.go
```diff
@@ -79,8 +82,13 @@ func newTotalAccountWatcher(
 
 // TotalAccountWatcher will trigger AwsLimitUpdate every `scanInternal` and only stop if the operator is killed or a
 // message is sent on the stopCh
-func (s *totalAccountWatcher) Start(log logr.Logger, stopCh <-chan struct{}) {
+func (s *totalAccountWatcher) Start(log logr.Logger, stopCh <-chan struct{}, client client.Client, watchInterval time.Duration) {
 	log.Info("Starting the totalAccountWatcher")
+	// This should run only the first time
+	if !TotalAccountWatcherInitialized {
+		Initialize(client, watchInterval)
```
2uasimojo, line 89: Big +1 combining `Initialize` and `Start`. You could take it a step further and make `Initialize` private.

pkg/totalaccountwatcher/totalaccountwatcher.go
```diff
@@ -26,6 +26,9 @@ var ErrAwsAccountLimitExceeded = errors.New("AccountLimitExceeded")
 // TotalAccountWatcher global var for TotalAccountWatcher
 var TotalAccountWatcher *totalAccountWatcher
 
+// TotalAccountWatcherInitialized ensures async initialization triggers only once
+var TotalAccountWatcherInitialized bool = false
```
2uasimojo, line 30: How could it initialize more than once?

2uasimojo:
We've talked about this recently. It's got tradeoffs to be aware of:
Yes, it will make the operator "start" more quickly. But the account controller will fail to reconcile any CRs that require AWS account creation until the TAW is ready. As you know that can take a couple of minutes. 
If we're running at the limit, we pile up such CRs until the limit is increased. And we increase limits by redeploying from app-interface. As things stand, that redeploy will wait to start the controller until the TAW is ready, so we'll be able to service those waiting-for-account CRs right away at that point. But with this change, those reconciles will fail until the TAW is ready.
Without #475, we would have hot looped during that interval; but then serviced those CRs as soon as the TAW was ready (taking automatic backoff into account). But with #475, those CRs will be requeued after 15 minutes.
The effect in this scenario is that CRs in this state will not even start to reconcile until after the operator has been up for 15 minutes.
TL;DR: If you do decide to go ahead with this change, I suggest reducing that delay.

2uasimojo (commented):
IMO we don't need the extra variable. Just L175 and L89.

pkg/totalaccountwatcher/totalaccountwatcher.go
drpaneas: I had a pair programming session yesterday with @rogbas and he mentioned 

```go
go totalaccountwatcher.TotalAccountWatcher.Start(log, stopCh, awsClient, totalWatcherInterval)
```

Starts the method in an async way,  so we need make sure we do not constantly initialising the TAW. So the first time we start the go routine, initialize is running, all fine. But after a period of time (e.g. 15 m), we run it for a second time, we don't need to re-initialize the watcher.

I am puzzled now though, because @2uasimojo observations seems correct. How this can run a second time?

pkg/totalaccountwatcher/totalaccountwatcher.go
drpaneas: yup, good idea ;)

pkg/totalaccountwatcher/totalaccountwatcher.go
2uasimojo: Right; this isn't running in a controller, it's running as part of startup, so it should only be hit once.

rogbas:
@rogbas please review/run it locally and comment

pkg/totalaccountwatcher/totalaccountwatcher.go
rogbas: Yes, nvm, the Start function will be called only once, there's no need for having this check to prevent running Initialize twice.

rogbas:
Considering Eric's comment and other discussions we had in slack, I suggest we update the `totalWatcherInterval` (as well as the RequeuAfter interval from #475) to 5min instead of 15min

rogbas:
@drpaneas I'm constantly getting a segfault around 1min after starting the operator when I run this PR locally, is it working fine for you? It's being triggered by the return of the ListAccount call, which is getting the total account count correctly from AWS btw:

```
{"level":"info","ts":1601672680.0781806,"logger":"aws-account-operator","msg":"Updating total from 0 to 6036"}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1724ca4]

goroutine 312 [running]:
github.com/openshift/aws-account-operator/pkg/totalaccountwatcher.(*totalAccountWatcher).Start(0x0, 0x2144c80, 0xc00040ab40, 0xc000053aa0, 0x2144c20, 0xc0000d6410, 0xd18c2e2800)
	/home/rogbas/go/src/github.com/openshift/aws-account-operator/pkg/totalaccountwatcher/totalaccountwatcher.go:94 +0x84
created by main.main
	/home/rogbas/go/src/github.com/openshift/aws-account-operator/cmd/manager/main.go:175 +0x5e2
Error: failed to run operator locally: (failed to exec []string{"build/_output/bin/aws-account-operator-local"}: exit status 2)

```

drpaneas:
> @drpaneas I did some testing today, until without a conclusion, we can sync on Monday's standup.
> 
> I was suspecting the Go GC was causing it and ran the operator locally with GC turned off (variable GOGC=off), but still got the same nil pointer segfault.

I have tried that as well last week, GOGC=off doesn't make a difference for me as well. Let's sync on Monday ;(

drpaneas:
It doesn't crash anymore. @rogbas please review once more :)

rogbas:
@drpaneas can you please check the failing prow test and rebase?

drpaneas:
> @drpaneas can you please check the failing prow test and rebase?

done

iamkirkbater:
Tested locally with the newest changes.

/lgtm
/approve
/hold cancel

PR #480, opened 2020-09-25
Adds Concurrent Reconciles.

iamkirkbater:
Adds options for concurrent reconciles.  Also fixes some linting errors.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -25,8 +25,6 @@ import (
 	"github.com/operator-framework/operator-sdk/pkg/leader"
 	"github.com/operator-framework/operator-sdk/pkg/log/zap"
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
-	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
@@ -116,6 +114,21 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Define a kubeClient for any processes that need to run during operator startup or independent routines
+	kubeClient, err := client.New(cfg, client.Options{})
+	if err != nil {
+		log.Error(err, "")
+		os.Exit(1)
+	}
+
+	errors := utils.InitControllerMaxReconciles(kubeClient)
+	if len(errors) > 0 {
+		log.Info("There was at least one error initializing controller max reconcile values.")
+		for _, err := range errors {
+			log.Error(err, "")
+		}
+	}
+
 	log.Info("Registering Components.")
 
 	// Setup Scheme for all resources
@@ -169,17 +182,11 @@ func main() {
 	// work
 	stopCh := signals.SetupSignalHandler()
 
-	// Define an awsClient for any processes that need to run during operator startup or independent routines to use
-	awsClient, err := client.New(cfg, client.Options{})
-	if err != nil {
-		log.Error(err, "")
-	}
-
 	// Initialize our ConfigMap with default values if necessary.
-	initOperatorConfigMapVars(awsClient)
+	initOperatorConfigMapVars(kubeClient)
 
 	// Initialize the TotalAccountWatcher
-	totalaccountwatcher.Initialize(awsClient, totalWatcherInterval)
+	totalaccountwatcher.Initialize(kubeClient, totalWatcherInterval)
 	go totalaccountwatcher.TotalAccountWatcher.Start(log, stopCh)
 
 	log.Info("Starting the Cmd.")
@@ -204,9 +211,7 @@ func initOperatorConfigMapVars(kubeClient client.Client) {
 		return
 	}
 
-	// Check if config map exists.
-	cm := &corev1.ConfigMap{}
-	err = kubeClient.Get(context.TODO(), types.NamespacedName{Namespace: awsv1alpha1.AccountCrNamespace, Name: awsv1alpha1.DefaultConfigMap}, cm)
+	cm, err := utils.GetOperatorConfigMap(kubeClient)
 	if err != nil {
 		log.Error(err, "There was an error getting the default configmap.")
 		return
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -23,7 +23,6 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	kubeclientpkg "sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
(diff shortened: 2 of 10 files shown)


clcollins:
/lgtm

Those "fmt.Sprintf(ALREADY A STRING)" lint errors are insidious!  They creep in everywhere.

pkg/controller/account/account_controller.go
```diff
@@ -67,6 +67,8 @@ const (
 	iamUserNameSRE = "osdManagedAdminSRE"
 
 	controllerName = "account"
+
+	maxConcurrentReconciles = 4
```
2uasimojo, line 71: Whoah!

This controller is *not* ready to go concurrent!

2uasimojo (commented):
Unless I missed something major, I'm almost certain it's not this simple.

rogbas:
I'd like to put this PR on hold, we actually don't this requirement currently.

I would consider this kind of change if the outcome of the load tests was indicating we have a bottle-neck on the account claiming process. So far the results from our load tests with 500 regular clusters are not indicating any of that.

On top of that, IMO we have some prep work to do in order to start parallel processing with the operator.

iamkirkbater:
@rogbas Ahh, okay, I just saw this in the TODO for this sprint and knew I had the prior knowledge of how "simple" this would be, so I figured I'd just knock this out.

@2uasimojo I'd like to discuss with you sometime offline your thoughts on why this wouldn't work, as my preliminary tests during the spike had shown that this would work, and that it was this simple (albeit this was before InitRegions went async).  But then again, I could easily be missing something.

With all of that said, though, I've also pulled out the linting error fixes into #481 in order to not block those.

pkg/controller/account/account_controller.go
2uasimojo: Sorry folks, I didn't mean to drop this bomb and run, I was on my way out and wanted to express concern.

More at the top level...

2uasimojo:
> Unless I missed something major, I'm almost certain it's not this simple.

My strong reaction was because (at least) the account controller is not idempotent. If we can prove that we never enter the Reconcile twice for the same CR in parallel (other than during async region init) then I'd be much less concerned.

However, we should consider @rogbas's question: Do we have evidence that we need to do this in the first place?

iamkirkbater:
I agree with the "Let's see if we need this".

As of this moment, this _should_ be the required work to add this.  I removed and force-pushed the linting commit and moved that out (did I say that already? I'm not scrolling back up).

Worst case scenario, we close this as of right now, but we have the diff to show what needs to be done (not that it's really a lot, anyway...)

iamkirkbater:
Still working on this.  Should be able to finish first thing next week after my Primary shift is over.

pkg/controller/awsfederatedrole/awsfederatedrole_controller.go
```diff
@@ -25,7 +25,7 @@ import (
 )
 
 const (
-	controllerName = "awsdfederatedrole"
```
iamkirkbater, line 28: this is what I like to call a pro-gamer move

iamkirkbater:
This builds off of #470 so to view only the changes related to this PR use this link: https://github.com/openshift/aws-account-operator/pull/480/files/b2cbc44ec50860a629e618c38e9c704d034b39d5..a3ce3632242f9fa97de5b904d04baa1e98662cec

rogbas:
@iamkirkbater Could you please rebase and squash?

2uasimojo (commented):
This actually looks like it'll fail if the CM doesn't have an entry for a given controller. Is that what we want?

iamkirkbater:
@rogbas @fahlmant @clcollins @drewandersonnz @drpaneas I would like to hear your thoughts on the above question from @2uasimojo :

> This actually looks like it'll fail if the CM doesn't have an entry for a given controller. Is that what we want?

Related code: https://github.com/openshift/aws-account-operator/pull/480/files#diff-435eecb1d2af40ee96747f88ab71eb2758da989c92f76dcb1f714fc1b300c633R116-R123

fahlmant:
@iamkirkbater I think we should default to 1 if there's no specific config in the CM

cmd/manager/main.go
```diff
@@ -108,6 +106,21 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Define an kubeClient for any processes that need to run during operator startup or independent routines to use
```
2uasimojo, line 109: femtonit (also latent)
```suggestion
	// Define a kubeClient for any processes that need to run during operator startup or independent routines
```

hack/templates/aws_v1alpha1_configmap.tmpl
```diff
@@ -16,3 +16,8 @@ objects:
     root: ${ROOT}
     base: ${BASE}
     account-limit: ${ACCOUNTLIMIT}
+    MaxReconciles.account: "1"
```
2uasimojo, line 19: We're going to need to parameterize these in hack/olm-registry/olm-artifacts-template.yaml if they're to be settable via app-interface.

pkg/controller/account/account_controller.go
```diff
@@ -89,13 +89,18 @@ func newReconciler(mgr manager.Manager) reconcile.Reconciler {
...
 	}
+
 	return utils.NewReconcilerWithMetrics(reconciler, controllerName)
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
 func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	// Create a new controller
-	c, err := controller.New("account-controller", mgr, controller.Options{Reconciler: r})
+	maxConcurrentReconciles, err := utils.GetControllerMaxReconciles(controllerName)
```
2uasimojo, line 99: I wonder if it's worth factoring this repeating pattern out into a helper. Would at least make it easier to tweak the behavior e.g. when someone says "These should be warning, not error" :P

pkg/controller/utils/utils_test.go
```diff
@@ -230,3 +233,69 @@ func TestRemove(t *testing.T) {
...
+			},
+		},
+		{
+			name:        "Tests not valid str->int conversion",
+			expectedErr: fmt.Errorf("strconv.Atoi: parsing \"fourty-two\": invalid syntax"),
+			expectedVal: 0,
+			configMap: &corev1.ConfigMap{
+				ObjectMeta: validObjectMeta,
+				Data: map[string]string{
+					"MaxReconciles.test-controller": "fourty-two",
```
2uasimojo, line 264: femtonit:
```suggestion
			expectedErr: fmt.Errorf("strconv.Atoi: parsing \"forty-two\": invalid syntax"),
			expectedVal: 0,
			configMap: &corev1.ConfigMap{
				ObjectMeta: validObjectMeta,
				Data: map[string]string{
					"MaxReconciles.test-controller": "forty-two",
```

2uasimojo (commented):
I like this.

pkg/controller/utils/controllerwrapper.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+var controllerMaxReconciles map[string]int = map[string]int{}
+
+func NewControllerWithMaxReconciles(log logr.Logger, controllerName string, mgr manager.Manager, r reconcile.Reconciler) (controller.Controller, error) {
+	maxConcurrentReconciles, err := getControllerMaxReconciles(controllerName)
+	if err != nil {
+		log.Error(err, "")
+	}
+	return controller.New("account-controller", mgr, controller.Options{Reconciler: r, MaxConcurrentReconciles: maxConcurrentReconciles})
```
2uasimojo, line 23:
```suggestion
	return controller.New(fmt.Sprintf("%s-controller", controllerName), mgr, controller.Options{Reconciler: r, MaxConcurrentReconciles: maxConcurrentReconciles})
```

2uasimojo (commented):
Getting the controller name right will be important for logging.

Also: planning to do the hack/olm-registry/olm-artifacts-template.yaml parameterization in this PR?

pkg/controller/utils/controllerwrapper.go
iamkirkbater: ohh deerrrppppppp.

Something something mondays.

Thanks for pointing that out :)

2uasimojo (commented):
Nice @iamkirkbater.

/lgtm

Squash?

2uasimojo:
Trivial rebase and squash since previous

/lgtm

rogbas:
@iamkirkbater can you please check the failing prow/coverage test and remove hold?

iamkirkbater:
/retest

Looks like it's that bug that Eric fixed with the coverage job failing due to the missing file.

iamkirkbater:
@rogbas - PR has been rebased and is ready for final review.

rogbas:
@rogbas update respective Jira card and close PR

PR #481, opened 2020-09-25
Fixes linting errors

iamkirkbater:
Pulled these out from another PR that needs more discussion.


2uasimojo (approved):
I would have even inlined these messages rather than keeping the throwaway variables. But +1 to these fixes.

/lgtm

PR #485, opened 2020-10-07
Created more account and accountclaim conditions

ArielLima:
Each existing code path that leads to a failed account has a new, representative condition with a descriptive TYPE, REASON and MESSAGE.

Card -> https://issues.redhat.com/browse/OSD-5396

### DIFF
--- a/pkg/apis/aws/v1alpha1/account_types.go
+++ b/pkg/apis/aws/v1alpha1/account_types.go
@@ -89,12 +89,24 @@ const (
 	AccountReady AccountConditionType = "Ready"
 	// AccountFailed is set when account creation has failed
 	AccountFailed AccountConditionType = "Failed"
+	// AccountCreationFailed is set during AWS account creation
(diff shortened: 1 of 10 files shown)


pkg/apis/aws/v1alpha1/account_types.go
```diff
@@ -89,6 +89,8 @@ const (
 	AccountReady AccountConditionType = "Ready"
 	// AccountFailed is set when account creation has failed
 	AccountFailed AccountConditionType = "Failed"
+	// AccountCreationFailed is set when an error occured during account creations
```
clcollins, line 92: Would you mind making this "...during AWS account creation" to specify that it was a "real" account creation that failed, and not our operator CR object that failed?  Just to be clear what the difference is until the "AccountFailed" condition goes away.

clcollins (changes requested):

rogbas:
@ArielLima there is some lint and build errors, could you please check it? thanks

PR #486, opened 2020-10-08
Updates predeploy step and cleans operator

iamkirkbater:
This PR updates the Makefile to fix the predeploy step so that it works
with your downloaded AWS Profiles, and also adds a clean step to the
make test-all step so that the tests won't be impacted by any dangling
resources.

Satisfies [OSD-5401](https://issues.redhat.com/browse/OSD-5401)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,13 +67,14 @@ create-account: check-aws-account-id-env
 .PHONY: delete-account
 delete-account:
 	# Delete Account
-	test/integration/api/delete_account.sh
+	test/integration/api/delete_account.sh || true
 	# Delete Secrets
 	test/integration/api/delete_account_secrets.sh
 
 # Test Account creation
 .PHONY: test-account-creation
-test-account-creation: create-account test-secrets delete-account
+test-account-creation: delete-account create-account test-secrets
+	test/integration/api/delete_account.sh || true
 
 # Create account claim namespace
 .PHONY: create-accountclaim-namespace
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -175,13 +176,13 @@ create-ccs-2-namespace:
...
 	# Delete CCS namespace
-	@oc process -p NAME=${CCS_NAMESPACE_NAME} -f hack/templates/namespace.tmpl | oc delete -f -
+	@oc process -p NAME=${CCS_NAMESPACE_NAME} -f hack/templates/namespace.tmpl | oc delete -f - || true
 
 # Delete CCS (BYOC) namespace
 .PHONY: delete-ccs-2-namespace
 delete-ccs-2-namespace:
 	# Delete CCS namespace
-	@oc process -p NAME=${CCS_NAMESPACE_NAME_2} -f hack/templates/namespace.tmpl | oc delete -f -
+	@oc process -p NAME=${CCS_NAMESPACE_NAME_2} -f hack/templates/namespace.tmpl | oc delete -f - || true
```
2uasimojo, line 185: Not sure I agree with all the `|| true`s. See below for an alternative.

Makefile
```diff
@@ -345,4 +342,10 @@ test-aws-ou-logic: check-ou-mapping-configmap-env check-ou-mapping-configmap-env
...
 #s Test all
 .PHONY: test-all
-test-all: lint test-account-creation test-ccs test-reuse test-awsfederatedaccountaccess test-awsfederatedrole test-aws-ou-logic 
+test-all: lint clean-operator test-account-creation test-ccs test-reuse test-awsfederatedaccountaccess test-awsfederatedrole test-aws-ou-logic 
+
+.PHONY: clean-operator
+clean-operator: delete-accountclaim-namespace delete-ccs-namespace delete-ccs-2-namespace
+	oc delete accounts --all -n ${NAMESPACE}
+	oc delete awsfederatedaccountaccess --all -n ${NAMESPACE}
+	oc delete awsfederatedrole --all -n ${NAMESPACE}
```
2uasimojo, line 351: Rather than making the targets themselves always succeed, you could make them do so only when run under `clean-operator` via something like:

```suggestion
clean-operator:
	$(MAKE) delete-accountclaim-namespace || true
	$(MAKE) delete-ccs-namespace || true
	$(MAKE) delete-ccs-2-namespace || true
	oc delete accounts --all -n ${NAMESPACE}
	oc delete awsfederatedaccountaccess --all -n ${NAMESPACE}
	oc delete awsfederatedrole --all -n ${NAMESPACE}
```

2uasimojo (commented):
Good change.

Not sure there's truly a context where I care if `make delete-*` fails, but in case there is, suggestion inline.

Makefile
iamkirkbater: Thanks for the suggestion.  Added.

2uasimojo:
Cool, thanks @iamkirkbater.
/lgtm

rogbas:
@rogbas please test locally and approve

iamkirkbater:
Rebased with new changes.  Make tests all passed again.

PR #487, opened 2020-10-12
Adds AWS Error Code to Metrics

iamkirkbater:
AWS's `awserr.Code()` seems to be an ENUM of strings that are fixed and not changeable (i.e. they won't be "ErrNoAccount12345" but it will be "ErrNoAccount" or something, so we should be able to filter down on these error codes.  This should include any Rate Limit errors (although they seem to be different for each service :( ))

Satisfies OSD-5391

### DIFF
--- a/pkg/controller/utils/clientwrapper.go
+++ b/pkg/controller/utils/clientwrapper.go
@@ -77,7 +77,7 @@ func (cmt *ControllerMetricsTripper) RoundTrip(req *http.Request) (*http.Respons
 
 	// Count this call, if it worked (where "worked" includes HTTP errors).
 	if err == nil {
-		localmetrics.Collector.AddAPICall(cmt.Controller, req, resp, time.Since(start).Seconds())
+		localmetrics.Collector.AddAPICall(cmt.Controller, req, resp, time.Since(start).Seconds(), nil)
 	}
 
 	return resp, err
--- a/pkg/localmetrics/localmetrics.go
+++ b/pkg/localmetrics/localmetrics.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"net/http"
 	neturl "net/url"
-	"strconv"
 	"strings"
 
 	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
@@ -142,7 +141,7 @@ func NewMetricsCollector(store cache.Cache) *MetricsCollector {
 			Help:        "Distribution of the number of seconds a Reconcile takes, broken down by controller",
 			ConstLabels: prometheus.Labels{"name": operatorName},
 			Buckets:     []float64{0.001, 0.01, 0.1, 1, 5, 10, 20},
-		}, []string{"controller", "error", "aws_error"}),
+		}, []string{"controller", "error", "error_source"}),
 
 		// apiCallDuration times API requests. Histogram also gives us a _count metric for free.
 		apiCallDuration: prometheus.NewHistogramVec(prometheus.HistogramOpts{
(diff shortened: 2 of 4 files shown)


pkg/localmetrics/localmetrics.go
```diff
@@ -294,7 +295,11 @@ func (c *MetricsCollector) AddAccountReuseCleanupFailure() {
 
 // SetReconcileDuration describes the time it takes for the operator to complete a single reconcile loop
 func (c *MetricsCollector) SetReconcileDuration(controller string, duration float64, err error) {
-	c.reconcileDuration.WithLabelValues(controller, strconv.FormatBool(err != nil)).Observe(duration)
+	awsErrorCode := ""
+	if aerr, ok := err.(awserr.Error); ok {
```
2uasimojo, line 299: Summary of offline discussion:
- This is goodness. Keep it.
- But also see if you can add the dimension to the [`apiCallDuration` metric](https://github.com/openshift/aws-account-operator/pull/487/files#diff-2afe99d72c5464037a873b976d1699b8R155) by passing an `error` arg to [`AddAPICall`](https://github.com/openshift/aws-account-operator/pull/487/files#diff-2afe99d72c5464037a873b976d1699b8R310) based on the `Error` field of the AWS `request` object available to our [callback](https://github.com/openshift/aws-account-operator/blob/48be5995045f04924e28ada9c5f465274d26fafd/pkg/awsclient/client.go#L380)

pkg/localmetrics/localmetrics.go
```diff
@@ -141,7 +142,7 @@ func NewMetricsCollector(store cache.Cache) *MetricsCollector {
 			Help:        "Distribution of the number of seconds a Reconcile takes, broken down by controller",
 			ConstLabels: prometheus.Labels{"name": operatorName},
 			Buckets:     []float64{0.001, 0.01, 0.1, 1, 5, 10, 20},
-		}, []string{"controller", "error"}),
+		}, []string{"controller", "error", "aws_error"}),
```
2uasimojo, line 145: I don't love the idea of making an aws-specific label here. Why not just make it something like `error_code` and use a special value like `{OTHER}` when it's one we don't have special parsing for (which would include AWS errors that didn't cast). And you could also include the a-a-o errors with specific class names using the `%T` we talked about last week.

For that matter, you could do all of the above in the existing `error` dimension. It's not _exactly_ backward compatible with the existing bool, but I don't think that much matters (and it's copacetic wrt data type, since it's a string in prom anyway).

2uasimojo: Per PR review: make two dimensions like `error_source` ("aws", "k8s", etc) and `error_code` (or leave it as `error`) for the specifics.

pkg/localmetrics/localmetrics_test.go
```diff
@@ -94,3 +95,32 @@ func TestPathParse(t *testing.T) {
...
+			name:     "Test No Error gives empty strings",
+			err:      nil,
+			expected: []string{"", ""},
+		},
+		{
+			name:     "Test AWS Error gives aws codes",
+			err:      awserr.New("RateLimit", "This is a message", nil),
+			expected: []string{"RateLimit", "aws"},
+		},
+	}
```
2uasimojo, line 116: Add a test case for `{OTHER}`?

2uasimojo (commented):
Very cool, @iamkirkbater.

Still considering doing similar in `apiCallDuration`?

pkg/localmetrics/localmetrics_test.go
iamkirkbater: Good catch!

2uasimojo:
Nice.
/lgtm
Happy to re-look/re-stamp if you decide to do `apiCallDuration` in this PR.
May also want to consider squashing whenever yer done.

pkg/controller/utils/clientwrapper.go
```diff
@@ -77,7 +77,7 @@ func (cmt *ControllerMetricsTripper) RoundTrip(req *http.Request) (*http.Respons
 
 	// Count this call, if it worked (where "worked" includes HTTP errors).
 	if err == nil {
-		localmetrics.Collector.AddAPICall(cmt.Controller, req, resp, time.Since(start).Seconds())
+		localmetrics.Collector.AddAPICall(cmt.Controller, req, resp, time.Since(start).Seconds(), nil)
```
2uasimojo, line 80: Could add a TODO here to count the errors.

2uasimojo (approved):
Very cool!

Based on live tests we've talked about offline, this is working well, though we would still like to see how rate limit errors appear.

Counting kube errors isn't in scope here, so I think this is done.

/lgtm

iamkirkbater:
Rebased on top of newest changes.  Make tests all passed.

rogbas:
@rogbas please test locally and approve

PR #488, opened 2020-10-13
Fix CRD generation

fahlmant:
There was a minor typo in the AccountClaim spec that was preventing CRD generation. PR fixes the typo and re-generates the CRDs


deploy/crds/aws_v1alpha1_awsfederatedrole_crd.yaml
```diff
@@ -49,13 +49,23 @@ spec:
...
                       effect:
                         type: string
                       principal:
                         properties:
                           AWS:
                             description: aws account id
-                            type: string
+                            items:
+                              type: string
+                            type: array
```
rogbas, line 68: Should this field be an array of strings?

rogbas: Answered by: https://github.com/openshift/aws-account-operator/blob/master/pkg/apis/aws/v1alpha1/awsfederatedrole_types.go

PR #489, opened 2020-10-14
Fixed some makefile problems

iamkirkbater:
There were a few .PHONY targets with dots in front of their name, which made them not work from the command line and get the error "There is nothing to do"

Removed the `create-ou-map` and `delete-ou-map` from the create and delete accountclaim targets in order to allow successive tests on the same operator-local environment now that the configmap is expected to persist.

Removed the double check for ou env vars in the test-aws-ou-logic target.

There is no card, I just noticed these and wanted to fix them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ default: gobuild
 
 # Extend Makefile after here
 
-.PHONY: .check-aws-account-id-env
+.PHONY: check-aws-account-id-env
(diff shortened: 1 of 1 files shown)


fahlmant (approved):
/lgtm

drpaneas (approved):
lgtm

PR #492, opened 2020-10-19
Updates to use KIND as a local dev environment

rogbas:


2uasimojo (commented):
Seems reasonable.

/lgtm

PR #496, opened 2020-10-22
Fix region init timeout

2uasimojo:
We forgot a multiplication by `time.Minute`, so we were ending up with 60s + 50ns instead of 60s + 50min. Fix.


fahlmant (approved):
/lgtm

PR #497, opened 2020-10-23
Adds retry steps to other IAM components

iamkirkbater:
This satisfies the simple solution to [osd-5610](https://issues.redhat.com/browse/osd-5610).

### DIFF
--- a/pkg/controller/account/iam.go
+++ b/pkg/controller/account/iam.go
@@ -117,17 +117,77 @@ func getStsCredentials(reqLogger logr.Logger, client awsclient.Client, iamRoleNa
 	return assumeRoleOutput, nil
 }
 
+func retryIfAwsServiceFailureOrInvalidToken(err error) bool {
+	if aerr, ok := err.(awserr.Error); ok {
+		switch aerr.Code() {
+		// ServiceFailure may be an unspecified server-side error, and is worth retrying
+		case "ServiceFailure":
+			return true
+		// InvalidClientTokenId may be a transient auth issue, retry
+		case "InvalidClientTokenId":
+			return true
+		}
+	}
+	// Otherwise, do not retry
+	return false
+}
+
+func listAccessKeys(client awsclient.Client, iamUser *iam.User) (*iam.ListAccessKeysOutput, error) {
+	var result *iam.ListAccessKeysOutput
+	var err error
+
+	// Default is 1/10 of a second, but any retries we need to make should be delayed a few seconds
+	// This also defaults to an exponential backoff, so we only need to try ~5 times, default is 10
+	retry.DefaultDelay = 3 * time.Second
+	retry.DefaultAttempts = uint(5)
+	err = retry.Do(
+		func() (err error) {
+			result, err = client.ListAccessKeys(&iam.ListAccessKeysInput{UserName: iamUser.UserName})
+			return err
+		},
+
+		// Retry if we receive some specific errors: access denied, rate limit or server-side error
+		retry.RetryIf(retryIfAwsServiceFailureOrInvalidToken),
+	)
+
+	return result, err
+}
+
+func deleteAccessKey(client awsclient.Client, accessKeyID *string, username *string) (*iam.DeleteAccessKeyOutput, error) {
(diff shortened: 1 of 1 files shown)


2uasimojo (commented):
+1

Can we DRY this up? Make a standardized retry wrapper that accepts the func() pointer specific to the call?

iamkirkbater:
DRY'd up the error handler, but not sure if there's a non-hacky way to do the retry.do().

pkg/controller/account/iam.go
```diff
@@ -117,17 +117,79 @@ func getStsCredentials(reqLogger logr.Logger, client awsclient.Client, iamRoleNa
...
+	var result *iam.ListAccessKeysOutput
+	var err error
+
+	// Default is 1/10 of a second, but any retries we need to make should be delayed a few seconds
+	// This also defaults to an exponential backoff, so we only need to try ~5 times, default is 10
+	retry.DefaultDelay = 3 * time.Second
+	retry.DefaultAttempts = uint(5)
+	err = retry.Do(
+		func() (err error) {
+			// Create new access key for user
```
2uasimojo, line 145: copypasta

pkg/controller/account/iam.go
```diff
@@ -117,17 +117,79 @@ func getStsCredentials(reqLogger logr.Logger, client awsclient.Client, iamRoleNa
...
+	var result *iam.DeleteAccessKeyOutput
+	var err error
+
+	// Default is 1/10 of a second, but any retries we need to make should be delayed a few seconds
+	// This also defaults to an exponential backoff, so we only need to try ~5 times, default is 10
+	retry.DefaultDelay = 3 * time.Second
+	retry.DefaultAttempts = uint(5)
+	err = retry.Do(
+		func() (err error) {
+			// Create new access key for user
```
2uasimojo, line 167: copypasta

2uasimojo (approved):
> DRY'd up the error handler, but not sure if there's a non-hacky way to do the retry.do().

Eugh, yeah, getting the real `result` out would be a serious PITA.

This is better, though.

Couple of trivial comment fixups if you like, but this

/lgtm

pkg/controller/account/iam.go
```diff
@@ -117,17 +117,79 @@ func getStsCredentials(reqLogger logr.Logger, client awsclient.Client, iamRoleNa
...
+}
+
+func listAccessKeys(client awsclient.Client, iamUser *iam.User) (*iam.ListAccessKeysOutput, error) {
+	var result *iam.ListAccessKeysOutput
+	var err error
+
+	// Default is 1/10 of a second, but any retries we need to make should be delayed a few seconds
+	// This also defaults to an exponential backoff, so we only need to try ~5 times, default is 10
+	retry.DefaultDelay = 3 * time.Second
+	retry.DefaultAttempts = uint(5)
```
2uasimojo, line 142: Hm, I just noticed: these seem to be global? That's kind of icky. Not a big deal, probably, since we're setting the same wherever we're using them. But does the library provide a way to set them explicitly per use?

pkg/controller/account/iam.go
iamkirkbater: d'oh

pkg/controller/account/iam.go
iamkirkbater: From my quick once-over of the library: no, there's not a way to set them explicitly per use.  I might have missed it though :(

2uasimojo: I didn't dig in too far. It looks like it's certainly possible based on the examples, but it wouldn't be worthwhile if we had to write huge custom handlers like those.

Anyway, like I said, it's not a huge deal since we're setting them the same everywhere -- but to emphasize the fact that they're global, perhaps we should do those assignments (once) at the top level of the file?

rogbas: We can revisit the scope of those variables when we review the general retry approach for AWS

rogbas (approved):
Checked and tested locally, in the next load test we'll give this code a test with a higher volume.

/lgtm
/approve

PR #498, opened 2020-10-27
Add AccessDenied retry.

iamkirkbater:
Retries on AccessDenied within IAM errors due to eventual consistency issues.

Satisfies [OSD-5677](https://issues.redhat.com/browse/OSD-5677)


jharrington22:
@iamkirkbater this condition is only handling delete and list calls to the AWS API. The errors we saw are you able confirm it was one of these two paths? There should be no delete called.

iamkirkbater:
According to what Rogerio posted in Slack from the error message the call that specifically failed was `iam:ListAccessKeys`, so this should fix that error path.

jharrington22:
Great @iamkirkbater

/lgtm

PR #508, opened 2020-11-17
Refactor mocks to avoid circular imports

2uasimojo:
We hadn't regenerated mocks since the last time the awsclient was modified. Doing so as is resulted in a circular import (code in `awsclient` was referring to `mock` which refers to `awsclient`).

This commit moves the mock-related `IBuilder` implementation into the `mock` package to obviate this circular import.

It also syncs the `go:generate` directive and the generated file's name as `zz_generated.mock_client.go`. (The `zz_generated` prefix, in addition to being a nice eyecatcher, will cause the generated file to be ignored for code coverage reporting when that's put in place later.)


pkg/awsclient/mock/mock_client.go
```diff
@@ -1,879 +1,39 @@
-// Code generated by MockGen. DO NOT EDIT.
-// Source: pkg/awsclient/client.go
-
-// Package mock is a generated GoMock package.
 package mock
```
2uasimojo, line 1: Note to reviewers: This is a new file. It just happens to have the same name as the generated file used to. Ignore the "delta".

pkg/awsclient/mock/zz_generated.mock_client.go
```diff
@@ -0,0 +1,919 @@
+// Code generated by MockGen. DO NOT EDIT.
```
2uasimojo, line 1: Note to reviewers: This _should_ have simply showed up as a rename of `mock_client.go`. The contents are identical.

iamkirkbater:
/lgtm

Pulled and ran locally without issue.

Thanks for the extra comments about the file movement/generation @2uasimojo!!

PR #511, opened 2021-02-01
Fix `make coverage`

2uasimojo:
Adds a workaround for https://github.com/golang/go/issues/27333 to make `make coverage` work.


iamkirkbater:
Works on my machine!

/lgtm

PR #516, opened 2021-03-07
Add CloudWatchReadOnlyAccess managed policy to Read only users

maorfr:
related to https://issues.redhat.com/browse/OSD-6671

OCM users use the AWS Infrastructure Access feature to gain access to the cluster's AWS account. They can either choose read-only or network-mgmt roles.

To enable OSD users to properly use the new Log Forwarding feature, it is expected that users can go to the cluster's AWS account and get the logs from there. When trying to do that currently, a user gets a permission error (`not
 authorized to perform: logs:DescribeLogGroups`).

This PR adds the `CloudWatchReadOnlyAccess` managed policy to the Read only access policy.


fahlmant (approved):
/lgtm

PR #518, opened 2021-03-08
Improve failure path messages

dastergon:
Added messages in errors that weren't returning a meaningful message.

PTAL

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -87,14 +87,14 @@ func main() {
 	// are started.
 	// It must also be done exactly once -- see the docstring.
 	if err := utils.InitOperatorStartTime(); err != nil {
-		log.Error(err, "")
+		log.Error(err, "Failed to initialize Operator start time")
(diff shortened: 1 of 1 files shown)


drpaneas:
Thanks :)

Since you are going down the rabbithole of errors, could you possibly add/revisit some relative text error messages for better logging on the various instances found at the same file (that is `cmd/manager/main.go`)? e.g. `log.Error(err, "Coffee not found")`

iamkirkbater:
Agreeing with Panos here.  Can you just take a quick look at the other errors in this file, and then squash the commits down to a single one?  LGTM so far :)  Thanks!

dastergon:
@iamkirkbater unfortunately, I can't invest time to work on this PR right now due to the Cluster Relocation epic, but I will get back to it soon.

dastergon:
@iamkirkbater I've updated my changes. PTAL

iamkirkbater:
Nice work! 

/lgtm

PR #519, opened 2021-03-10
Updates operator to have STS Mode

iamkirkbater:
After some quick experimentation this PR adds "STSMode" to the operator. Fulfills [OSD-6555](https://issues.redhat.com/browse/OSD-6555)

This still depends on some upstream dependencies.

To test:
* Follow the instructions in the updated Install docs to create the jump role and access role.
    * Create the Jump Role and Policy
    * Create the Access Role and attach Admin policy
    * Update the aws-account-operator configmap
    * Add the required environment variable `STS_ROLE_ARN`
* I've found it's just easier to manually verify through the CLI you can use your payer-account user creds to assume role into the jump role, and then assume the access role.
* start the operator and run `make test-sts-accountclaim`

TODO/TO CARD:

- [x] We need to create the Jump roles and accounts (should we have more than one? one for Prod, one for staging, etc?)
- [x] We need to deploy the jump roles into the existing config maps
- [x] Ensure the CRDs get updated (is that done automatically or is that a manual step?)
- [x] Write an integration test for this (for `make test-all`)
- [x] There are some opportunities for refactoring in here

### DIFF
--- a/docs/1.0-Installation.md
+++ b/docs/1.0-Installation.md
@@ -1,5 +1,13 @@
 # 1.0 Installation
 
+### 1.0.1 - Glossary
+
+Terms used in this guide and what they mean for this guide:
+
+* Payer Account - AWS Organizations give you a "unified bill" so to speak under a single account. Sub accounts are created under this "Payer Account", which is the highest-level account in this organization. SREs all should also have an IAM user with credentials on each payer account, which has AssumeRole Permissions as administrator into each of these sub-accounts.
+* Jump Role - Using AWS Assume-Role Chaining, a Jump Role is an intermediary role that sits between a given IAM user and a customer's access role, providing a single point of entry.
+* Access Role - A role inside the cluster's account that we assume to perform tasks.
+
 ## 1.1 - Prerequisites
 
 ### 1.1.1 - IAM User and Secret
@@ -48,7 +56,99 @@ OPERATOR_SECRET_ACCESS_KEY
 
 Alternatively, if you have the credentials stored in your `~/.aws/credentials` file, you can use the script in `scripts/set_operator_credentials.sh` to create the secret from that profile: `./scripts/set_operator_credentials.sh my-aws-profile`
 
-### 1.1.2 Config Map
+### 1.1.2 AWS Accounts
+
+As you're working on the operator you should have two accounts dedicated for your use of this operator, under each of the osd-staging payer accounts.  Work with your manager or functional team lead to figure out what these account numbers are and save them, you'll need them later.
+
+### 1.1.3 STS Roles
+
+To run the operator tests for STS mode you will need to create two roles &mdash; one in each of the AWS accounts you have been assigned.  The name of these roles doesn't matter when you create it, however you will need to know the ARN of both in order to configure the operator to use them.
+
+#### 1.1.3.1 - Jump Role
+
+The Jump Role will be a simulation of the role that we use as a bastion of sorts. For the STS Architecture, we have an IAM user that assumes a specific role in a specific account, and then using that role they can assume role into the cluster's account in order to run operations as necessary.
+
+You will need to create this jump role in the first osd-staging aws account assigned to you. You will need to add your Payer Account IAM user credentials as the the Principal in the Trust Relationship. This user should also be the same user that you use to create the AWS Account Operator credentials.
+
+The Jump role should only require the permissions to assume-roles.  No other permissions should be necessary. You will probably need to create the IAM policy first before the Role.
+
+Example Assume Role Policy:
+```
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Effect": "Allow",
+            "Action": "sts:AssumeRole",
+            "Resource": "*"
+        }
+    ]
+}
+```
+
+Example Trust Relationship for role:
+```
+{
+    "Version": "2012-10-17",
+        "Statement": [
+        {
+            "Effect": "Allow",
+            "Principal": {
+                "AWS": "arn:aws:iam::[payer-account-id]:user/[your-username]"
+            },
+            "Action": "sts:AssumeRole",
+            "Condition": {}
+        }
+        ]
+}
+```
+
+You will need to put the ARN for this jump role into the Operator Config Map created later.
+
+#### 1.1.3.2 - Access Role
+
+The Access Role will be a simulation of a role on a customer's account that gives us access to initalize the regions and initialize a cluster.  On external customer accounts this is assumed to be a very locked-down role with ONLY the necessary permissions needed to run the operator or install the cluster resources. Minimal permissions required for the role are provided below.
+
+Example Trust Relationship:
+```
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Principal": {
+        "AWS": "arn:aws:iam::[jump-role-account-id]:role/[jump-role-name]"
+      },
+      "Action": "sts:AssumeRole",
+      "Condition": {}
+    }
+  ]
+}
+```
+
+Example minimal permissions Policy Document
+```
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": [
+        "ec2:RunInstances",
+        "ec2:TerminateInstances",
+        "ec2:DescribeInstanceStatus",
+        "ec2:CreateTags",
+        "ec2:DeleteTags"
+      ],
+      "Resource": "*"
+    }
+  ]
+}
+```
+
+You will need to put the ARN for this access role into your environment in order for this to be used to create accountclaims within the make tests.
+
+### 1.1.3 Config Map
 
 The operator also needs a ConfigMap that has:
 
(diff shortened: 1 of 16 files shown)


docs/1.0-Installation.md
```diff
@@ -48,21 +56,95 @@ OPERATOR_SECRET_ACCESS_KEY
...
+        }
+        ]
+}
+```
+
+You will need to put the ARN for this jump role into the Operator Config Map created later.
+
+#### 1.1.3.2 - Access Role
+
+The Access Role will be a simulation of a role on a customer's account that gives us access to initalize the regions and initialize a cluster.  On external customer accounts this is assumed to be a very locked-down role with ONLY the necessary permissions needed to run the operator or install the cluster resources. For development purposes, attaching the Administrator IAM Policy to this role will work.
```
billmvt, line 110: > For development purposes, attaching the Administrator IAM Policy to this role will work.
Once we've figured it out, let's update the IAM Policy we use for dev/test to reflect whatever set of streamlined permissions we tell Customers to use when they create Policies and Roles for OSD. Can catch issues earlier in SDLC that way.

iamkirkbater: Absolutely! I just didn't want to hold up development implementations for that aspect, even though that's the next logical step from here.

pkg/controller/account/byoc.go
```diff
@@ -394,6 +410,54 @@ func DeleteRole(reqLogger logr.Logger, byocRole string, byocAWSClient awsclient.
...
+		return nil, nil, cmErr
+	}
+
+	stsAccessARN := cm.Data["STS-Jump-Role"]
+	if stsAccessARN == "" {
+		log.Error(awsv1alpha1.ErrInvalidConfigMap, "configmap key missing", "keyName", "STS-Jump-Role")
+		return nil, nil, cmErr
+	}
+
+	jumpRoleCreds, err := getStsCredentials(log, operatorAWSClient, stsAccessARN, "awsAccountOperator")
```
jharrington22, line 428: @iamkirkbater this is not what the AAO does today but you can save some lines here an use a provided function from  the AWS SDK [here](https://docs.aws.amazon.com/sdk-for-go/api/aws/credentials/stscreds/#NewCredentials)

Hive are actually using it [here](https://github.com/openshift/hive/pull/1306/files#diff-781a1cdc5ca06b860f587e10c6144a65f58ba703efe886db21d61f5b1fb22978R549) if you want an example. This might be a second PR to refactor.

iamkirkbater:
/hold cancel

This should be ready for prime-time whenever it's reviewed.

boilerplate/_lib/container-make
```diff
@@ -14,13 +14,13 @@ CONTAINER_ENGINE=$(command -v podman || command -v docker)
 
 # First set up a detached container with the repo mounted.
 banner "Starting the container"
-container_id=$($CONTAINER_ENGINE run -d -v "$REPO_ROOT":"$REPO_ROOT" $IMAGE_PULL_PATH tail -f /dev/null)
+container_id=$($CONTAINER_ENGINE run -d -v "$REPO_ROOT":"/go/src/github.com/openshift/aws-account-operator" $IMAGE_PULL_PATH tail -f /dev/null)
```
rogbas, line 17: Any reason for moving away from the $REPO_ROOT variable here?

docs/1.0-Installation.md
```diff
@@ -1,5 +1,13 @@
 # 1.0 Installation
 
+### 1.0.1 - Glossary
+
+Terms used in this guide and what they mean for this guide:
+
+* Payer Account - AWS Organizations give you a "unified bill" so to speak under a single account. Sub accounts are created under this "Payer Account", which is the highest-level account in this organization. SREs all should also have an IAM user with credentials on each payer account, which has AssumeRole Permissions as administrator into each of these sub-accounts.
+* Jump Role - Using AWS Assume-Role Chaining, a Jump Role is an intermediary role that sits between a given IAM user and a customer's access role, providing a single point of entry.
+* Access Role - A role inside the cluster's account that we assume to perform tasks.
+
```
rogbas, line 10: +100000 nice touch adding the glossary

hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -59,6 +61,7 @@ objects:
     account-limit: "${ACCOUNT_LIMIT}"
     root: ${ROOT_OU_ID}
     base: ${BASE_OU_ID}
+    STS-Jump-Role: ${STS_JUMP_ROLE}
```
rogbas, line 64: Can we continue with the standard of having ConfigMap keys in lower case?

pkg/controller/account/byoc.go
```diff
@@ -72,12 +73,16 @@ func (r *ReconcileAccount) initializeNewCCSAccount(reqLogger logr.Logger, accoun
 	// before setting the account claim to an error state
 	var clientErr error
 	var client awsclient.Client
-	for i := 0; i < 5; i++ {
-		client, clientErr = r.getCCSClient(account, accountClaim)
-		if clientErr == nil {
-			break
+	if account.Spec.ManualSTSMode {
+		client, _, clientErr = r.getSTSClient(reqLogger, accountClaim, awsSetupClient)
```
rogbas, line 77: Do we need special retry logic here, similar to the non-STSMode flow below?

pkg/controller/account/byoc.go
```diff
@@ -408,8 +472,14 @@ func (r *ReconcileAccount) getCCSClient(currentAcct *awsv1alpha1.Account, accoun
 	return ccsAWSClient, nil
 }
 
-func validateBYOCClaim(accountClaim *awsv1alpha1.AccountClaim) error {
+func validateSTSClaim(accountClaim *awsv1alpha1.AccountClaim) error {
+	if accountClaim.Spec.BYOCAWSAccountID == "" {
```
rogbas, line 476: Should we also validate if `STSRoleARN` is empty?

pkg/controller/account/iam.go
```diff
@@ -71,16 +71,10 @@ func (r *ReconcileAccount) CreateSecret(reqLogger logr.Logger, account *awsv1alp
 
 // getStsCredentials returns STS credentials for the specified account ARN
 // Takes a logger, an awsClient, a role name to assume, and the target AWS account ID
-func getStsCredentials(reqLogger logr.Logger, client awsclient.Client, iamRoleName string, awsAccountID string) (*sts.AssumeRoleOutput, error) {
-	// Use the role session name to uniquely identify a session when the same role
-	// is assumed by different principals or for different reasons.
-	var roleSessionName = "awsAccountOperator"
+func getStsCredentials(reqLogger logr.Logger, client awsclient.Client, roleArn string, roleSessionName string) (*sts.AssumeRoleOutput, error) {
```
crozzy, line 74: Maybe this should be named `getSTSCredentials` to match Go convention and `validateSTSClaim` etc.

boilerplate/_lib/container-make
iamkirkbater: Good catch, this actually shouldn't be modified here.

pkg/controller/account/byoc.go
iamkirkbater: Good catch.  Updated to actually validate what matters for an STS account claim.

pkg/controller/account/iam.go
iamkirkbater: Done

iamkirkbater:
Codecov why you attack me like this :(

pkg/controller/account/account_controller.go
```diff
@@ -502,12 +534,20 @@ func (r *ReconcileAccount) assumeRole(
 	roleToAssume string,
 	ccsRoleID string) (awsclient.Client, *sts.AssumeRoleOutput, error) {
 
+	// The role ARN made up of the account number and the role which is the default role name
+	// created in child accounts
+	var roleArn = fmt.Sprintf("arn:aws:iam::%s:role/%s", currentAcctInstance.Spec.AwsAccountID, roleToAssume)
```
dkeohane, line 539: Similar to `adminAccessArn` const, we could move this string into its own const like `roleArnFormat`

pkg/controller/account/iam.go
```diff
@@ -69,18 +69,12 @@ func (r *ReconcileAccount) CreateSecret(reqLogger logr.Logger, account *awsv1alp
 	return nil
 }
 
-// getStsCredentials returns STS credentials for the specified account ARN
+// getSTSCredentials returns STS credentials for the specified account ARN
```
dkeohane, line 72: Minor: Update comment for new params list

pkg/controller/account/account_controller.go
iamkirkbater: Hmm, I'm not sure if I fully understand what you're trying to suggest.  Do you mean make a separate var like this: `roleArnFormat := "arn:aws:iam:%s:role/%s"`?

I think I'd have to see it in a separate PR in order to fully understand the scope of the change, but also I'm not sure if it's worth pulling out, if that will make the code easier to read or harder.  🤔   If you want, we can create a card for this and I think this would be a good chance for you to get familiar with the operator code as a first small PR?

dkeohane: Yup no worries, given this has been merged now anyways :slightly_smiling_face: 

The change was very minor/non-invasive, just to move the string into a const similar to [adminAccessArn](https://github.com/openshift/aws-account-operator/blob/master/pkg/controller/account/account_controller.go#L74). I can agree that it might make the code harder to read having to jump up to the top of the file - it was mainly from a code cleanliness viewpoint not to have hardcoded strings set in functions. I don't think it merits a card tbh - was only a small comment :smile:

PR #520, opened 2021-03-11
Prevents possible race condition while trying to access byoc secret

ArielLima:
Card -> https://issues.redhat.com/browse/OSD-6386

### DIFF
--- a/pkg/controller/account/byoc.go
+++ b/pkg/controller/account/byoc.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"time"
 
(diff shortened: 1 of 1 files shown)


iamkirkbater:
/hold

Holding for anyone else to share their thoughts. Ari and I discussed this implementation before he completed it and we figured this would be the easiest way to head-off any possible race condition without spending much more time than necessary on it.

iamkirkbater:
/lgtm

Local tests passed for me.  Still holding for comment.  Will hold until Wednesday sometime.

rogbas:
This is looking good, and if we merge it today we can get it promoted on time for tomorrow's load test.

/hold cancel

PR #526, opened 2021-03-24
Auto Author Assign

drpaneas:
In most cases, pull request author should be assigned as assignee of the pull request. This action automatically assigns PR author
as an assignee.


iamkirkbater:
What does being an assignee entail or mean?

drpaneas:
> What does being an assignee entail or mean?

Look at the right side of GitHub, it says:

```
Assignees
No one -- assign yourself
```

It basically pings you when someone comments and in case you use any GitHub automation in the future, you can use this field accordingly. e.g. for GitHub Projects the automation can move the card to ToDo list taken from the Backlog list.

rogbas:
+1, that's super nice.

/lgtm
/approve

2uasimojo:
> misconfigured

Yeah, the file is empty, which I thought might mean it had some magical default behavior based on the name, but (thankfully) that doesn't seem to be the case. Looking at one of the [runs](https://github.com/2uasimojo/aws-account-operator/actions/runs/883561188) it complains "No event triggers defined in `on`".

https://github.com/marketplace/actions/auto-author-assign shows some sample content that I guess we should adopt/adapt.

That's assuming we want to do this in the first place. Having the author as an assignee doesn't make a lot of sense to me, personally, as GH is set up by default to notify the author on PR activity. If not, IMO this should be a [personal choice](https://github.com/settings/notifications), not foisted upon me by the repo.

PR #530, opened 2021-03-26
Update deprecated operator-sdk commands 

dastergon:
The flag `--local` has been deprecated in the last release of operator-sdk (0.19.x) before 1.0 and we should use `run local` instead. Also, flag `--namespace` has been deprecated as well, and we should use `--watch-namespace` instead.

PTAL

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -290,7 +290,7 @@ predeploy: predeploy-aws-account-operator deploy-aws-account-operator-credential
 
 .PHONY: deploy-local
 deploy-local:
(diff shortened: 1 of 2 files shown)


drpaneas:
Have you verified the operator builds successfully with this version of operator-sdk?
Looking at https://github.com/openshift/aws-account-operator/blob/2b6621dc54ed31b8b3a86a936f8a48c40322edda/go.mod#L16 says `v0.16.0`. I would suggest to update this dependency first ;)

dastergon:
It looks like we will have to wait until we get to a bump in v0.16.0 in a-a-o. I will close this ticket and send another PR once we are able to work on it in a future sprint.

PR #532, opened 2021-03-29
Delete hanging micro instances on account creation

ArielLima:
CARD -> https://issues.redhat.com/browse/OSD-5810

### DIFF
--- a/pkg/awsclient/mock/zz_generated.mock_client.go
+++ b/pkg/awsclient/mock/zz_generated.mock_client.go
@@ -147,6 +147,21 @@ func (mr *MockClientMockRecorder) DeleteSnapshot(arg0 interface{}) *gomock.Call 
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "DeleteSnapshot", reflect.TypeOf((*MockClient)(nil).DeleteSnapshot), arg0)
 }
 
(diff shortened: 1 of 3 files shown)


pkg/awsclient/mock/zz_generated.mock_client.go
```diff
@@ -1,5 +1,5 @@
 // Code generated by MockGen. DO NOT EDIT.
-// Source: ./client.go
+// Source: pkg/awsclient/client.go
```
iamkirkbater, line 2: @ArielLima this looks like you got hit with the same-ish `generate` issue that I was running into a few weeks ago.

Try using `./boilerplate/_lib/container-make generate` and then commit those changes.

iamkirkbater:
Other than the issue with the generated file, this works well.  To test I ran `make create-account` and then `^C`'d the operator right after it spun up the instances.  Double checked that the instances didn't go away right away, and then ran the operator again, and the instances then went away.

Then I went back to the `master` branch, did that a few more times so I had 2-3 hanging instances, and then ran the `make test-account-creation` and watched it clean up the account.

Well done @ArielLima!!

iamkirkbater:
This still LGTM.  Can we squash this and then I'll :+1:?

PR #533, opened 2021-04-01
Ignore .envrc

dastergon:
`.envrc` is used with `direnv`.

PTAL


iamkirkbater:
I'd also recommend adding `.envrc` to your global git ignore as well.  Good catch, though.

PR #514, opened 2021-02-18
Add script to generate configmap

jharrington22:
Add script to generate configmap for running the aws-account-operator.

Fix template to use a parameter reference for the namespace as oc template will strip out the value if its fixed.

PR #535, opened 2021-04-06
Improve ConfigMap creation & deletion process

dastergon:
A few updates:
* Added the `check-ou-mapping-configmap-env` target to execute before we create the ConfigMap, because missing the BASE and ROOT OU IDs will cause the Operator to return errors. 
*  Since we merged #514, I use the script to generate the ConfigMap. I am not sure about the VCPU Quota default to 1, but I believe it should be fine. 
* Fix `delete-ou-map` to explicitly set the OPERATOR_NAMESPACE

PTAL


iamkirkbater (approved):
This looks sane.  Thanks!

/lgtm

PR #538, opened 2021-04-07
Decouple Makefile

dastergon:
* Cleaned up Makefile from duplicate code 
* Introduced the `deploy-local-debug` target to run the operator locally with delve server enabled
* Fixed `create-accountclaim` target because it overlapped with `create-account`.

PTAL

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,8 +70,7 @@ delete-accountclaim-namespace: ## Delete account claim namespace
 
 .PHONY: create-accountclaim
 create-accountclaim: ## Create AccountClaim
-	# Create Account
-	test/integration/api/create_account.sh
+	$(MAKE) create-account
 	# Create accountclaim
 	@oc process --local -p NAME=${ACCOUNT_CLAIM_NAME} -p NAMESPACE=${ACCOUNT_CLAIM_NAMESPACE} -f hack/templates/aws.managed.openshift.io_v1alpha1_accountclaim_cr.tmpl | oc apply -f -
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -95,24 +93,19 @@ create-awsfederatedrole: ## Create awsFederatedRole "Read Only"
 delete-awsfederatedrole: ## Delete awsFederatedRole "Read Only"
 	# Delete Federated role
 	@oc delete -f deploy/crds/aws.managed.openshift.io_v1alpha1_awsfederatedrole_readonly_cr.yaml
-	# Delete Account
-	test/integration/api/delete_account.sh
-	# Delete Secrets
-	test/integration/api/delete_account_secrets.sh
+	$(MAKE) delete-account || true
```
iamkirkbater, line 96: Can you verify that this works with `make test-all`?

I remember running into an issue last year where `make` only calls a target once, so if you have multiple `create-account` calls, it only actually calls it once, which is why this all got sent to live in the shell scripts.

I don't think we were calling make explicitly like you are now, though, we were using it as the prereq targets (if that's what they're called) so it wasn't calling upon `make` a second time but using regular make logic.

dastergon: @iamkirkbater good catch, I fixed the create-account issue. The `$(MAKE)` works with `test-all`. I have updated the PR. PTAL

iamkirkbater:
Tested Locally.

/lgtm

PR #539, opened 2021-04-07
Add Route53 VPC Zone Association Rules

dustman9000:
Actions needed for PrivateLink and VPC to private zone associations.

https://issues.redhat.com/browse/OSD-6928

Ref: 
https://docs.aws.amazon.com/Route53/latest/APIReference/API_AssociateVPCWithHostedZone.html
https://github.com/openshift/hive/blob/master/docs/awsprivatelink.md#using-aws-private-link

### DIFF
--- a/deploy/crds/aws.managed.openshift.io_v1alpha1_awsfederatedrole_networkmgmt_cr.yaml
+++ b/deploy/crds/aws.managed.openshift.io_v1alpha1_awsfederatedrole_networkmgmt_cr.yaml
@@ -52,6 +52,7 @@ spec:
           - "ec2:ModifyNetworkInterfaceAttribute"
           - "ec2:DeleteNetworkInterface"
           - "ec2:CreateNetworkInterfacePermission"
+          - "ec2:DescribeVpcs"
         resource:
           - "*"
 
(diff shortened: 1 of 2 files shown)


deploy/crds/aws.managed.openshift.io_v1alpha1_awsfederatedrole_networkmgmt_cr.yaml
```diff
@@ -82,6 +83,11 @@ spec:
           - "route53resolver:GetResolverRuleAssociation"
           - "route53resolver:ListResolverRuleAssociations"
           - "route53resolver:ListResolverRules"
+          - "route53:CreateVPCAssociationAuthorization"
+          - "route53:DeleteVPCAssociationAuthorization"
+          - "route53:ListVPCAssociationAuthorizations"
```
abhinavdahiya, line 88: these shouldn't be required. these permissions are only required in the PrivateLink account where the hosted zone exists.

i think we need
```
route53:AssociateVPCWithHostedZone
route53:DisassociateVPCFromHostedZone

ec2:DescribeVpcs
```

the last one coming from AWS requirement.. https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/r53-api-permissions-ref.html#required-permissions-private-hosted-zones

dustman9000: @abhinavdahiya - The reason I added these was due to the Note here for AWS's API documentation: https://docs.aws.amazon.com/Route53/latest/APIReference/API_AssociateVPCWithHostedZone.html

I can remove for now if you feel we can move forward without them.

dustman9000: I've removed those for now, hopefully we will not need them.

PR #544, opened 2021-04-12
OSD-6975: Only evaluate AWS_IAM_ARN when needed

2uasimojo:
Move the evaluation of `AWS_IAM_ARN` into the `check-aws-account-id-env` target to avoid it being evaluated a) for targets where it's unused; and b) for autocomplete (where it's also unused).

Tested the following permutations:
- Unset in the environment
  - Autocomplete doesn't attempt to set it
  - `check-aws-account-id-env` attempts to set it. On failure, an error is output and the value is empty; so the `check-aws-account-id-env` target fails as desired.
- Set in the environment
  - Autocomplete doesn't attempt to set it
  - `check aws-account-id-env` succeeds with "nothing to do" (because the `ifndef` directives end us up with an empty target).

[OSD-6975](https://issues.redhat.com/browse/OSD-6975)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,8 +6,6 @@ REUSE_BUCKET_NAME=test-reuse-bucket-${REUSE_UUID}
 
 include hack/scripts/test_envs
 
-export AWS_IAM_ARN := $(shell aws sts get-caller-identity --profile=osd-staging-2 | jq -r '.Arn')
-
 # Boilerplate
 include boilerplate/generated-includes.mk
 
@@ -30,7 +28,8 @@ ifndef OSD_STAGING_2_AWS_ACCOUNT_ID
 	$(error OSD_STAGING_2_AWS_ACCOUNT_ID is undefined)
 endif
 ifndef AWS_IAM_ARN
-	$(error AWS_IAM_ARN is undefined)
+	$(eval export AWS_IAM_ARN := $(shell aws sts get-caller-identity --profile=osd-staging-2 | jq -r '.Arn'))
+	@if [[ -z "$(AWS_IAM_ARN)" ]]; then echo "AWS_IAM_ARN unset and could not be calculated!"; exit 1; fi
 endif
 
 .PHONY: check-ou-mapping-configmap-env


iamkirkbater:
I remember there was some discussion on slack.  Do we still need/want this?

rogbas:
@lisa @rogbas please follow up and cmmnt

Makefile
```diff
@@ -30,7 +28,8 @@ ifndef OSD_STAGING_2_AWS_ACCOUNT_ID
 	$(error OSD_STAGING_2_AWS_ACCOUNT_ID is undefined)
 endif
 ifndef AWS_IAM_ARN
-	$(error AWS_IAM_ARN is undefined)
+	$(eval export AWS_IAM_ARN := $(shell aws sts get-caller-identity --profile=osd-staging-2 | jq -r '.Arn'))
+	if [[ -z "$(AWS_IAM_ARN)" ]]; then $(error AWS_IAM_ARN unset and could not be calculated!); fi
```
lisa, line 32: It appears that the only instances of `AWS_IAM_ARN` in this Makefile appear here (and, previously, in the "set" operation formerly on line 9). If the goal is to only evaluate this when needed, let's move this setting into the [shell script](https://github.com/openshift/aws-account-operator/blob/27a325bfa3ff9c467741f41238353e5cd706378d/test/integration/delete_awsfederatedaccountaccess.sh) where it's used directly. The `create-accountclaim-namespace` and `delete-accountclaim-namespace` Make targets could be wrapped in a shell script and have this `sts get-caller-identity` call done there.

lisa (changes requested):

Makefile
2uasimojo: I did this, but I don't like it as much. The reason is that some of the bigger test targets, like [test-awsfederatedrole](https://github.com/openshift/aws-account-operator/blob/5e5feb5ee2de5662a7f8aa26e334d7bbb93abb7d/Makefile#L115-L120), run these scripts multiple times, and calculating the variable on each script invocation will slow things down significantly. You could get around that by overriding it from the env -- but you shouldn't need to. And the way this was set up before, you didn't have to.

Yes, we could wrap each of the three targets that use this in an uberscript, but that seems like unnecessary indirection/obfuscation.

2uasimojo: I also didn't realize we were including `test_envs` in the Makefile and in shell scripts. That seems like a brittle plan, as it relies on only the subset of syntax that's common to Makefile and bash, and ignores any differences in behavior between the two.

But I'm not going to fix that right away because, as noted above, I really don't like this solution and would prefer to revert it to the way it was before. Thoughts on that?

lisa: > I did this, but I don't like it as much. The reason is that some of the bigger test targets, like [test-awsfederatedrole](https://github.com/openshift/aws-account-operator/blob/5e5feb5ee2de5662a7f8aa26e334d7bbb93abb7d/Makefile#L115-L120), run these scripts multiple times

Ahh, I didn't realize this was the case. That's unfortunate. :(

2uasimojo:
/hold

@lisa showed me this is done broke.

2uasimojo:
/hold cancel

Fixed. Thanks @lisa

lisa (approved):
/lgtm
/approve

grr make, grr.

PR #545, opened 2021-04-12
apis: Move to separate module

vkareh:
To allow for better separation of API versus business logic, we move the apis to be its own module within the operator code. This allows external projects to import just the API as necessary without pulling in potentially-conflicting dependencies.

Also remove api dependency on controller-runtime to allow OCM to import the necessary APIs.

Please see https://issues.redhat.com/browse/OSD-6948


iamkirkbater:
Integration tests all ran fine locally.

jharrington22:
/lgtm

@rogbas @iamkirkbater I'll leave it to you to approve.

PR #546, opened 2021-04-13
Poc boilerplate appsre

2uasimojo:
/hold

This is what things will look like once https://github.com/openshift/boilerplate/pull/136 and https://github.com/openshift/boilerplate/pull/137 are merged and pulled into a-a-o; and the second commit herein is stacked on top of that.


2uasimojo:
The real thing is at #547

PR #548, opened 2021-04-14
Fix to errors seen in stage after merge of last PR

ArielLima:
Prevent erroring out from attempted account cleanup. Instead logging the error and continuing to account initialization normally.

### DIFF
--- a/pkg/controller/account/ec2.go
+++ b/pkg/controller/account/ec2.go
@@ -88,8 +88,10 @@ func (r *ReconcileAccount) InitializeRegion(
 	reqLogger.Info("initializing region", "region", region)
 
 	// Attempt to clean the region from any hanging resources
-	cleaned, err := cleanRegion(awsClient, reqLogger)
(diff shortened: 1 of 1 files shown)


fahlmant (approved):
/lgtm

fahlmant (approved):
/lgtm

pkg/controller/account/ec2.go
```diff
@@ -88,8 +88,10 @@ func (r *ReconcileAccount) InitializeRegion(
 	reqLogger.Info("initializing region", "region", region)
 
 	// Attempt to clean the region from any hanging resources
-	cleaned, err := cleanRegion(awsClient, reqLogger)
+	cleaned, err := cleanRegion(awsClient, reqLogger, account.Name)
 	if err != nil {
+		cleanErr := fmt.Sprintf("Error", err.Error(), "account", account.Name)
```
rogbas, line 93: @ArielLima I suggest having a more descriptive msg than Error here.

The previous written msg would work fine IMO, and then you could append the aws err string as you're doing

PR #554, opened 2021-04-19
apis: Move back to pkg/apis

vkareh:
Since the operator-sdk expects types to be in the pkg/apis directory, we
move them back and ensure that there is a go.mod file to allow importing
this directory individually.


drpaneas:
Fixed in https://github.com/openshift/aws-account-operator/pull/580/files

PR #558, opened 2021-04-21
create a read-only aws role for backplane

ArielLima:
As part of backplane M2 we need to ensure that all clusters (current and new) have a specific IAM role with read-only permissions.
Card: https://issues.redhat.com/browse/OSD-6444
Tested with `make`
This PR Ensure:
- The role is backfilled in to existing clusters

- New clusters will contain the new role


pkg/controller/account/account_controller.go
```diff
@@ -261,6 +261,42 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
+	}
+
+	tags := awsclient.AWSTags.BuildTags(currentAcctInstance).GetIAMTags()
+	roleID, err := CreateAWSManagedRole(reqLogger, awsSetupClient, awsAssumedRoleClient, awsv1alpha1.AWSManagedRoleArn, tags, SREAccessARN)
+	if err != nil {
+		reqLogger.Error(err, "Failed to sync AWSManagedRole")
+		return reconcile.Result{}, err
+	}
+	reqLogger.Info("AWSManagedRole created", "RoleID", roleID, "account", currentAcctInstance.Name)
+
```
iamkirkbater, line 299: I think this block needs to get moved back down to where it was, as we explicitly do _not_ want to create roles in STS-based accounts.

pkg/apis/aws/v1alpha1/shared_types.go
```diff
@@ -152,3 +152,9 @@ var DefaultConfigMap = "aws-account-operator-configmap"
 
 // DefaultConfigMapAccountLimit holds the fallback limit of aws-accounts
 var DefaultConfigMapAccountLimit = 100
+
+// AWSManagedRoleName is the name of the read-only role
+var AWSManagedRoleName = "AWSManagedRole"
+
+// AWSManagedRoleArn is the rean-only ARN for the AWSManagedRole
```
iamkirkbater, line 159: s/rean/read

pkg/apis/aws/v1alpha1/shared_types.go
```diff
@@ -152,3 +152,9 @@ var DefaultConfigMap = "aws-account-operator-configmap"
...
+var AWSManagedRoleArn = "arn:aws:iam::aws:policy/ReadOnlyAccess"
```
iamkirkbater, line 160: I think this might work well as a first pass, but as CEE's needs evolve and as we add more roles I'm not convinced this approach will scale.

iamkirkbater (changes requested):

pkg/controller/account/account_controller.go
ArielLima: The reason I moved it out is because I thought that since it was inside `currentAcctInstance.ReadyForInitialization()` it would only create the role when accounts are initialized and not create roles in old accounts. I could be wrong? Would an alternative just be checking `currentAcctInstance.Spec.ManualSTSMode`?

pkg/apis/aws/v1alpha1/shared_types.go
ArielLima: +1

pkg/apis/aws/v1alpha1/shared_types.go
ArielLima: I completely agree with this. The reason I did it this way was because I was trying to remain within the scope of the card.

pkg/controller/account/account_controller.go
ArielLima: I updated this to not create for STS accounts

lisa:
@ArielLima please rebase

dastergon:
@ArielLima PR needs rebase, and we can proceed

lisa:
We're going in a different direction with this. See OSD-8074.

PR #559, opened 2021-04-21
Instead of outputting cVPU quota error just log the defaulting value

ArielLima:
Card: https://issues.redhat.com/browse/OSD-6848

Instead of outputting vCPU quota error  simply log that the vCPU value was not found in the configmap and that the value is being defaulted to 0

### DIFF
--- a/pkg/controller/account/ec2.go
+++ b/pkg/controller/account/ec2.go
@@ -346,13 +346,13 @@ func (r *ReconcileAccount) getDesiredVCPUValue(reqLogger logr.Logger) (float64, 
 		err = awsv1alpha1.ErrInvalidConfigMap
 	}
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controller/account/ec2.go
```diff
@@ -346,13 +346,13 @@ func (r *ReconcileAccount) getDesiredVCPUValue(reqLogger logr.Logger) (float64,
 		err = awsv1alpha1.ErrInvalidConfigMap
 	}
 	if err != nil {
-		reqLogger.Error(err, "failed getting desired vCPU quota from configmap data")
+		reqLogger.Info(fmt.Sprintf("Failed getting desired vCPU quota from configmap data, defaulting quota to: %v", 0))
```
dkeohane, line 349: Minor: No need to Sprintf if you know we're defaulting to 0, can just log the string

PR #561, opened 2021-04-22
Initialize only the Regions available in the cluster account

ArielLima:
During the region initialization phase, initialize only the regions enabled for the account, instead of using a hardcoded array of covered regions.
Card: https://issues.redhat.com/browse/OSD-6825


pkg/apis/aws/v1alpha1/shared_types.go
```diff
@@ -18,9 +18,15 @@ var CoveredRegions = map[string]map[string]string{
 	"us-west-2": {
 		"initializationAMI": "ami-0a7e1ebfee7a4570e",
 	},
+	"af-south-1": {
+		"initializationAMI": "ami",
```
iamkirkbater, line 22: I think this is missing.

ArielLima: Hence the hold

fahlmant:
Given that we're moving away from this static list, maybe moving the ami IDs into a configmap would help here? In case the ami IDs change. Then a configmap update would be required, rather than a full new deploy of the operator.

ArielLima:
Still missing a few region AMIs since the regions are taking a while to initialize

> Given that we're moving away from this static list, maybe moving the ami IDs into a configmap would help here? In case the ami IDs change. Then a configmap update would be required, rather than a full new deploy of the operator.

I think this would be a good change

pkg/controller/account/account_controller.go
```diff
@@ -617,9 +629,29 @@ func (r *ReconcileAccount) initializeRegions(reqLogger logr.Logger, currentAcctI
...
+	})
+	if err != nil {
+		reqLogger.Error(err, "Failed to retrieve list of regions enabled in this account.")
+		return err
+	}
+
 	// This initializes supported regions, and updates Account state when that's done. There is
 	// no error checking at this level.
-	go r.asyncRegionInit(reqLogger, currentAcctInstance, creds)
+	go r.asyncRegionInit(reqLogger, currentAcctInstance, creds, regionAMIs, regionsEnabledInAccount)
```
iamkirkbater, line 654: NOOP for today: I feel like it would be cleaner to just parse the list of regions enabled and only pass the list of AMIs that it needs to in here, but that's a large refactor here, and can be tackled later.  Let's just keep that in mind.

pkg/controller/account/ec2.go
```diff
@@ -64,7 +65,7 @@ func (r *ReconcileAccount) InitializeRegion(
 	reqLogger logr.Logger,
 	account *awsv1alpha1.Account,
 	region string,
-	ami string,
+	instanceInfo awsv1alpha1.InstanceInfo,
```
iamkirkbater, line 68: I'm not sold that `InstanceInfo` is a good name for this.  At the beginning of my review it kind of made sense but when I'm looking at this now after being distracted it took a second look to realize what this was actually trying to pass in.

Maybe `InstanceParams`?  Idk if this sounds better though, because then I look at some of the other things passed in like the vCPUQuota and wonder why that wouldn't be passed in with the `InstanceParams`....

Maybe `AmiSpec` or something?

iamkirkbater (changes requested):

hack/templates/aws.managed.openshift.io_v1alpha1_configmap.tmpl
```diff
@@ -20,4 +20,26 @@ objects:
...
+      ap-northeast-1:ami-00b95502a4d51a07e:t2.micro
+      ap-northeast-2:ami-041b16ca28f036753:t2.micro
+      ap-northeast-3:ami-070dd2ec8c4a6df38:t2.micro
+      ap-east-1:ami-f4fab885:t3.micro
+      ap-south-1:ami-0963937a03c01ecd4:t2.micro
+      ap-southeast-1:ami-055c55112e25b1f1f:t2.micro
+      ap-southeast-2:ami-036b423b657376f5b:t2.micro
+      sa-east-1:ami-05c1c16cac05a7c0b:t2.micro
+      af-south-1:ami-0f4b49fefef9be45a:t3.micro
+      me-south-1:ami-0b41a37a62a4296fc:t3.micro
```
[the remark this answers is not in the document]
iamkirkbater: This is the only one that needs to be taken care of/decided on before we merge.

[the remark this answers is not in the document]
cblecker: @iamkirkbater You mean to https://github.com/openshift/aws-account-operator/blob/123eecb5c59940db997530a45ff05aeeaf0c8355/hack/olm-registry/olm-artifacts-template.yaml#L66-L72 ?

I'm personally fine with it residing here with the operator (as it was before when it was embedded in go), rather than in app-interface unless there is a strong reason we should move it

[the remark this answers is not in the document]
iamkirkbater: Yeah, either there or we define it here in the operator.  The way it's written right now it's _expecting_ it in the configmap, but because it's not in the OLM Template it's not going to actually be in the configmap.  

We should be able to hard-code it in that file, though.

pkg/controller/account/account_controller.go
ArielLima: I agree with you, the only reason I did it this way is that it is the last layer in which I can error out normally. Happy to change it in the future though.

pkg/controller/account/ec2.go
ArielLima: +1

iamkirkbater (approved):

PR #562, opened 2021-04-22
Adds tags to volumes on RunInstances calls

iamkirkbater:
For Conditional IAM policies to work, RunInstances must also be passed a
tag list for volume resource types, otherwise the RunInstances call will
not be able to create the backing volume it needs to run the instance.

Satisfies [OSD-6625](https://issues.redhat.com/browse/OSD-6625)

Depends on #572 to be merged first.  View only the commits from this PR here: [`0b4f8d8` (#562)](https://github.com/openshift/aws-account-operator/pull/562/commits/0b4f8d8a5711e8e449515554a6b6c594f9e0256e)

### DIFF
--- a/deploy/crds/aws.managed.openshift.io_accountclaims_crd.yaml
+++ b/deploy/crds/aws.managed.openshift.io_accountclaims_crd.yaml
@@ -86,6 +86,8 @@ spec:
                 - name
                 - namespace
               type: object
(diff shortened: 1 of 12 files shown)


dkeohane (changes requested):

dastergon (approved):
/lgtm

PR #563, opened 2021-04-26
OSD-6906: Relocate fields from .status to .spec

drpaneas:
Move the following fields from .status to .spec in the Account CR. It's not only a best practice on CRD design, but in order to support migrating clusters, we need to have all fields we want to persist as part of the spec object, otherwise they will be lost from the status after migration.

### DIFF
--- a/pkg/controller/accountpool/accountpool_controller.go
+++ b/pkg/controller/accountpool/accountpool_controller.go
@@ -113,7 +113,7 @@ func (r *ReconcileAccountPool) Reconcile(request reconcile.Request) (reconcile.R
 	claimedAccountCount := 0
 	for _, account := range accountList.Items {
 		// We don't want to count reused accounts here, filter by LegalEntity.ID
-		if !account.Status.Claimed && account.Spec.LegalEntity.ID == "" {
+		if !account.Spec.Claimed && account.Spec.LegalEntity.ID == "" {
 			if !account.IsFailed() {
 				unclaimedAccountCount++
 			}
--- a/pkg/controller/accountpool/accountpool_controller_test.go
+++ b/pkg/controller/accountpool/accountpool_controller_test.go
@@ -78,12 +78,12 @@ func TestReconcileAccountPool(t *testing.T) {
 							ID:   "",
 							Name: "",
 						},
+						Claimed:       false,
+						State:         "Ready",
+						SupportCaseID: "000000",
 					},
 					Status: awsv1alpha1.AccountStatus{
-						Claimed:           false,
-						SupportCaseID:     "000000",
 						Conditions:        []awsv1alpha1.AccountCondition{},
-						State:             "Ready",
 						RotateCredentials: false,
 					},
 				},
@@ -100,12 +100,12 @@ func TestReconcileAccountPool(t *testing.T) {
 							ID:   "",
 							Name: "",
 						},
+						Claimed:       false,
+						State:         "Ready",
+						SupportCaseID: "000000",
 					},
 					Status: awsv1alpha1.AccountStatus{
-						Claimed:           false,
-						SupportCaseID:     "000000",
 						Conditions:        []awsv1alpha1.AccountCondition{},
-						State:             "Ready",
 						RotateCredentials: false,
 					},
 				},
@@ -242,8 +242,8 @@ func verifyAccountCreated(c client.Client, expected *awsv1alpha1.AccountPool) bo
 	unclaimedAccountCount := 0
 	for _, account := range al.Items {
 		// We don't want to count reused accounts here, filter by LegalEntity.ID
-		if account.Status.Claimed == false && account.Spec.LegalEntity.ID == "" {
-			if account.Status.State != "Failed" {
+		if account.Spec.Claimed == false && account.Spec.LegalEntity.ID == "" {
+			if account.Spec.State != "Failed" {
 				unclaimedAccountCount++
 			}
 		}
--- a/pkg/controller/utils/status.go
+++ b/pkg/controller/utils/status.go
@@ -16,7 +16,7 @@ func SetAccountStatus(awsAccount *awsv1alpha1.Account, message string, ctype aws
 		UpdateConditionNever,
 		awsAccount.Spec.BYOC,
 	)
-	awsAccount.Status.State = state
+	awsAccount.Spec.State = state
 }
 
 // SetAccountClaimStatus sets the condition and state of an accountClaim
--- a/pkg/localmetrics/localmetrics.go
+++ b/pkg/localmetrics/localmetrics.go
@@ -226,27 +226,27 @@ func (c *MetricsCollector) collect() {
 	}
 
 	for _, account := range accounts.Items {
-		if account.Status.Claimed {
+		if account.Spec.Claimed {
 			claimed = "true"
 		} else {
 			claimed = "false"
 		}
 
-		if account.Status.Reused {
+		if account.Spec.Reused {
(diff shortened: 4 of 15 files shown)


dkeohane (commented):
Changes seem fine by me - not sure if relevant to this PR or if it will come later, but calling out for my own sanity - would be good to update the docs to include a section on cluster migration and how the new fields are utilised

2uasimojo:
Clearly this is still a WIP, but things I noticed at a glance:

> Move the following fields [1] from .status to .spec

- There's no [1] footnote.
- The fields were copied, not moved. Based on the [comment](https://github.com/openshift/aws-account-operator/pull/563/files#diff-90c45f97547bcd6df670d5b7e479105aef1041da92e8a80d342cb0e280544563R46) I assume that was intentional. Is it a good idea to have them in both places though? Or is this for some transitional phase where the status fields are being deprecated; and a subsequent change will actually remove them?
- There's no code yet to set the fields :) but I wonder what happens if there's a discrepancy. Since spec and status updates have to be two separate commands, one may fail; which is the source of truth?
- I assume the user won't be allowed to set these fields. How will you handle it if they try?

(Are these kinds of design discussions happening in a card somewhere?)

drpaneas:
> Clearly this is still a WIP, but things I noticed at a glance:

Yes, this PR is far from finished. Expect to be open for a while, until all commits are done. There is lot's of experimentation. Many thanks for your comments -- they are really valuable.

> * There's no [1] footnote.

what do you mean by `footnote` ? can you give an examle?

> * The fields were copied, not moved. Based on the [comment](https://github.com/openshift/aws-account-operator/pull/563/files#diff-90c45f97547bcd6df670d5b7e479105aef1041da92e8a80d342cb0e280544563R46) I assume that was intentional. Is it a good idea to have them in both places though? Or is this for some transitional phase where the status fields are being deprecated; and a subsequent change will actually remove them?

The idea behind this change is to make sure we can preserve the state during hive relocation. So the spec fields are going to be the ones to do that. That means all the logic the operator has at the very moment (e.g. if status.state X, then do Y), it will rely only the the spec.State.


> * There's no code yet to set the fields :) but I wonder what happens if there's a discrepancy. Since spec and status updates have to be two separate commands, one may fail; which is the source of truth?

After this PR, the source of truth should always be the spec.fields.

> * I assume the user won't be allowed to set these fields. How will you handle it if they try?

Good question. I haven't though of that. I can try using https://github.com/operator-framework/operator-sdk/issues/1587 but after reading a bit about it, it seems tough (a newer version of operator might be needed to support this).

> (Are these kinds of design discussions happening in a card somewhere?)

I will send you the doc in Slack ;)

drpaneas:
> would be good to update the docs to include a section on cluster migration and how the new fields are utilised

yes, there will be an SOP for this.

drpaneas:
As of this moment, the reconciliation failed with https://github.com/openshift/aws-account-operator/blob/bf6f58eb7a119e41905eca4ff287d45b52fa9b5a/pkg/controller/account/account_controller.go#L657 this error. @dastergon feel free to debug this.

drpaneas:
It seems to be working fine now. Things to fix:

* When I do `oc get account osd-creds-mgmt-osd-staging-1 -n aws-account-operator` I don't see the status. I have to do `-o json` to see it.
* Go tests need to be adjusted to the new code. They are not passing at the moment. To test this against integration tests at the moment, go to the Makefile and remote `lint` and `test` from the `make test-all` target.

2uasimojo:
> * When I do `oc get account osd-creds-mgmt-osd-staging-1 -n aws-account-operator` I don't see the status. I have to do `-o json` to see it.

You should be able to tailor what you want to see via [kubebuilder print columns](https://book.kubebuilder.io/reference/generating-crd.html#additional-printer-columns). You’ll need to `make op-generate` afterwards to get them into the CRD. 

> * Go tests need to be adjusted to the new code. They are not passing at the moment. To test this against integration tests at the moment, go to the Makefile and remote `lint` and `test` from the `make test-all` target.

Best way to debug these is to use `./boilerplate/_lib/container-make test lint`. That’ll run the tests in a container that matches the CI environment.

pkg/localmetrics/localmetrics.go
```diff
@@ -226,27 +226,27 @@ func (c *MetricsCollector) collect() {
...
+		if account.Spec.Reused {
 			reused = "true"
 		} else {
 			reused = "false"
 		}
 
-		if !account.Status.Claimed && account.Status.Reused &&
-			account.Status.State == "Ready" {
+		if !account.Spec.Claimed && account.Spec.Reused &&
+			account.Spec.State == "Ready" {
```
dkeohane, line 242: Not your code, but can we replace this "Ready" string with the const equivalent  - if for some bizzare reason the value of 'Ready' changed in the future, we'd still be left with this hardcoded string.

pkg/controller/utils/status.go
```diff
@@ -16,7 +16,7 @@ func SetAccountStatus(awsAccount *awsv1alpha1.Account, message string, ctype aws
 		UpdateConditionNever,
 		awsAccount.Spec.BYOC,
 	)
-	awsAccount.Status.State = state
+	awsAccount.Spec.State = state
```
dkeohane, line 19: In this func we're passing in a state of type string, in the bellow func we're passing in a state of type `awsv1alpha1.ClaimStatus` -> can we update this func to be the same (or whatever the equivalent const type it should be)

pkg/controller/accountpool/accountpool_controller_test.go
```diff
@@ -242,8 +242,8 @@ func verifyAccountCreated(c client.Client, expected *awsv1alpha1.AccountPool) bo
 	unclaimedAccountCount := 0
 	for _, account := range al.Items {
 		// We don't want to count reused accounts here, filter by LegalEntity.ID
-		if account.Status.Claimed == false && account.Spec.LegalEntity.ID == "" {
-			if account.Status.State != "Failed" {
+		if account.Spec.Claimed == false && account.Spec.LegalEntity.ID == "" {
+			if account.Spec.State != "Failed" {
```
dkeohane, line 246: Refer to my request on hardcoded strings swapped with their consts

pkg/controller/accountpool/accountpool_controller.go
```diff
@@ -113,8 +113,8 @@ func (r *ReconcileAccountPool) Reconcile(request reconcile.Request) (reconcile.R
 	claimedAccountCount := 0
 	for _, account := range accountList.Items {
 		// We don't want to count reused accounts here, filter by LegalEntity.ID
-		if !account.Status.Claimed && account.Spec.LegalEntity.ID == "" {
-			if account.Status.State != "Failed" {
+		if !account.Spec.Claimed && account.Spec.LegalEntity.ID == "" {
+			if account.Spec.State != "Failed" {
```
dkeohane, line 117: Refer to my request on hardcoded strings swapped with their consts

dkeohane (commented):
Awesome job getting this done so quickly :taco:  (Especially before you're being taken off it!)

All my comments were pedantry, so not super important - and some we had already discussed in our sync up. Shout if you have questions :+1:

lisa:
/hold

There's some pre-work (on cluster) to do before this is merged and deployed, so let's take care to merge when we're ready.

drpaneas:
Is this still relevant?

NautiluX:
This is not strictly required since cluster relocation has been handled manually, so I think we should close this PR.

PR #567, opened 2021-04-30
Adds necessary field for Support ARN

iamkirkbater:
Reverts openshift/aws-account-operator#566

Adds back in the supportRoleARN fields now that the change freeze is (almost) lifted.

Originally was PR #564.  Satisfies the same OSD requirements.


iamkirkbater:
/hold cancel

App interface PR has merged.  This should be ready to move forward.

PR #569, opened 2021-05-06
Fix namespace in UHC ClusterRoleBinding

NautiluX:
* typo in namespace of ClusterRoleBinding

### DIFF
--- a/deploy/uhc_cluster_role_binding.yaml
+++ b/deploy/uhc_cluster_role_binding.yaml
@@ -4,7 +4,7 @@ metadata:
   name: aws-account-operator-client
 subjects:
 - kind: ServiceAccount
-  namespace: aws-account-operator-client
+  namespace: aws-account-operator
   name: aws-account-operator-client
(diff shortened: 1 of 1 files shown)


deploy/uhc_cluster_role_binding.yaml
```diff
@@ -4,11 +4,11 @@ metadata:
   name: aws-account-operator-client
 subjects:
 - kind: ServiceAccount
-  namespace: aws-account-operator-client
+  namespace: aws-account-operator
```
2uasimojo, line 7: Note to reviewers: This bug was only a problem for local testing. It is not causing chaos in real deployments because CRBs are generated by OLM from the CSV, which [only includes the service account *name*](https://github.com/openshift/aws-account-operator/blob/17d1683aedd534af2f974326f662b8f7423a3bfb/boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py#L241-L249).

PR #570, opened 2021-05-06
Avoids nil pointer dereference

iamkirkbater:
This should avoid our nil pointer dereference we found this morning.  We're attempting to update the status when we cannot get the accountclaim object, so therefore it's probably always nil here.

Fixes [OSD-7226](https://issues.redhat.com/browse/OSD-7226)

### DIFF
--- a/pkg/controller/account/byoc.go
+++ b/pkg/controller/account/byoc.go
@@ -47,16 +47,20 @@ func (r *ReconcileAccount) initializeNewCCSAccount(reqLogger logr.Logger, accoun
 	if acctClaimErr != nil {
 		// TODO: Unrecoverable
 		// TODO: set helpful error message
-		utils.SetAccountClaimStatus(
-			accountClaim,
-			"Failed to get AccountClaim for CSS account",
-			"FailedRetrievingAccountClaim",
-			awsv1alpha1.ClientError,
-			awsv1alpha1.ClaimStatusError,
-		)
-		err := r.Client.Status().Update(context.TODO(), accountClaim)
-		if err != nil {
-			reqLogger.Error(err, "failed to update accountclaim status")
+		if accountClaim != nil {
+			utils.SetAccountClaimStatus(
+				accountClaim,
+				"Failed to get AccountClaim for CSS account",
+				"FailedRetrievingAccountClaim",
+				awsv1alpha1.ClientError,
+				awsv1alpha1.ClaimStatusError,
+			)
+			err := r.Client.Status().Update(context.TODO(), accountClaim)
+			if err != nil {
+				reqLogger.Error(err, "failed to update accountclaim status")
+			}
+		} else {
+			reqLogger.Error(acctClaimErr, "accountclaim is nil")
 		}
 		return "", reconcile.Result{}, acctClaimErr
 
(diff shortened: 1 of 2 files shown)


pkg/controller/account/byoc.go
```diff
@@ -47,16 +47,20 @@ func (r *ReconcileAccount) initializeNewCCSAccount(reqLogger logr.Logger, accoun
...
-			"Failed to get AccountClaim for CSS account",
-			"FailedRetrievingAccountClaim",
-			awsv1alpha1.ClientError,
-			awsv1alpha1.ClaimStatusError,
-		)
-		err := r.Client.Status().Update(context.TODO(), accountClaim)
-		if err != nil {
-			reqLogger.Error(err, "failed to update accountclaim status")
+		if accountClaim != nil {
+			utils.SetAccountClaimStatus(
```
dkeohane, line 51: Not sure if it's the golang way for param validation but you could potentially move this `accountClaim != nil` check into the `SetAccountClaimStatus` function, as well,  to resolve this for other calls to this func

iamkirkbater: Ahh, that's a great idea!

iamkirkbater: Doesn't look like we have a logger there and I don't feel like adding _additional_ fields to that 🙄 so we'll just exit out for now.

pkg/controller/utils/status.go
```diff
@@ -21,6 +21,9 @@ func SetAccountStatus(awsAccount *awsv1alpha1.Account, message string, ctype aws
 
 // SetAccountClaimStatus sets the condition and state of an accountClaim
 func SetAccountClaimStatus(awsAccountClaim *awsv1alpha1.AccountClaim, message string, reason string, ctype awsv1alpha1.AccountClaimConditionType, state awsv1alpha1.ClaimStatus) {
+	if awsAccountClaim == nil {
+		return
+	}
```
jharrington22, line 26: @iamkirkbater this shouldn't be required or we should move the error handling here instead of adding this validation in two places.

2uasimojo: Everywhere this is called (once in account_controller.go, thrice in byoc.go), it is immediately followed by a `r.Client.Status().Update(context.TODO(), accountClaim)`. IMO it is worth refactoring to
- include that `Update` in this method;
- include this `nil` check; and
- return an `error` value to the caller.

iamkirkbater: We have a card open to refactor this directly into the AccountClaim struct itself: https://issues.redhat.com/browse/OSD-7194.

So this nil check will only persist until that is refactored, as if we can't get the accountClaim then there's no struct to call the update on, and we'll have to handle that in each of those calls respectively.

2uasimojo: At a 5-minute perusal, that refactor doesn't seem like it would be a whole lot bigger than this change. Why wait? :)

("This NPE is causing regular problems in the field" is a valid answer btw.)

iamkirkbater: "This NPE is causing regular problems in the field" ;)

Also, just from my personal experience "oh, this refactor should be really simple and quick" rarely happens to turn out that way hahaha.  At least with our current test suite :/  So I'd rather punt that and tackle it the right way then try to rush a refactor as a fix.

dastergon (approved):
/lgtm

PR #572, opened 2021-05-07
Adds support for Customer Provided Tags

iamkirkbater:
This PR adds support for Customer Provided Tags.  Fulfills OSD-7203.


pkg/awsclient/tags.go
```diff
@@ -46,25 +46,40 @@ func (t *AWSAccountOperatorTags) GetEC2Tags() []*ec2.Tag {
 }
 
 // BuildTags initializes AWSTags with required tags
-func (t *AWSAccountOperatorTags) BuildTags(account *awsv1alpha1.Account) AWSTagBuilder {
-	ClusterAccountNameTag := AWSTag{
+func (t *AWSAccountOperatorTags) BuildTags(account *awsv1alpha1.Account, managedTags []AWSTag, customTags []AWSTag) AWSTagBuilder {
```
dkeohane, line 49: Minor: Could merge the managed and custom tags to reduce the number of params - but I'd understand if you'd want to keep separate

pkg/controller/account/ec2.go
```diff
@@ -163,9 +168,9 @@ func (r *ReconcileAccount) InitializeRegion(
 }
 
 // BuildAndDestroyEC2Instances runs an ec2 instance and terminates it
-func (r *ReconcileAccount) BuildAndDestroyEC2Instances(reqLogger logr.Logger, account *awsv1alpha1.Account, awsClient awsclient.Client, ami string) error {
+func (r *ReconcileAccount) BuildAndDestroyEC2Instances(reqLogger logr.Logger, account *awsv1alpha1.Account, awsClient awsclient.Client, ami string, managedTags []awsclient.AWSTag, customerTags []awsclient.AWSTag) error {
```
dkeohane, line 171: Similar to my previous comment about params, but this method has quite a few params already

pkg/controller/account/account_controller.go
```diff
@@ -925,3 +925,72 @@ func getBuildIAMUserErrorReason(err error) (string, awsv1alpha1.AccountCondition
...
+	tags := []awsclient.AWSTag{}
+
+	accountClaim, err := r.getAccountClaim(account)
+	if err != nil {
+		log.Error(err, "Error getting AccountClaim to get custom tags")
+		return tags
+	}
+
+	if accountClaim.Spec.CustomTags == "" {
+		return tags
```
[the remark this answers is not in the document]
iamkirkbater: I think since managed tags are probably going to be defined on every cluster and they're the going to be the same across all instances, I'm going to leave the message the way it is.  Managed Tags not being defined means that every cluster doesn't get tags when they should, where there's only a very small amount of clusters that will define custom tags, so that would spam the logs a lot more and not really add much value.

Granted, if there were no managed tags defined it would spam the logs just as much, but I think that gets offset because the managed tags _should always_ be defined.  

I'm not quite sure if that makes sense the way it's written or not, but I'm happy to clarify further.

pkg/controller/account/ec2.go
iamkirkbater: I think I'm going to leave it this way for now as we're on a time crunch to get this in, but let's create a card and we can figure out a better way to pass these through.

Maybe it would be better to overload the awsclient with the tags, then we're not even passing them into the stack at all.  Then we can just add the tags up front and just pass the client through.

PR #580, opened 2021-05-18
Make API a separate pkg

drpaneas:
OCM need to import the API of the operator, but it cannot do it because it has conflicts with the version of controller-runtime library we are using. As a result, this PR removes the dependency for the controller-runtime library for the API by copy-pasting the `scheme` go file from upstream. It also makes the API a separate sub-module in this Go project, so OCM should be able to import _just that_ instead of the whole operator.

This PR addresses JIRA https://issues.redhat.com/browse/OSD-7287 ticket


iamkirkbater:
Wow, this was really fast! Will definitely be reviewing once I get my kids to school this morning!  Thanks a ton @drpaneas!!!

drpaneas:
operator-sdk doesn't seem to like this:


```
$ operator-sdk generate crds --verbose
DEBU[0000] Debug logging is set                         
INFO[0000] Running CRD generator.                       

-: pattern ./pkg/apis/...: main module (github.com/openshift/aws-account-operator) does not contain package github.com/openshift/aws-account-operator/pkg/apis
FATA[0034] error generating CRDs from APIs in pkg/apis: error generating CRD manifests: error running CRD generator: not all generators ran successfully 
```

drpaneas:
Going down the `controller-gen` route, seems to be working:

```
[drpaneas@pgeorgia-mac apis (changeapi ✗)]$ cd pkg/apis; controller-gen crd paths=./aws/v1alpha1 output:dir=../../deploy/crds
[drpaneas@pgeorgia-mac apis (changeapi ✗)]$ git status
On branch changeapi
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   aws/v1alpha1/account_types.go
	modified:   aws/v1alpha1/awsfederatedaccountaccess_types.go
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	../../deploy/crds/aws.managed.openshift.io_accountclaims.yaml
	../../deploy/crds/aws.managed.openshift.io_accountpools.yaml
	../../deploy/crds/aws.managed.openshift.io_accounts.yaml
	../../deploy/crds/aws.managed.openshift.io_awsfederatedaccountaccesses.yaml
	../../deploy/crds/aws.managed.openshift.io_awsfederatedroles.yaml
no changes added to commit (use "git add" and/or "git commit -a")
```

If this is the preferred way of moving forward, then we need to modify:
1. the makefile
2. the documentation
3. the boilerplate code
4. the CI that triggers operator-sdk (probably doing 3 should be enough)

drpaneas:
This is blocked until https://github.com/openshift/boilerplate/pull/171 gets merged.

Makefile
```diff
@@ -340,3 +340,7 @@ clean-operator: ## Clean Operator
 	oc delete accounts --all -n ${NAMESPACE}
 	oc delete awsfederatedaccountaccess --all -n ${NAMESPACE}
 	oc delete awsfederatedrole --all -n ${NAMESPACE}
+
+.PHONY: op-generate
+op-generate: ## Generate CRDs and k8s
+  ./generate_crds.sh
```
2uasimojo, line 346: This needs to be tab
```suggestion
	./generate_crds.sh
```

generate_crds.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
```
2uasimojo, line 1: Suggest moving this into `hack/`.

hack/scripts/generate_crds.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+set -eo pipefail
+
+# This bash is the replacement for: ./boilerplate/_lib/container-make op-generate
+#                               or:  make -n generate
+
+# What you need to run this
+# 1. operator-sdk v0.16.0
+# 2. CONTROLLER_GEN_VERSION=v0.3.0
+#    go get sigs.k8s.io/controller-tools/cmd/controller-gen@${CONTROLLER_GEN_VERSION}
```
iamkirkbater, line 10: THANK YOU SO MUCH for providing install instructions :joy: <3 <3 <3

jharrington22:
@drpaneas this looks reasonable to me.

To fix CI add the [ensure.sh](https://github.com/openshift/aws-account-operator/blob/master/boilerplate/openshift/golang-osd-operator/operator-sdk-generate.sh#L21) call to get operator-sdk installed.

This make target is performing the same function as the boilerplate make target than I think we're good to merge. Overwriting this is only temporary as we can update boilerplate to take these features in if all is working as expected. Slowly removing some of our dependencies on operator-sdk.

This will no effect the running on the AWS account operator, we know separating the APIs as a package already works from [here](https://github.com/openshift/aws-account-operator/pull/545) hive is doing the same thing [here](https://github.com/openshift/hive/pull/1324).

hack/scripts/generate_crds.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+set -eo pipefail
+
+# This bash is the replacement for: ./boilerplate/_lib/container-make op-generate
+#                               or:  make -n generate
+
+# What you need to run this
```
2uasimojo, line 7: Or run it in the boilerplate backing image, which [contains](https://github.com/openshift/boilerplate/blob/753a9302de749bac0542e0cf4e8787f0ee389df3/config/build.sh) all of these things already.

drpaneas: True, that's actually where I got all this stuff from.

dkeohane (approved):

hack/scripts/generate_crds.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+set -eo pipefail
+
+# This bash is the replacement for: ./boilerplate/_lib/container-make op-generate
+#                               or:  make -n generate
```
2uasimojo, line 5: You're only overriding `op-generate` in your Makefile, but you're providing the guts of `go-generate` and `openapi-generate` in this script. Seems those could have been omitted, yes?

hack/scripts/generate_crds.sh
```diff
@@ -0,0 +1,68 @@
...
+# Don't forget to commit generated files
+echo "--> Generating API files using openapi-gen ..."
+find ./pkg/apis/ -maxdepth 2 -mindepth 2 -type d | xargs -t -n1 -I% \
+		openapi-gen --logtostderr=true \
+			-i % \
+			-o "" \
+			-O zz_generated.openapi \
+			-p % \
+			-h /dev/null \
+			-r "-"
```
2uasimojo, line 68: As noted above, this should all be redundant with `go-generate` and `openapi-generate`, which are still being run under the `generate` umbrella.

2uasimojo: fixing this as part of #585

PR #582, opened 2021-05-21
[OSD-7072] Adding new role to non-STS accounts

dkeohane:
[OSD-7072](https://issues.redhat.com/browse/OSD-7072)
- Now calling `createBYOCAdminAccessRole` for all non-STS accounts within the `account_controller.go`
- Removed existing call to `createBYOCAdminAccessRole` from `byoc.go` in lieu of the change above.
- **To Note:** the big if block added to `account_controller.go` could be refactored down even further if we did not have the dependency on the RoleID within the `assumeRole` func - there's an [outstanding question over its use](https://coreos.slack.com/archives/CNJPJSG3Z/p1623934075068100), therefore this refactor may still occur

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -213,7 +213,7 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 		var result reconcile.Result
 		var initErr error
 
-		result, initErr = r.initializeNewCCSAccount(reqLogger, currentAcctInstance, awsSetupClient, adminAccessArn)
+		result, initErr = r.initializeNewCCSAccount(reqLogger, currentAcctInstance)
 		if initErr != nil {
 			// TODO: If we have recoverable results from above, how do we allow them to requeue if state is failed
 			_, stateErr := r.setAccountFailed(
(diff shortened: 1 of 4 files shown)


pkg/controller/account/account_controller.go
```diff
@@ -323,12 +319,105 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
+
+			if err != nil {
+				reqLogger.Error(err, "createBYOCAdminAccessRole err", roleID)
+				return reconcile.Result{}, err
+			}
+		}
+
+		// Use the same ID applied to the account name for IAM usernames
+		iamUserUHC := fmt.Sprintf("%s-%s", iamUserNameUHC, currentAccInstanceID)
+		iamUserSRE := fmt.Sprintf("%s-%s", iamUserNameSRE, currentAccInstanceID)
```
dastergon, line 420: this user is already deleted at https://github.com/openshift/aws-account-operator/pull/607/files#diff-998c9da647c44633b8acec24f0b19b1446779dd895f2a2a8f6e4c6a56bc89b05L332 so I guess a rebase is needed

dastergon (approved):
/lgtm 

but you need to rebase to catchup with the latest changes in the controllers

dastergon (approved):
/lgtm

dastergon (approved):
/lgtm

lisa (approved):
/hold
/lgtm
/approve


Remove the hold when you're ready to merge.

PR #585, opened 2021-05-26
Get rid of `operator-sdk generate`

2uasimojo:
PR #580 replaced `operator-sdk generate crds` with a controller-gen command (plus some hacks to make it match up with what operator-sdk did). This commit similarly replaces `operator-sdk generate k8s` with a controller-gen command, reducing our dependency on operator-sdk to just go imports.

While I was in here, I noticed that some framework-generated comments were making their way into the Description of some of the CRD fields and corresponding generated openapi. Nixed those.

Also cleaned up the redundant parts of `generate_crds.sh` which were running go and openapi generators. These were already part of the boilerplate-provided `generate` target as `go-generate` and `openapi-generate`.

Note that regenerating deepcopy code under controller-gen results in a trivial delta to each method whereby the implicit no-op `return` statement is removed.

TODO: Some of the "make compatible with operator-sdk" parts of `generate_crds.sh` are *definitely* not necessary (e.g. nothing cares whether the CRD files are named with `_crd`) and some are *probably* not necessary (I doubt anything breaks if things like `annotations`, `creationTimestamp`, `status`, and list type fields are included). If we can get rid of these, it should be easy enough to collapse this script back into one or two lines in the Makefile.


2uasimojo:
Superseded by #588

PR #588, opened 2021-05-27
Boilerplate update removing `operator-sdk generate`

2uasimojo:
This PR comprises three commits:
- Automated boilerplate update. Importantly, this brings in
  - Ability to securely suppress boilerplate deltas in GitHub via `.gitattributes`. (Suppression is _not_ included in this PR.)
  - The operator image is now built against golang 1.16 (was: 1.13).
  - The OLM catalog images are now built against registry 4.8.0 (was: 4.7.0). This upgrade includes some security fixes.
- Cleanup and code regeneration pursuant to the above.
- Another boilerplate update. Should have zero effect. Just brings in some commits that happened while this PR was waiting for review.

See individual commit messages for details.

PR #589, opened 2021-05-27
Update boilerplate and generate v1 CRDs

2uasimojo:
This updates boilerplate to bring in tooling for generating v1 CRDs. It does that generation, and does some cleanup that's necessary as a result.

The generated CRDs install cleanly, and `oc explain` works.

Reviewers should pay attention to the fact that the account CRD's Status.Conditions and Status.Conditions.Type fields are now required. This was necessary to make the CRD install.


lisa:
/hold

@2uasimojo Do we still need this, how does it relate to #588?

lisa:
/hold

for "do not merge" request

2uasimojo:
> Do we still need this, how does it relate to #588?

This PR is just a PoC/demo of the first step we would need to take to move from CRD v1beta1 to v1, which we'll want to do as soon as v3 is dead. The reason it's stacked on #588 is because it uses that PR's new controller-gen-oriented CRD generator. Otherwise, we would be continuing to tie CRD gen into operator-sdk, which would be (to use another acronym) a PITA.

See [OSD-5869](https://issues.redhat.com/browse/OSD-5869?focusedCommentId=16238129&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16238129) for more.

If you want to close this PR to get it off the board, that's fine. It'll be a little while before we're ready to pull the trigger anyway.

2uasimojo:
I've updated this now that all the pieces are in place and we're starting to move on the CRDv1 effort.

PR #590, opened 2021-05-28
Remove 'auto author assign' workflow

2uasimojo:
This workflow configuration, introduced via #526, was empty, meaning it would always fail ([example](https://github.com/2uasimojo/aws-account-operator/actions/runs/883561188).

We could populate it based on the [docs](https://github.com/marketplace/actions/auto-author-assign); but I contend (and, by this commit, assert) that it's not a useful action. If I've switched off the [setting](https://github.com/settings/notifications) that notifies me by default on activity on a PR I've authored, it's because I don't want to be notified. If I want to make an exception, I'll assign myself. Otherwise, it ain't right for the repository to make that switch for me.


drpaneas (approved):
I don't know what/how happened and the file was empty. In any case, this workflow is not needed. It's useful only if you use GitHub Projects for your scrum/kanban board functionality. e.g. there is some automation happening when somebody assigns the PR, so it gets visible to the relevant GH Project review column. Since we are _not_ using this feature (we use JIRA instead) I guess it doesn't make sense to keep it around.

/lgtm

iamkirkbater:
Works for me.  Was an interesting experiment at the least.

/lgtm

PR #578, opened 2021-05-17
[OSD-6207] Documentation Improvements

dkeohane:
[OSD-6207](https://issues.redhat.com/browse/OSD-6207)
- Added [Go Report Card](https://goreportcard.com/) Badge to README
- Added Go Doc Badge to README
- Added docs README, Glossary, Debugging pages
- Updated Installation docs & moved prerequisites into their own file.
- Grouped all CRs and Controllers by their type (AccountPool, Account etc) instead of their function (CR vs Controllers) as I felt it's easier to read this way

PR #592, opened 2021-06-01
Generate docs

drpaneas:
Populate `https://openshift.github.io/aws-account-operator/` with
HTML static files generated automatically from within `docs/` dir.

- New Makefile options `serve` serves the site locally.
- GH Action for publishing the docs on push (master branch)
- Configure `mkdocs` to use `docs` as the source of markdown files.

Note: Configure GH-Pages for AAO to assume that docs reside in `docs`
      dir in the repo root.

REF: [OSD-6207](https://issues.redhat.com/browse/OSD-6207)

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -1,0 +1,23 @@
+site_name: aws-account-operator
+site_description: 'Documentation for aws-account-operator'
+site_url: https://openshift.github.io/aws-account-operator/
+repo_url: https://github.com/openshift/aws-account-operator
+theme:
+  name: material
+markdown_extensions:
+    - toc:
+        permalink: True
+        separator: "_"
+nav:
+  - Home: README.md
+  - Installation: 1.0-Installation.md
+  - Custom Resources: 2.0-Custom-Resources-and-Controllers.md
+  - Controllers:
+    - Account Pool: 2.1-AccountPool.md
+    - Account: 2.2-Account.md
+    - Account Claim: 2.3-AccountClaim.md
+    - AWSFederatedRole: 2.4-AWSFederatedRole.md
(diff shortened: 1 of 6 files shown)


docs/index.md
```diff
@@ -0,0 +1,58 @@
+# AWS Account Operator
```
iamkirkbater, line 1: Should we point the main README to link to this page instead, so we don't have to update multiple places.

iamkirkbater (changes requested):

mkdocs.yml
```diff
@@ -0,0 +1,16 @@
...
+repo_url: https://github.com/openshift/aws-account-operator
+theme:
+  name: material
+markdown_extensions:
+    - toc:
+        permalink: True
+        separator: "_"
+nav:
+  - Home: index.md
+  - Installation: 1.0-Installation.md
```
iamkirkbater, line 13: Once #578 is merged, can we add the Glossary and Debugging pages here?

iamkirkbater (changes requested):

.github/workflows/publish_docs.yml
```diff
@@ -0,0 +1,20 @@
+name: 'Publish docs'
```
iamkirkbater, line 1: Looks like we won't be able to use Github Actions for this because of org legacy reasons.  However if we really want to explore this option we can do something like integrate with Netlify or something else, but this would have to happen in a broader discussion.  We can still update the docs with the other changes, but let's remove this workflow file.

dkeohane (approved):
Changes look good - but we don't have it anywhere in our documentation on what this does, can you please add a section detailing this :)

PR #594, opened 2021-06-01
Add depandabot configuration

drpaneas:
Whenever there is a new version of the 3rd party dependencies
used by the project (namely whatever is inside go.mod and go.sum)
then let the project update itself without developer's intervention.

Do the same for the APIs, as now, it's a separate isolated pkg.

The configuration makes sure this check runs on weekly basis.

Note: To make this work you need an admin for AAO to install https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/upgrading-from-dependabotcom-to-github-native-dependabot#upgrading-to-github-native-dependabot

REF: [OSD-6207](https://issues.redhat.com/browse/OSD-6207)


lisa:
Closed pending discussion on how to handle things like this in the future.

PR #595, opened 2021-06-01
Ensure region in the AccountClaim is returned from DescribeRegions API call

ArielLima:
The AWS account operator will call `DescribeRegions` to gather a list of regions to initialize when setting up a AWS account.
If a user chooses one of the 4 opt-in regions but hasn't enabled it in AWS the region will not be returned by the DescribeRegions AWS API call and we want to fail.

Card: https://issues.redhat.com/browse/OSD-7417

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -652,6 +652,7 @@ func (r *ReconcileAccount) initializeRegions(reqLogger logr.Logger, currentAcctI
 		reqLogger.Error(err, connErr)
 	}
 
+	// Get a list of regions enabled in the current account
 	regionsEnabledInAccount, err := awsClient.DescribeRegions(&ec2.DescribeRegionsInput{
 		AllRegions: aws.Bool(false),
 	})
@@ -660,22 +661,22 @@ func (r *ReconcileAccount) initializeRegions(reqLogger logr.Logger, currentAcctI
 		return err
 	}
 
-	if !currentAcctInstance.IsBYOC() { // This is a normal account creation
-		reqLogger.Info("Normal account region initialization")
+	// For accounts created by the accountpool we want to ensure we initiate all regions
+	if !currentAcctInstance.IsBYOC() {
 		go r.asyncRegionInit(reqLogger, currentAcctInstance, creds, regionAMIs, castAWSRegionType(regionsEnabledInAccount.Regions))
 		return nil
 	}
 
+	// For non OSD accounts we check the desired region from the accountclaim and ensure that the account has
+	// that region enabled, fail otherwise
 	accountClaim, acctClaimErr := r.getAccountClaim(currentAcctInstance)
 	if acctClaimErr != nil {
 		reqLogger.Info("Accountclaim not found")
 		return acctClaimErr
 	}
-
 	for _, wantedRegion := range accountClaim.Spec.Aws.Regions {
 		found := false
 		for _, enabledRegion := range regionsEnabledInAccount.Regions {
-			reqLogger.Info("Comparing regions", "wantedRegion", wantedRegion, "enabledRegion", enabledRegion)
 			if wantedRegion.Name == *enabledRegion.RegionName {
 				found = true
 			}
@@ -685,7 +686,7 @@ func (r *ReconcileAccount) initializeRegions(reqLogger logr.Logger, currentAcctI
 				r.Client,
 				reqLogger,
 				currentAcctInstance,
-				fmt.Sprintf("AWS region %s is not support supported for AWS account %s", wantedRegion, currentAcctInstance.Name),
+				fmt.Sprintf("AWS region %s is not supported for AWS account %s", wantedRegion, currentAcctInstance.Name),
 				awsv1alpha1.AccountInitializingRegions,
 			)
 			return err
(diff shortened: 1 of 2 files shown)


ArielLima:
@jharrington22

pkg/controller/account/account_controller.go
```diff
@@ -660,9 +660,41 @@ func (r *ReconcileAccount) initializeRegions(reqLogger logr.Logger, currentAcctI
...
+		}
+		if !found {
+			err := utils.SetAccountStatus(
+				r.Client,
+				reqLogger,
+				currentAcctInstance,
+				fmt.Sprintf("AWS region %s is not support supported for AWS account %s", wantedRegion, currentAcctInstance.Name),
+				awsv1alpha1.AccountInitializingRegions,
+			)
+			return err
```
iamkirkbater, line 691: This should be moved out of the for loop entirely.

iamkirkbater (changes requested):
Overall looks good.  One smol nit.

pkg/controller/account/account_controller.go
ArielLima: I don't understand how moving it outside the for loop is beneficial, all it would accomplish is make it less efficient.

iamkirkbater: You're right.  I got the way this works backwards in my brain.  That's what I get for "just popping on and reviewing this real quick" after hours haha.

pkg/controller/account/account_controller.go
```diff
@@ -660,9 +660,41 @@ func (r *ReconcileAccount) initializeRegions(reqLogger logr.Logger, currentAcctI
...
+	accountClaim, acctClaimErr := r.getAccountClaim(currentAcctInstance)
+	if acctClaimErr != nil {
+		reqLogger.Info("Accountclaim not found")
+		return acctClaimErr
+	}
+
+	for _, wantedRegion := range accountClaim.Spec.Aws.Regions {
+		found := false
+		for _, enabledRegion := range regionsEnabledInAccount.Regions {
+			reqLogger.Info("Comparing regions", "wantedRegion", wantedRegion, "enabledRegion", enabledRegion)
```
jharrington22, line 678: Do we want this log message for every region in the list on every CCS claim?

pkg/controller/account/account_controller.go
```diff
@@ -660,9 +660,41 @@ func (r *ReconcileAccount) initializeRegions(reqLogger logr.Logger, currentAcctI
 		return err
 	}
 
+	if !currentAcctInstance.IsBYOC() { // This is a normal account creation
+		reqLogger.Info("Normal account region initialization")
+		go r.asyncRegionInit(reqLogger, currentAcctInstance, creds, regionAMIs, castAWSRegionType(regionsEnabledInAccount.Regions))
+		return nil
+	}
```
jharrington22, line 667: @ArielLima its possible at the moment for a non BYOC user to select this region which should result in a failed cluster install. I think we can remove this so that "normal" AWS account claims fail.

jharrington22 (changes requested):

pkg/controller/account/account_controller.go
iamkirkbater: The problem with this is if we remove it for non-ccs accounts, they're created before the AccountClaim because they're created by the accountpool.  So there would be no "wantedRegion" in those accounts because there's no accountclaim.

Instead we could check in the accountclaim controller to fail if it's a non-ccs account AND it's selecting a non-standard region.

pkg/controller/account/account_controller.go
ArielLima: No this was for debugging, thanks for the catch

pkg/controller/account/account_controller.go
ArielLima: @jharrington22 In addition to what @iamkirkbater just said, I thought the  idea was that “normal” accounts wouldn’t have the option of being one of the 4 opt-in regions so it wouldn’t be an issue. Additionally while that isn’t yet implemented this will still work. Reminder, **it only failed because we set it up to fail** - it won’t fail again since the region was re-enabled.

@iamkirkbater I would much rather improve the comment than check in the accountclaim controller since it is much simpler the way it currently is. As you like to say though, I won’t die on this hill but it would take more time to implement.

pkg/controller/account/account_controller.go
```diff
@@ -660,9 +660,41 @@ func (r *ReconcileAccount) initializeRegions(reqLogger logr.Logger, currentAcctI
...
+			if wantedRegion.Name == *enabledRegion.RegionName {
+				found = true
+			}
+		}
+		if !found {
+			err := utils.SetAccountStatus(
+				r.Client,
+				reqLogger,
+				currentAcctInstance,
+				fmt.Sprintf("AWS region %s is not support supported for AWS account %s", wantedRegion, currentAcctInstance.Name),
```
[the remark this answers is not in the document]
ArielLima: +1

pkg/controller/account/account_controller.go
```diff
@@ -660,9 +660,41 @@ func (r *ReconcileAccount) initializeRegions(reqLogger logr.Logger, currentAcctI
...
+		return nil
+	}
+
+	accountClaim, acctClaimErr := r.getAccountClaim(currentAcctInstance)
+	if acctClaimErr != nil {
+		reqLogger.Info("Accountclaim not found")
+		return acctClaimErr
+	}
+
+	for _, wantedRegion := range accountClaim.Spec.Aws.Regions {
```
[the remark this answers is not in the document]
ArielLima: +1

pkg/controller/account/account_controller.go
iamkirkbater: Yeah, that could definitely be another body of work.  It would be better to just remove the option from being able to be selected by OCM, but I know they have a different deployment schedule than we do.  I'll create a card here to do the check and we can circle back to that.

iamkirkbater:
/lgtm

Thanks!

iamkirkbater (approved):

PR #597, opened 2021-06-03
Adds log message to see if it's a client error

iamkirkbater:
Attempting to debug https://issues.redhat.com/browse/SDA-4272 and I _think_ this is where this is failing but because the accountclaim cannot be updated because of a conflict it's failing to log the clientErr

### DIFF
--- a/pkg/controller/account/byoc.go
+++ b/pkg/controller/account/byoc.go
@@ -83,6 +83,7 @@ func (r *ReconcileAccount) initializeNewCCSAccount(reqLogger logr.Logger, accoun
 		time.Sleep(500 * time.Millisecond)
 	}
 	if clientErr != nil {
+		reqLogger.Error(clientErr, "There was an error creating a Customer AWS Client")
 		err := utils.SetAccountClaimStatus(
 			r.Client,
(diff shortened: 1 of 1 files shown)


pkg/controller/account/byoc.go
```diff
@@ -121,16 +124,20 @@ func (r *ReconcileAccount) initializeNewCCSAccount(reqLogger logr.Logger, accoun
 			awsv1alpha1.InvalidAccountClaim,
 			awsv1alpha1.ClaimStatusError,
 		)
+		if err != nil {
```
jharrington22, line 127: We already know that the error is being raised in this block, some logging in the `Validate()` func would help us understand what about the AccountClaim is invalid.

iamkirkbater: The Validate function is part of the API definition, so I don't want to introduce the logger into that package.  Each `validateErr` that should come back should be distinct enough.

We could remove the log message here now that we're explicitly returning the `validateErr` below instead of returning `err`.  Thoughts?

jharrington22 (approved):

PR #598, opened 2021-06-03
Ensure we validate STS AccountClaims first

jharrington22:
STS AccountClaims are entering the error state because during validate we're checking to see if credentials exist on the AccountClaim which by design wont exist for STS.


iamkirkbater:
/lgtm

Once the CI jobs all finish successfully this should merge :)

PR #599, opened 2021-06-03
Make sure that an account Errors if region initialization fails

ArielLima:
We now only initialize the necessary regions during account creation therefore if the initialization fails we want to fail the account as well.
Card: https://issues.redhat.com/browse/OSD-2982


pkg/controller/account/ec2.go
```diff
@@ -42,22 +48,35 @@ func (r *ReconcileAccount) InitializeSupportedRegions(reqLogger logr.Logger, acc
...
+			regionInitFailed = true
+			// If we fail to initialize the desired region we want to fail the account
+			reqLogger.Error(errors.New(errMsg.ErrorMsg), errMsg.ErrorMsg)
+			err := utils.SetAccountStatus(
+				r.Client,
+				reqLogger,
+				account,
+				fmt.Sprintf("Account %s failed to initialize expected region %s", account.Name, errMsg.Region),
+				awsv1alpha1.AccountInitializingRegions,
+			)
```
lisa, line 71: I am concerned that we'll have a problem if we need to call this method twice (that is, for example, when `i=0` and `i=2`) then we may encounter an error from Kubernetes where we are trying to update an old object.

Some options could be to reload the object, or to collect all errors and update once, outside of the loop to reflect all of the errors encountered.

/hold

lisa (changes requested):

pkg/controller/account/ec2.go
ArielLima: Latest PR fixes this issue

pkg/controller/account/ec2.go
```diff
@@ -42,22 +47,35 @@ func (r *ReconcileAccount) InitializeSupportedRegions(reqLogger logr.Logger, acc
...
 	// Wait for all go routines to send a message or error to notify that the region initialization has finished
 	for i := 0; i < len(regions); i++ {
 		select {
 		case msg := <-ec2Notifications:
 			reqLogger.Info(msg)
 		case errMsg := <-ec2Errors:
-			reqLogger.Error(errors.New(errMsg), errMsg)
+			regionInitFailed = true
+			// If we fail to initialize the desired region we want to fail the account
+			reqLogger.Error(errors.New(errMsg.ErrorMsg), errMsg.ErrorMsg)
```
iamkirkbater, line 64: Can we make this a reqLogger.Info() call, and then add a Error() call where we're setting the account status below?

This would help keep the logs cleaner.  If we had a "Warning" level log, I'd suggest using that, but we don't have that here :/

ArielLima: The region init logic is a little convoluted and I think that knowing exactly what went wrong and where is important. Also, If we don't have this here we might miss out on potential issues i.e. what if for some reason OSD accounts start failing region init but we ready the account anyways, debugging it would be a nightmare (we don't fail those accounts because they could be new - only fail CCS and BYOC on region init failure)? Since we are ignoring those failures I think these logs are even more valuable the way they currently are. In addition, I don't think we create many more AWS accounts nowadays and those are the ones that would needlessly spam the logs. What do you think? @lisa

pkg/controller/account/ec2.go
```diff
@@ -42,22 +47,35 @@ func (r *ReconcileAccount) InitializeSupportedRegions(reqLogger logr.Logger, acc
...
+			regionInitFailedRegion = append(regionInitFailedRegion, errMsg.Region)
 		}
 	}
-
-	reqLogger.Info("Completed initializing all supported regions")
+	// // If an account is BYOC or CCS and region initialization fails we want to fail the account else output success log
+	if regionInitFailed && len(regions) == 1 {
```
dkeohane, line 69: I'm curious here why the `== 1` check? Should we not be checking `len(regionInitFailedRegion) > 0`

iamkirkbater: We only care if region init failed for accounts that are only initializing a single region, as that means that the future install will also fail, but it will be a longer time before it fails.  The only way an account would be initializing a single region would be if it's CCS or ROSA, essentially if it's not being created out of the accountpool and being created by an accountclaim.

In the case of non-ccs accounts, we're initializing on all regions that are available, and those _may_ fail because of the way that AWS makes your "first attempt to spend money in each region fail" for whatever business reason they had for that decision.  Since that's expected behavior we don't want to put the account into an error state for that reason alone.

The argument _could_ be made that we should just say "is not a non-ccs account" instead of `len(regions)` but I'm not sure if that would be any more clear.

This question being asked though means that the comment on line 68 isn't descriptive enough to describe why we're checking for this, and probably should be updated.

pkg/controller/account/ec2.go
```diff
@@ -42,22 +47,35 @@ func (r *ReconcileAccount) InitializeSupportedRegions(reqLogger logr.Logger, acc
...
-			reqLogger.Error(errors.New(errMsg), errMsg)
+			regionInitFailed = true
+			// If we fail to initialize the desired region we want to fail the account
+			reqLogger.Error(errors.New(errMsg.ErrorMsg), errMsg.ErrorMsg)
+			regionInitFailedRegion = append(regionInitFailedRegion, errMsg.Region)
 		}
 	}
-
-	reqLogger.Info("Completed initializing all supported regions")
+	// // If an account is BYOC or CCS and region initialization fails we want to fail the account else output success log
```
lisa, line 68:
```suggestion
	// If an account is BYOC or CCS and region initialization fails we want to fail the account else output success log
```

pkg/controller/account/ec2.go
lisa: I agree with @ArielLima  on this one. If it's too chatty (or not enough) we can adjust it later.

PR #600, opened 2021-06-04
Don't wait for accounts to succeed if they are missing permissions to complete operation in the first place

ArielLima:
This piece of code causes a client without permission to attempt region initialization multiple times even though it is not possible. This PR will simply raise the error up to a higher level where the account can be put into an error state.

This should be merged after https://github.com/openshift/aws-account-operator/pull/599 which is the PR that will actually fail accounts on failure to initialize a given region.

Card: https://issues.redhat.com/browse/OSD-7431

### DIFF
--- a/pkg/controller/account/ec2.go
+++ b/pkg/controller/account/ec2.go
@@ -199,10 +199,10 @@ func (r *ReconcileAccount) BuildAndDestroyEC2Instances(reqLogger logr.Logger, ac
 		time.Sleep(time.Duration(currentWait) * time.Second)
 		var code int
 		code, DescError = DescribeEC2Instances(reqLogger, awsClient, instanceID)
-		if code == 16 {
+		if code == 16 { // 16 represents a successful region initialization
 			reqLogger.Info(fmt.Sprintf("EC2 Instance: %s Running", instanceID))
 			break
-		} else if code == 401 {
+		} else if code == 401 { // 401 represents an UnauthorizedOperation error
 			// Missing permission to perform operations, account needs to fail
 			reqLogger.Error(DescError, fmt.Sprintf("Missing required permissions for account %s", account.Name))
 			return err
--- a/pkg/controller/account/ec2.go
+++ b/pkg/controller/account/ec2.go
@@ -202,6 +202,10 @@ func (r *ReconcileAccount) BuildAndDestroyEC2Instances(reqLogger logr.Logger, ac
 		if code == 16 {
 			reqLogger.Info(fmt.Sprintf("EC2 Instance: %s Running", instanceID))
 			break
+		} else if code == 401 {
+			// Missing permission to perform operations, account needs to fail
+			reqLogger.Error(DescError, fmt.Sprintf("Missing required permissions for account %s", account.Name))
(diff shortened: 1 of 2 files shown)


pkg/controller/account/ec2.go
```diff
@@ -202,6 +202,10 @@ func (r *ReconcileAccount) BuildAndDestroyEC2Instances(reqLogger logr.Logger, ac
 		if code == 16 {
 			reqLogger.Info(fmt.Sprintf("EC2 Instance: %s Running", instanceID))
 			break
+		} else if code == 401 {
+			// Missing permission to perform operations, account needs to fail
+			reqLogger.Error(DescError, fmt.Sprintf("Missing required permissions for account %s", account.Name))
```
lisa, line 207: Do we know at this point what those permissions are (will the error contain them)? It'd be nice to be able to see what is missing here instead of _something_ is wrong.

ArielLima: Hmmm, I don't think so. This is the error we are seeing from was
```
{"level":"error","ts":1622740881.6889632,"logger":"controller_account","msg":"New AWS Error while describing EC2 instance,\n\t\t\t\tAWS Error Code: UnauthorizedOperation,\n\t\t\t\tAWS Error Message: You are not authorized to perform this operation.","Controller":"account","Request.Namespace":"aws-account-operator","Request.Name":"osd-creds-mgmt-h8bmqm","error":"UnauthorizedOperation: You are not authorized to perform this operation.\n\tstatus code: 403, request id: bf78e94b-2f2d-43b9-8fe2-ac6a513fbd7f",
"stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\tpkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128\ngithub.com/openshift/aws-account-operator/pkg/controller/utils.LogAwsError\n\tsrc/github.com/openshift/aws-account-operator/pkg/controller/utils/utils.go:147\ngithub.com/openshift/aws-account-operator/pkg/controller/account.DescribeEC2Instances\n\tsrc/github.com/openshift/aws-account-operator/pkg/controller/account/ec2.go:310\ngithub.com/openshift/aws-account-operator/pkg/controller/account.(*
ReconcileAccount).BuildAndDestroyEC2Instances\n\tsrc/github.com/openshift/aws-account-operator/pkg/controller/account/ec2.go:201\ngithub.com/openshift/aws-account-operator/pkg/controller/account.(*ReconcileAccount).InitializeRegion\n\tsrc/github.com/openshift/aws-account-operator/pkg/controller/account/ec2.go:153"}   
```

dkeohane: For readability, it would be good to put the codes 16 and 401 into consts

ArielLima: I added comments explaining the values. I think declaring these values would make it less readable (someone would have to scroll or even swap files to see their values) and unnecessary since they are only used in one place.

dkeohane (approved):

lisa (approved):
/lgtm
/approve

PR #568, opened 2021-04-30
[OSD-7116] Adding validation to Secretref Name and Namespace

dkeohane:
Fix for [OSD-7116](https://issues.redhat.com/browse/OSD-7116)

- Added validation to ensure AWSCredentials SecretRef was set
- Added Unit tests for Account Types and Account Claim Types
- Removed redundant redeclaration of AwsSecretName variable

PR #601, opened 2021-06-09
Revert validation secret ref

jharrington22:
Do not merge without testing.

This PR reverts 3 PR's the first three are summarized as follows:

#598 was added to fix validation of STS account claims that were failing
#597 was added to provide extra logging only to debug issues with STS account claims failing (resulting in #598)
#568 reverts the PR which introduced the issues in #598 and the current issues with incident in [OHSS-4577](https://issues.redhat.com/browse/OHSS-4577)

#597 and #598 touch the same code that was changed in #568 and is preventing it from reverting cleanly. Both of those PR's are not functionally required if #568 is reverted. Like wise I don't see any core functionality changes that need to be kept in #568 other than the validation which is ok to be without. 

Changes in between the merging of #568 and the last PR #577 are minimal:
#577 Fixes issues with image digests and doesn't touch operator code
#598 and #597 are explained above
#595 is the only PR of significance but should be safe since its only going to fail region initialization. I suspect #568 is causing the current failed regions issues and not this code
#590 Removes a blank workflow auto author assign

If the local tests run for this I'd consider merging it to to stage and immediately validating a cluster install of all three types
- CCS
- OSD
- STS

If that is ok I'd test reuse with the same legal identity and then promote to production.

I don't have my local ENV setup any more and it'll take me a while to get it set so I'd appreciate someone else helping.


iamkirkbater (approved):
I will create a card to follow up in order to re-add the AWS Secret Ref Validation.

I'll run this locally to validate.

rogbas:
Validated the ` resetAccountSpecStatus`  flow to confirm we would be setting `status.state` correctly, it's looking good.

PR #602, opened 2021-06-09
[OSD-7460] check against pre-defined account states 

dastergon:
Instead of checking for a wide range of states in Account Pool controller, I use an already existing func that checks against pre-defined account states and tells us whether an Account operation has failed or not, depending on the state. That way we avoid the case of blocking the controller because a random state doesn't conform with Account operations.

PTAL @iamkirkbater @rogbas @lisa 

See https://issues.redhat.com/browse/OSD-7460

### DIFF
--- a/pkg/controller/accountpool/accountpool_controller.go
+++ b/pkg/controller/accountpool/accountpool_controller.go
@@ -114,7 +114,7 @@ func (r *ReconcileAccountPool) Reconcile(request reconcile.Request) (reconcile.R
 	for _, account := range accountList.Items {
 		// We don't want to count reused accounts here, filter by LegalEntity.ID
 		if !account.Status.Claimed && account.Spec.LegalEntity.ID == "" {
-			if account.Status.State != "Failed" {
+			if !account.IsFailed() {
 				unclaimedAccountCount++
 			}
 		} else {


drpaneas (approved):
/lgtm

pkg/controller/accountpool/accountpool_controller.go
```diff
@@ -114,7 +114,7 @@ func (r *ReconcileAccountPool) Reconcile(request reconcile.Request) (reconcile.R
 	for _, account := range accountList.Items {
 		// We don't want to count reused accounts here, filter by LegalEntity.ID
 		if !account.Status.Claimed && account.Spec.LegalEntity.ID == "" {
-			if account.Status.State != "Failed" {
+			if account.Status.State == string(awsv1alpha1.AccountCreating) || account.Status.State == string(awsv1alpha1.AccountPending) || account.Status.State == string(awsv1alpha1.AccountPendingVerification) || account.Status.State == string(awsv1alpha1.AccountReady) {
```
lisa, line 117: This should be put into a function to check because all of these checks are hard to encapsulate what logic we care about.

lisa:
@dastergon Your PR description reads like code, which is basically the same as looking at the code diff. What would be more helpful to reviewers here is to summarize in human language what we're hoping to achieve and how your PR addresses it. In other words, be a step higher than code changes.

pkg/controller/accountpool/accountpool_controller.go
lisa: Or, even better, to put this into the Account object to ask something like `if account.IsFailed()` or similar so that the `Account` object is responsible for knowing how to answer that question; the controller here doesn't need to know what it means for an `Account` to be in a good state, just that it can get an answer.

lisa (changes requested):

lisa (approved):

PR #603, opened 2021-06-11
[WIP] Reintrodce SetAccountStatus refactor

dkeohane:


pkg/controller/utils/status.go
```diff
@@ -1,29 +1,68 @@
...
+		return nil
+	}
+
+	if !awsv1alpha1.IsValidAccountConditionType(ctype) { // TODO Should we error out here, or just log and return?
+		return nil
+	}
+
+	if !awsv1alpha1.IsValidAccountStateStatus(state) { // TODO Should we error out here, or just log and return?
+		return nil
+	}
```
dkeohane, line 31: Should we error out for each of these? Or just log and return nil?

pkg/apis/aws/v1alpha1/account_types.go
```diff
@@ -290,3 +302,54 @@ func (a *Account) GetCondition(conditionType AccountConditionType) *AccountCondi
 	}
 	return nil
 }
+
```
iamkirkbater, line 305: Maybe we should consider adding an extra file for `account_conditions.go` and contain some of this logic in there?  This file is starting to get a bit long, but I'd also be open to arguments against, like keeping _all_ of the account type defs in the same file too, granted what I'm seeing below doesn't seem to be public.

pkg/controller/account/account_controller.go
```diff
@@ -188,21 +188,28 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
 				reqLogger.Error(stateErr, "failed setting account state", "desiredState", "Failed")
 			}
 			reqLogger.Error(initErr, "failed initializing new CCS account")
 			return result, initErr
 		}
-		utils.SetAccountStatus(currentAcctInstance, AccountCreating, awsv1alpha1.AccountCreating, AccountCreating)
-		updateErr := r.statusUpdate(currentAcctInstance)
-		if updateErr != nil {
+
+		err = utils.SetAccountStatus(
```
iamkirkbater, line 200: Maybe instead of updating this util we should consider making this a bit more encapsulated in the account itself?  So instead of the controller telling the account "Hey, you're failed here's your status" it would say "Hey, account, mark yourself as failed for this reason"

Then we can call it something like this `currentAcctInstance.MarkFailed("Account Creation Failed or whatever we want because this is just literally a string to string field but the actual enum logic is handled in the account def itself")`, save the account and exit.

iamkirkbater: This would also give us the benefit of slowly rolling out that change to different pieces, instead of having to do large sweeping refactors, which I think is what bit us last time.

dkeohane: @iamkirkbater I've thought about this and think it might be best to place the `MarkFailed` style changes within their own PR once this one has been merged - I don't want to wrap even more content into this PR to make reviewing more complicated

iamkirkbater: That's fine with me.

dastergon (changes requested):

pkg/apis/aws/v1alpha1/account_types.go
```diff
@@ -163,7 +175,7 @@ func (a *Account) IsFailed() bool {
 		string(AccountInternalError),
 	}
 	for _, state := range failedStates {
-		if a.Status.State == state {
+		if string(a.Status.State) == state {
```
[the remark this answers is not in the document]
dkeohane: This is actually a great catch!
Locally I just tried changing this to an array of `AccountStateStatus` instead of strings because those are the valid account `a.Status.State` values, but it can't compile, because here we're checking if the state (of type `AccountStateStatus`) is any failed `AccountConditionType` values.

pkg/controller/account/account_controller.go
```diff
@@ -710,22 +752,43 @@ func (r *ReconcileAccount) asyncRegionInit(reqLogger logr.Logger, currentAcctIns
...
+		)
 	} else {
 		if currentAcctInstance.GetCondition(awsv1alpha1.AccountReady) != nil {
 			msg := "Account support case already resolved; Account Ready"
-			utils.SetAccountStatus(currentAcctInstance, msg, awsv1alpha1.AccountReady, AccountReady)
+			err = utils.SetAccountStatus(
+				r.Client,
+				reqLogger,
+				currentAcctInstance,
+				msg,
```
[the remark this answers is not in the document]
dkeohane: Here and below it's used in the logger directly afterwards to I'm just going to leave these 2

pkg/controller/account/account_controller.go
```diff
@@ -710,22 +752,43 @@ func (r *ReconcileAccount) asyncRegionInit(reqLogger logr.Logger, currentAcctIns
...
+			)
 			reqLogger.Info(msg)
 		} else {
 			msg := "Account pending AWS limits verification"
-			utils.SetAccountStatus(currentAcctInstance, msg, awsv1alpha1.AccountPendingVerification, AccountPendingVerification)
+			err = utils.SetAccountStatus(
+				r.Client,
+				reqLogger,
+				currentAcctInstance,
+				msg,
```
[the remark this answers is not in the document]
dkeohane: What I said above

dastergon (approved):
/lgtm

pkg/apis/aws/v1alpha1/account_types.go
dkeohane: added

pkg/controller/account/account_controller.go
```diff
@@ -407,26 +420,34 @@ func (r *ReconcileAccount) handleAccountInitializingRegions(reqLogger logr.Logge
 		// At the time of this writing, that specifically applies to a) non-CCS accounts, and
 		// b) more than 25 minutes between initial Account creation and the Reconcile after
 		// this one.
-		msg := "Recovering from stale region initialization."
+
 		// We're no longer InitializingRegions
-		utils.SetAccountCondition(
+		currentAcctInstance.Status.Conditions = utils.SetAccountCondition(
```
dkeohane, line 425: https://coreos.slack.com/archives/CNJPJSG3Z/p1625737914004200

pkg/apis/aws/v1alpha1/account_types.go
```diff
@@ -151,6 +107,10 @@ func init() {
 
 // Helper Functions
 
+// TODO - Here we're comparing AccountStateStatus to a list of AccountConditionType
```
dkeohane, line 110: Added this to avoid delaying merging this PR even longer - please shout if you disagree

dastergon (approved):
/lgtm 

You also need to rebase

pkg/apis/aws/v1alpha1/account_types.go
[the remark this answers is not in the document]
dastergon: ok, I think we should quit the effort of trying to match AccountConditionType and AccountStateStatus in this PR. It's good that you added a todo, but I guess the PR is already too complex to get another refactor for that. We could create a ticket as a good-first-issue

iamkirkbater:
Is this still on hold or is this ready for final review?

dkeohane:
> Is this still on hold or is this ready for final review?

Let's get it in for final review

georgettica:
@dkeohane seems this is only waiting an approver for the repo
see https://github.com/openshift/release/blob/master/ci-operator/config/openshift/aws-account-operator/OWNERS for the list :)

pkg/apis/aws/v1alpha1/account_types.go
```diff
@@ -299,13 +258,24 @@ func (a *Account) IsOwnedByAccountPool() bool {
 	return false
 }
 
-// GetCondition finds the condition that has the
-// specified condition type in the given list. If none exists, then returns nil.
-func (a *Account) GetCondition(conditionType AccountConditionType) *AccountCondition {
-	for i, condition := range a.Status.Conditions {
-		if condition.Type == conditionType {
-			return &a.Status.Conditions[i]
+const MaximumAccountStateStatus = 6
```
iamkirkbater, line 261: This feels like a strange optimization that I feel will come back to bite us.

pkg/apis/aws/v1alpha1/account_conditions.go
```diff
@@ -0,0 +1,102 @@
...
+func (a *Account) GetCondition(conditionType AccountConditionType) *AccountCondition {
+	for i, condition := range a.Status.Conditions {
+		if condition.Type == conditionType {
+			return &a.Status.Conditions[i]
+		}
+	}
+	return nil
+}
+
+const MaximumAccountConditionTypes = 14
```
iamkirkbater, line 74: This feels like a strange optimization that I feel will come back to bite us.

iamkirkbater (commented):
Nothing here seems too crazy to me.  I still need to test locally, if I can do that.

I do have a few questions about the optimizations made with the length of the list being returned for account states and conditions, I almost feel like that will just cause pain in the future.

pkg/apis/aws/v1alpha1/account_types.go
```diff
@@ -290,3 +302,55 @@ func (a *Account) GetCondition(conditionType AccountConditionType) *AccountCondi
 	}
 	return nil
 }
+
+const MaximumAccountConditionTypes = 14
```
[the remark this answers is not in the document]
dkeohane: Yup it's just for the max length. I've added it to ensure that people just don't hand bomb in a load of values into the array without thinking about it - we've abused these strings in the past, re-declaring const strings in many places,  I'm just trying to find a way to reduce that. This may not be the best solution, so if you have other ideas, I'm 100% interested in hearing! :)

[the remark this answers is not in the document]
iamkirkbater: 🤔   I get where you're coming from now.

What happens if we do something like add more values without increasing this number?  Will it break in a way that's apparent that what we're doing is wrong?  I'm more worried that if we do something like this where we are explicitly setting the length (let's use this one as the example of 14) and then we do something like add two more fields and then existing fields just silently drop out of the list (or the new ones never show up in the list) and we don't catch it right away.

Let me run a test on the go playground.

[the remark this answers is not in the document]
iamkirkbater: Nvm, this looks like it will be a relatively "easy" error to spot and fix: https://go.dev/play/p/nrJ7gQTpwfN

TL;DR: Trying to add 6 elements to a 5-length defined list throws the following build error: `./prog.go:14:3: array index 5 out of bounds [0:5]`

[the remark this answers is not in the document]
dkeohane: The most we could potentially do there is to add a comment to where the values are defined, like here:
https://github.com/openshift/aws-account-operator/pull/603/files#diff-babfe6bb2a6c13e52d68a586e1375736447e90c292c499956cac9e56583a7a37R31

To indicate that the corresponding MAX value should be updated, and the equivalent `getAll*Types()` func needs to have the new value added to its array?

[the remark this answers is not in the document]
iamkirkbater: Yeah, maybe that would be a good compromise here.  We still get the security of just not adding items to the list, however we also have a comment of what the var does and why we chose to do it that way, for those who come after us.

iamkirkbater:
@dkeohane - Do you still think this is worth trying to get in?  If so can you rebase this and bug me about it until I merge it XD

PR #610, opened 2021-06-25
Region init OptInRequired retry

ArielLima:
Card -> https://issues.redhat.com/browse/OSD-7616


fahlmant (approved):
/lgtm

pkg/controller/account/account_controller.go
```diff
@@ -646,10 +638,24 @@ func (r *ReconcileAccount) initializeRegions(reqLogger logr.Logger, currentAcctI
...
+				return nil
+			}
+		}
 		reqLogger.Error(err, "Failed to retrieve list of regions enabled in this account.")
 		return err
 	}
 
+	// We're about to kick off region init in a goroutine. This status makes subsequent
+	// Reconciles ignore the Account (unless it stays in this state for too long).
+	utils.SetAccountStatus(currentAcctInstance, "Initializing Regions", awsv1alpha1.AccountInitializingRegions, AccountInitializingRegions)
```
jharrington22, line 653: What is the reason to move the status update to after the DescribeRegions call? If the DescibeRegions call fails what is the status of the account? Will the operator retry initializing this account?

ArielLima: If the describe regions call fails we want to reconcile and that is why I moved the part where we set the account status to after the describeRegions call. That way if the describeRegions call fails with an "OptInRequired" error we just reconcile and try again. I was considering failing the account if the call failed for any other reasons but decided against it but looking at it again maybe we should?

ArielLima: I didn't want to fail an account that could potentially self resolve I guess? But to answer your question better, yes it just tries to initialize the region again which it wouldn't if we updated the status sooner.

rogbas: IMO moving the status update to after the describe-regions makes sense, the logic should be able to catch up and try again calling initializeRegions.

The potential issue I see is if AWS never recovers from the OptInRequired state for the describe-regions call. It will loop indefinitely and would never mark the account as Failed. But I guess short term we don't have a better option.

rogbas:
LGTM, with the assumption it gets revisited next week to avoid looping indefinitely s described in the comment above.

lisa:
/hold
/approve

Please remove the hold when you're ready to merge.

ArielLima:
cancel /hold

pkg/controller/account/account_controller.go
jharrington22: Makes sense for it to try again thanks.

PR #612, opened 2021-07-02
Add tags to non-ccs accounts on creation

katherinelc321:
This adds tags to non-ccs accounts on creation. This contains their respective shard names. [OSD-7360](https://issues.redhat.com/browse/OSD-7360)

### DIFF
--- a/pkg/awsclient/client.go
+++ b/pkg/awsclient/client.go
@@ -16,9 +16,9 @@ package awsclient
 import (
 	"context"
 	"fmt"
-	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/service/route53/route53iface"
 	"github.com/aws/aws-sdk-go/service/s3/s3manager"
 	"github.com/aws/aws-sdk-go/service/servicequotas"
@@ -103,6 +103,7 @@ type Client interface {
 	CreateOrganizationalUnit(*organizations.CreateOrganizationalUnitInput) (*organizations.CreateOrganizationalUnitOutput, error)
 	ListOrganizationalUnitsForParent(*organizations.ListOrganizationalUnitsForParentInput) (*organizations.ListOrganizationalUnitsForParentOutput, error)
 	ListChildren(*organizations.ListChildrenInput) (*organizations.ListChildrenOutput, error)
+	TagResource(*organizations.TagResourceInput) (*organizations.TagResourceOutput, error)
 
 	//sts
 	AssumeRole(*sts.AssumeRoleInput) (*sts.AssumeRoleOutput, error)
@@ -316,6 +317,10 @@ func (c *awsClient) ListChildren(input *organizations.ListChildrenInput) (*organ
 	return c.orgClient.ListChildren(input)
 }
 
+func (c *awsClient) TagResource(input *organizations.TagResourceInput) (*organizations.TagResourceOutput, error) {
+	return c.orgClient.TagResource(input)
+}
+
 func (c *awsClient) AssumeRole(input *sts.AssumeRoleInput) (*sts.AssumeRoleOutput, error) {
 	return c.stsClient.AssumeRole(input)
 }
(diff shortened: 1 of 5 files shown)


pkg/controller/account/account_controller.go
```diff
@@ -528,6 +528,33 @@ func (r *ReconcileAccount) nonCCSAssignAccountID(reqLogger logr.Logger, currentA
...
+		return err
+	}
+
+	hiveName, ok := configMap.Data["shard-name"]
+
+	if ok {
+		// Add tag
+		inputTag := &organizations.TagResourceInput{
+			ResourceId: aws.String(awsAccountID),
+			Tags: []*organizations.Tag{
```
iamkirkbater, line 543: Do we also need to add the `claimed: true` tag here as well?

pkg/controller/account/account_controller.go
```diff
@@ -528,6 +528,33 @@ func (r *ReconcileAccount) nonCCSAssignAccountID(reqLogger logr.Logger, currentA
...
+			},
+		}
+
+		_, err = awsSetupClient.TagResource(inputTag)
+		if err != nil {
+			return err
+		}
+
+	}
+
```
iamkirkbater, line 557: This entire block of code would probably be better suited as a function on the reconcile object that is called here.

So what I would do would be to take this block and add it to a new function called `r.tagAccount`, so here would be replaced with something like:

```
err = r.tagAccount(awsSetupClient)
if err != nil {
  reqLogger.Info("Unable to tag aws account.", "account", currentAcctInstance.Name, "AWSAccountID", awsAccountID, "Error", error.Error())
}
```

And then tagAccount would be defined further down and would look like this:

```
func (r *ReconcileAccount) tagAccount(awsSetupClient awsClient.Client) error {
  [code above]
}
```

pkg/controller/account/account_controller.go
```diff
@@ -528,6 +528,33 @@ func (r *ReconcileAccount) nonCCSAssignAccountID(reqLogger logr.Logger, currentA
...
 
+	configMap, err := controllerutils.GetOperatorConfigMap(r.Client)
+	if err != nil {
+		reqLogger.Error(err, "failed getting shard name from configmap data")
+		return err
+	}
+
+	hiveName, ok := configMap.Data["shard-name"]
+
+	if ok {
```
iamkirkbater, line 539: With the `if/else` thing we talked about when we paired up last week, let's try to fail fast here and check for 

```
if !ok { 
  return fmt.Errorf("shard-name key not available in configmap")
} 
... rest of inputTag operation
```

It's easier to read the code when your errors are caught earlier and returned before the "happy-path" code is executed.

pkg/awsclient/client.go
```diff
@@ -16,9 +16,10 @@ package awsclient
 import (
 	"context"
 	"fmt"
-	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws/endpoints"
```
iamkirkbater, line 21: Since this got moved let's just group this with the other aws sdk bits below.

iamkirkbater (changes requested):
Overall looks pretty good, a few small changes for code readability.

dkeohane (commented):
Read through the PR and all looks good once Kirk's requested changes are added.
Would like to see a unit test created for Kirk's proposed tagAccount function as well.
Great job :t-rex:

pkg/controller/account/account_controller.go
katherinelc321: we decided that non-ccs accounts will only have their hive name tagged

pkg/controller/account/account_controller.go
```diff
@@ -91,6 +91,17 @@ func newReconciler(mgr manager.Manager) reconcile.Reconciler {
 		scheme:           mgr.GetScheme(),
 		awsClientBuilder: &awsclient.Builder{},
 	}
+
+	configMap, err := controllerutils.GetOperatorConfigMap(reconciler.Client)
+	if err != nil {
+		log.Error(err, "failed getting shard name from configmap data")
```
dkeohane, line 97: here it's a failure in retrieving the configmap and the shard name is the next section

pkg/controller/account/account_controller.go
```diff
@@ -536,9 +548,35 @@ func (r *ReconcileAccount) nonCCSAssignAccountID(reqLogger logr.Logger, currentA
...
+		Tags: []*organizations.Tag{
+			{
+				Key:   aws.String("owner"),
+				Value: aws.String(r.shardName),
+			},
+		},
+	}
+
+	_, err := awsSetupClient.TagResource(inputTag)
+	if err != nil {
```
dkeohane, line 573: Here you can just return err directly I'm guessing? As it should return nil if it's not an err, so no need to check

katherinelc321: it's telling me i need 2 variables so i'm the "_" in there

drpaneas (approved):
LGTM

iamkirkbater (approved):
/lgtm

PR #616, opened 2021-07-29
Iterate through all IAM users

jharrington22:
We only iterate through 3 pages of IAM users, we should iterate through every IAM user.


dastergon (approved):
/lgtm

PR #617, opened 2021-08-06
[OSD-7720] Fix AWS Federated Account Access from leaking Policies

dastergon:
In the AWS Federated Account Access controller, there was a code for deleting policies after all operations are done, but due to a wrong check between the policy name and the actual attached policy name, the condition was never met and the deletion path was not executed and that lead to leaking policies. The issue was that the attached policy in AWS has a uid label as a suffix appended to the actual policy name. 

Example format:
- BillingReadOnlyAccess-48zkrs
- CustomerAdministratorAccess-6pvv82

But the policy name in the if the condition was just `BillingReadOnlyAccess`. This PR fixes the policy leaking.

PTAL 
See: https://issues.redhat.com/browse/OSD-7720

### DIFF
--- a/pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
+++ b/pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
@@ -668,7 +668,9 @@ func (r *ReconcileAWSFederatedAccountAccess) cleanFederatedRoles(reqLogger logr.
 					return err
 				}
 			}
-			if *policy.PolicyName == federatedRoleCR.Spec.AWSCustomPolicy.Name {
+
+			awsCustomPolicyname := federatedRoleCR.Spec.AWSCustomPolicy.Name + "-" + uidLabel
+			if *policy.PolicyName == awsCustomPolicyname {
 				_, err = awsClient.DeletePolicy(&iam.DeletePolicyInput{PolicyArn: policy.PolicyArn})
 				if err != nil {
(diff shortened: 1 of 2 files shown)


fahlmant (approved):
/lgtm

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -668,7 +668,9 @@ func (r *ReconcileAWSFederatedAccountAccess) cleanFederatedRoles(reqLogger logr.
 					return err
 				}
 			}
-			if *policy.PolicyName == federatedRoleCR.Spec.AWSCustomPolicy.Name {
+
+			awsCustomPolicyname := federatedRoleCR.Spec.AWSCustomPolicy.Name + "-" + uidLabel
+			if *policy.PolicyName == awsCustomPolicyname {
```
dkeohane, line 673: I did a quick check in this file alone for other instances of `federatedRoleCR.Spec.AWSCustomPolicy.Name` and found one in `deleteNonAttachedCustomPolicy` - can we ensure that the naming concatenation doesn't need to be done there as well? 

Would potentially be worth breaking `federatedRoleCR.Spec.AWSCustomPolicy.Name + "-" + uidLabel` into its own function?

lisa:
/approve
/lgtm

Dan will write up a card to do the improvements he mentioned in his comment.

PR #619, opened 2021-08-11
Revert "Update boilerplate and generate v1 CRDs"

iamkirkbater:
Reverts openshift/aws-account-operator#589 to see if the updated CRs are the cause for the CCS account creation issue.


fahlmant (approved):
/lgtm

PR #621, opened 2021-08-12
Rename variable and add uidLabel check

katherinelc321:
Comments left on https://issues.redhat.com/browse/OSD-8047


pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -670,8 +670,8 @@ func (r *ReconcileAWSFederatedAccountAccess) cleanFederatedRoles(reqLogger logr.
 			}
 
 			awsCustomPolicyname := federatedRoleCR.Spec.AWSCustomPolicy.Name + "-" + uidLabel
-			if *policy.PolicyName == awsCustomPolicyname {
-				_, err = awsClient.DeletePolicy(&iam.DeletePolicyInput{PolicyArn: policy.PolicyArn})
+			if *attachedPolicy.PolicyName == awsCustomPolicyname {
```
dkeohane, line 673: Could still potentially refactor this to something like:
```
err := compareAndDeletePolicy(crPolicyName, policyName, policyArn)
if err != nil {
    return err
}
```
Where `crPolicyName` can have the label applied to it in the func, and all error handling can be managed within the func. This new function can then be unit tested individually.

dkeohane (commented):
We can sync up and discuss this a bit when you're available

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -768,3 +753,37 @@ func hasLabel(awsFederatedAccountAccess *awsv1alpha1.AWSFederatedAccountAccess,
 	}
 	return false
 }
+
+func checkAndDeletePolicy(reqLogger logr.Logger, awsClient awsclient.Client, uidLabel string, crPolicyName string, policyName *string, policyArn *string) error {
+
+	var awsCustomPolicyname string
+	awsCustomPolicyname = checkUidLabel(awsCustomPolicyname, crPolicyName, uidLabel)
```
dkeohane, line 760: With your current implementation of `checkUidLabel` you'd always be checking `if "" != crPolicyName+"-"+uidLabel`, so this check is irrelevant.  

What we discussed in our sync is checking f the value in `crPolicyName` already contains the `"-"+uidLabel` postfix. Ideally this function should check `crPolicyName` to see if it already contains the uidLabel, if so, return `crPolicyName` as is, if not, return `crPolicyName+"-"+uidLabel`. If any of the above isn't clear we can sync up again to discuss :)

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -768,3 +753,37 @@ func hasLabel(awsFederatedAccountAccess *awsv1alpha1.AWSFederatedAccountAccess,
...
+			} else {
+				reqLogger.Error(err, "Other error while trying to detach policies")
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func checkUidLabel(awsCustomPolicyname string, crPolicyName string, uidLabel string) string {
```
dkeohane, line 784: Missing unit test

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+func setupDefaultMocks(t *testing.T, localObjects []runtime.Object) *mocks {
+	mocks := &mocks{
+		mockCtrl: gomock.NewController(t),
+	}
+
+	return mocks
+}
+
+func TestCheckAndDeletePolicy(t *testing.T) {
```
dkeohane, line 29: We're currently only testing the happy path here, would be good to test a failed DeletePolicy call

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -709,15 +694,25 @@ func (r *ReconcileAWSFederatedAccountAccess) cleanFederatedRoles(reqLogger logr.
...
 	return nil
 }
 
-func (r *ReconcileAWSFederatedAccountAccess) deleteNonAttachedCustomPolicy(reqLogger logr.Logger, awsClient awsclient.Client, federatedRoleCR *awsv1alpha1.AWSFederatedRole) error {
+func (r *ReconcileAWSFederatedAccountAccess) deleteNonAttachedCustomPolicy(reqLogger logr.Logger, awsClient awsclient.Client, currentFAA *awsv1alpha1.AWSFederatedAccountAccess, federatedRoleCR *awsv1alpha1.AWSFederatedRole) error {
+
+	// Get the UID
+	uidLabel, ok := currentFAA.Labels[awsv1alpha1.UIDLabel]
+	if !ok {
+		if currentFAA.Status.State != awsv1alpha1.AWSFederatedAccountStateReady {
```
dkeohane, line 710: This block is new - checking for the `Ready` state, is this required?

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller_test.go
dkeohane: Still relevant

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -768,3 +750,45 @@ func hasLabel(awsFederatedAccountAccess *awsv1alpha1.AWSFederatedAccountAccess,
...
+
+	var awsCustomPolicyname string
+	awsCustomPolicyname = checkUidLabel(awsCustomPolicyname, crPolicyName, uidLabel)
+
+	if *policyName == awsCustomPolicyname {
+		_, err := awsClient.DeletePolicy(&iam.DeletePolicyInput{PolicyArn: policyArn})
+		if err != nil {
+			if aerr, ok := err.(awserr.Error); ok {
+				switch aerr.Code() {
+				case iam.ErrCodeNoSuchEntityException:
```
dkeohane, line 764: Are these new cases not a change in functionality? Here we'll be logging at info level and then step out of the block to `return nil`. You'll need to add `return err` to each case

dkeohane (changes requested):

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
dkeohane: Why are these errors being logged at info level instead of error level?

PR #624, opened 2021-09-04
Improve on IAM test coverage

dkeohane:
Created a PR of some tests to expand golang testing knowledge and our test coverage for iam.go

### DIFF
--- a/pkg/controller/account/iam.go
+++ b/pkg/controller/account/iam.go
@@ -45,7 +45,10 @@ type StatementEntry struct {
 }
 
 var (
-	DefaultDelay = 3 * time.Second
+	defaultDelay      = 3 * time.Second
+	defaultSleepDelay = 500 * time.Millisecond
+	// testSleepModifier is set to 0 in tests so that tests don't sleep and cause a slowdown
+	testSleepModifier int = 1
 )
 
 // CreateSecret creates a secret for placing IAM Credentials
@@ -97,7 +100,7 @@ func getSTSCredentials(
 	assumeRoleOutput := &sts.AssumeRoleOutput{}
 	var err error
 	for i := 0; i < 100; i++ {
-		time.Sleep(500 * time.Millisecond)
+		time.Sleep(defaultSleepDelay)
 		assumeRoleOutput, err = client.AssumeRole(&assumeRoleInput)
 		if err == nil {
 			break
@@ -146,7 +149,7 @@ func listAccessKeys(client awsclient.Client, iamUser *iam.User) (*iam.ListAccess
 
 	// Default is 1/10 of a second, but any retries we need to make should be delayed a few seconds
 	// This also defaults to an exponential backoff, so we only need to try ~5 times, default is 10
-	retry.DefaultDelay = DefaultDelay
+	retry.DefaultDelay = defaultDelay
 	retry.DefaultAttempts = uint(5)
 	err = retry.Do(
 		func() (err error) {
@@ -167,7 +170,7 @@ func deleteAccessKey(client awsclient.Client, accessKeyID *string, username *str
 
 	// Default is 1/10 of a second, but any retries we need to make should be delayed a few seconds
 	// This also defaults to an exponential backoff, so we only need to try ~5 times, default is 10
-	retry.DefaultDelay = DefaultDelay
+	retry.DefaultDelay = defaultDelay
 	retry.DefaultAttempts = uint(5)
 	err = retry.Do(
 		func() (err error) {
@@ -246,7 +249,7 @@ func CreateIAMUser(reqLogger logr.Logger, client awsclient.Client, userName stri
 					utils.LogAwsError(reqLogger, "CreateIAMUser: Unexpected AWS Error during creation of IAM user", nil, err)
 					return &iam.CreateUserOutput{}, err
 				}
-				time.Sleep(time.Duration(time.Duration(attempt*5) * time.Second))
+				time.Sleep(time.Duration(time.Duration(attempt*5*testSleepModifier) * time.Second))
 			} else {
 				return &iam.CreateUserOutput{}, err
 			}
@@ -266,7 +269,7 @@ func AttachAdminUserPolicy(client awsclient.Client, iamUser *iam.User) (*iam.Att
 	attachPolicyOutput := &iam.AttachUserPolicyOutput{}
 	var err error
 	for i := 0; i < 100; i++ {
-		time.Sleep(500 * time.Millisecond)
+		time.Sleep(defaultSleepDelay)
 		attachPolicyOutput, err = client.AttachUserPolicy(&iam.AttachUserPolicyInput{
 			UserName:  iamUser.UserName,
 			PolicyArn: aws.String("arn:aws:iam::aws:policy/AdministratorAccess"),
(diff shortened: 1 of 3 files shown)


pkg/controller/account/iam.go
```diff
@@ -434,10 +436,14 @@ func deleteIAMUser(reqLogger logr.Logger, awsClient awsclient.Client, user *iam.
 	return nil
 }
 
+var (
+	listIAMUsers = awsclient.ListIAMUsers
+)
```
ArielLima, line 441: Just curious, what is the advantage of it this way vs `var listIAMUsers = awsclient.ListIAMUsers` for a single variable?

pkg/controller/account/iam.go
```diff
@@ -434,10 +436,14 @@ func deleteIAMUser(reqLogger logr.Logger, awsClient awsclient.Client, user *iam.
...
+
 func deleteIAMUsers(reqLogger logr.Logger, awsClient awsclient.Client, accountCR *awsv1alpha1.Account) error {
 	reqLogger.Info("Cleaning up IAM users")
 
-	users, err := awsclient.ListIAMUsers(reqLogger, awsClient)
+	users, err := listIAMUsers(reqLogger, awsClient)
```
ArielLima, line 446: In my opinion
```
users, err := awsclient.ListIAMUsers(reqLogger, awsClient)
```
is much better, cleaner and easier to understand than:
```
var listIAMUsers = awsclient.ListIAMUsers
users, err := listIAMUsers(reqLogger, awsClient)
```
I prefer leaving it the way it was unless there is a reason we are changing it that I don't understand

pkg/controller/account/iam_test.go
```diff
@@ -93,7 +96,33 @@ func TestGetSTSCredentials(t *testing.T) {
...
+	old := DefaultSleepDelay
+	DefaultSleepDelay = 0 * time.Millisecond
+	creds, err = getSTSCredentials(
+		nullLogger,
+		mockAWSClient,
+		"",
+		"",
+		"",
+	)
+	DefaultSleepDelay = old
```
ArielLima, line 123: `DefaultSleepDelay` looks like a local variable, what happens if you remove this line?

pkg/controller/account/iam_test.go
```diff
@@ -256,35 +277,112 @@ func TestDeleteAllAccessKeys(t *testing.T) {
...
+				mocks := setupDefaultMocks(t, []runtime.Object{})
+				test.setupAWSMock(mocks.mockAWSClient.EXPECT())
+				// This is necessary for the mocks to report failures like methods not being called an expected number of times.
+				// after mocks is defined
+				defer mocks.mockCtrl.Finish()
+
+				old := DefaultScale
+				DefaultScale = 0
+				createUserOutput, err := CreateIAMUser(nullLogger, mocks.mockAWSClient, usernameStr)
+				DefaultScale = old
```
ArielLima, line 379: Similar to `DefaultSleepDelay`, is it local? Could we get away with just re-declaring it once?

pkg/controller/account/iam_test.go
```diff
@@ -302,6 +401,21 @@ func TestAttachAdminUserPolicy(t *testing.T) {
...
+	mockAWSClient.EXPECT().AttachUserPolicy(gomock.Any()).Return(
+		&iam.AttachUserPolicyOutput{},
+		expectedError, // no error
+	).Times(100)
+
+	// retries took long, need to patch it to 0 delay.
+	old := DefaultSleepDelay
+	DefaultSleepDelay = 0 * time.Millisecond
+	attachAdminUserPolicy, err = AttachAdminUserPolicy(mockAWSClient, &user)
+	DefaultSleepDelay = old
```
ArielLima, line 416: local variable?

pkg/controller/account/iam.go
dkeohane: Apologies, I should have left a comment above explaining my reasoning for this - I'll add it in the morning. 
I'm doing this to monkey patch the function `awsclient.ListIAMUsers` being called with my own defined 'mocked' func found [here](https://github.com/openshift/aws-account-operator/pull/624/files#diff-2d6c1df4c89463ea67c8a31c387b6ae3c450fe8c265fd99f446b73d3236a309eR579). 

The way I would normally mock out these func calls is by using some form of `mockedawsclient.EXPECT().funcCall` - unfortunately this wasn't cleanly available to me here. The [awsclient.ListIAMUsers](https://github.com/openshift/aws-account-operator/blob/master/pkg/awsclient/iam.go#L44)  isn't defined in the client interface, therefore it isn't mocked out. What is mocked out though is the function which ListIAMUsers calls, [ListUsersPages](https://github.com/openshift/aws-account-operator/blob/master/pkg/awsclient/iam.go#L49-L53). `ListUsersPages`'s second parameter is a function pointer which modifies a locally defined slice - meaning I'd need to do a major refactor just to mock this function out differently.

pkg/controller/account/iam_test.go
dkeohane: These are not local variables, they are package scoped and defined [here](https://github.com/openshift/aws-account-operator/pull/624/files#diff-06877cbab8eaefc13650b26e9c0b26fbb438a30862eb2b9666e0224421f51129R48-R50). I didn't think of another way of injecting a change to the retry time outside of doing something like this - if you have something cleaner in mind please shout :)

pkg/controller/account/iam.go
```diff
@@ -45,7 +45,9 @@ type StatementEntry struct {
 }
 
 var (
-	DefaultDelay = 3 * time.Second
+	DefaultDelay          = 3 * time.Second
+	DefaultSleepDelay     = 500 * time.Millisecond
+	DefaultScale      int = 1
```
lisa, line 50: `DefaultScale` is used for what (I can see where it is used in code, but what semantic meaning)?

dkeohane: Given the way it's used in the code:
```
time.Sleep(time.Duration(time.Duration(attempt*5*DefaultScale) * time.Second))
```
`attempt` is an int which increases per loop attempt, so I couldn't just set the overall sleep time for this - so I decided to add a multiplier scale factor which I would set to 0 for tests, while remaining at 1 for the normal use it shouldn't affect the previous implementation

dkeohane: - I'm happy to rename if you have a different idea
- I can also leave a comment to further describe it

pkg/controller/account/iam_test.go
ArielLima: I still don't think you need to reset these values back, I'm going to pull these changes down and test my theory just in case.

dkeohane: For isolated testing purposes, probably not, but I generally feel it's good practice for our tests to not carry state from one to the next.

pkg/controller/account/iam.go
dkeohane: Renamed `DefaultScale` to `TestSleepModifier`

pkg/controller/account/iam_test.go
dkeohane: Moved these to `init()` func to avoid resetting values per-test

pkg/controller/account/iam.go
```diff
@@ -45,7 +45,9 @@ type StatementEntry struct {
 }
 
 var (
-	DefaultDelay = 3 * time.Second
+	DefaultDelay          = 3 * time.Second
+	DefaultSleepDelay     = 500 * time.Millisecond
+	TestSleepModifier int = 1
```
lisa, line 50:
```suggestion
	// TestSleepModifier is set to 0 in tests so that tests don't sleep and cause a slowdown
	TestSleepModifier int = 1
```

lisa (changes requested):

lisa (approved):
/hold
/lgtm
/approve


unhold when you want to merge.

lisa (commented):
/lgtm
/approve

PR #625, opened 2021-09-08
AWS Account Operator should check for the condition status of creating rather than the creation timestamp

ArielLima:
Card -> https://issues.redhat.com/browse/OSD-7693

Replaced `IsOlderThan` method  for `CreationConditionOlderThan`, updated tests.


dkeohane (approved):
/lgtm

PR #627, opened 2021-09-10
Add unit tests to awsfederatedaccountaccess_controller.go

katherinelc321:


pkg/controller/awsfederatedaccountaccess/cov.html
```diff
@@ -0,0 +1,873 @@
+
```
iamkirkbater, line 1: I don't think we need this file.

iamkirkbater (changes requested):
Great start! I'd love to see a few more of those error cases, like the `else` where the UID can't be found tested as well.

I'm specifically looking at lines 351 and 400.  There's something wrong with those lines (I know you're not introducing the problem, but since we're writing tests this is a great opportunity to fix those issues since they're relatively small).  Also let me know if you can't find the problem :)

The biggest change I think that needs to be made though is just removing the `cov.html` file.  If _your_ workflow is to make the cov.html files generated in the directory of the code, then please look into adding `cov.html` to your global gitignore file.

pkg/controller/awsfederatedaccountaccess/cov.html
katherinelc321: committed by mistake heh

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller_test.go
```diff
@@ -0,0 +1,351 @@
...
+					AWSCustomPolicy: newTestAwsCustomPolicyBuilder().awsCustomPol},
+			}
+
+			afaa := awsv1alpha1.AWSFederatedAccountAccess{
+				ObjectMeta: v1.ObjectMeta{
+					Labels: test.uidLabel,
+				}}
+
+			createPolicyOutput, err := r.createIAMPolicy(mockAWSClient, afr, afaa)
+			if test.expectedErr != nil {
```
dkeohane, line 114: This if statement isn't needed

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller_test.go
```diff
@@ -0,0 +1,351 @@
...
+			afaa := awsv1alpha1.AWSFederatedAccountAccess{
+				ObjectMeta: v1.ObjectMeta{
+					Labels: test.uidLabel},
+				Spec: awsv1alpha1.AWSFederatedAccountAccessSpec{
+					ExternalCustomerAWSIAMARN: "",
+				},
+			}
+
+			createRoleOutput, err := r.createIAMRole(mockAWSClient, afr, afaa)
+			if test.expectedErr != nil {
```
dkeohane, line 184: This if statement isn't needed

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller_test.go
```diff
@@ -0,0 +1,351 @@
...
+		awsOutputGci,
+		nil,
+	)
+
+	mockAWSClient.EXPECT().DeletePolicy(gomock.Any()).Return(
+		&iam.DeletePolicyOutput{},
+		nil,
+	)
+
+	mockAWSClient.EXPECT().CreatePolicy(gomock.Any()).Return(
```
dkeohane, line 214: Can you move this CreatePolicy EXPECT up above the DeletePolicy to better represent the order in which the calls are made?

dkeohane: Please change the gomock.Any() in the DeletePolicy and CreatePolicy to better represent what we're testing - shouldn't be to big of a change as it's just passing in afr and afaa which you've defined below

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller_test.go
```diff
@@ -0,0 +1,351 @@
...
+	createRoleOutput := &iam.CreateRoleOutput{
+		Role: &iam.Role{},
+	}
+
+	mockAWSClient.EXPECT().DeleteRole(gomock.Any()).Return(
+		&iam.DeleteRoleOutput{},
+		nil,
+	)
+
+	mockAWSClient.EXPECT().CreateRole(gomock.Any()).Return(
```
dkeohane, line 259: Similar comments here for the Role as for the Policy

dkeohane (commented):
Awesome start - thanks for having the initiative to start writing these :+1:

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller_test.go
```diff
@@ -0,0 +1,422 @@
...
+
+	afaa := awsv1alpha1.AWSFederatedAccountAccess{
+		ObjectMeta: v1.ObjectMeta{
+			Labels: map[string]string{"uid": ""}},
+	}
+
+	uidLabel := afaa.Labels["uid"]
+	policyName := afr.Spec.AWSCustomPolicy.Name + "-" + uidLabel
+
+	type awsStatement struct {
```
dkeohane, line 210: You shouldn't need to redefine this? There's one available from utils package with the same definition from what I can see

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller_test.go
```diff
@@ -0,0 +1,422 @@
...
+	policyDoc := struct {
+		Version   string
+		Statement []awsStatement
+	}{
+		Version:   "2012-10-17",
+		Statement: statements,
+	}
+
+	jsonPolicyDoc, err := json.Marshal(&policyDoc)
+	assert.Nil(t, err)
```
dkeohane, line 233: Don't think this needs to be asserted either as it's static data, we're not randomly seeding anything so there's no speculation for data variance

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller_test.go
```diff
@@ -0,0 +1,422 @@
...
+
+	mocks := setupDefaultMocks(t, []runtime.Object{})
+
+	mockAWSClient := mock.NewMockClient(mocks.mockCtrl)
+
+	defer mocks.mockCtrl.Finish()
+
+	afr := awsv1alpha1.AWSFederatedRole{
+		Spec: awsv1alpha1.AWSFederatedRoleSpec{
+			AWSCustomPolicy: newTestAwsCustomPolicyBuilder().awsCustomPol},
```
dkeohane, line 199: For my own pedantry
```suggestion
			AWSCustomPolicy: newTestAwsCustomPolicyBuilder().awsCustomPol,
		},
```

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller_test.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	afaa := awsv1alpha1.AWSFederatedAccountAccess{
+		ObjectMeta: v1.ObjectMeta{
+			Labels: map[string]string{"uid": ""}},
+	}
+
+	type awsStatement struct {
```
dkeohane, line 277: Similar to my comment around the Policy, this still needs to be addressed

PR #628, opened 2021-09-13
Adds a check for AccessDenied on setup client init

iamkirkbater:
This check on the setup client for Access Denied allows us to clean out
accounts that are denied access by the account owner or are otherwise
inaccessable to our operator

Fullfils [OSD-8400](https://issues.redhat.com/browse/OSD-8400)

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -148,7 +148,6 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 	}
 
 	// We expect this secret to exist in the same namespace Account CR's are created
-	// We build the input here but the client later because we handle the error differently
 	awsSetupClient, err := r.awsClientBuilder.GetClient(controllerName, r.Client, awsclient.NewAwsClientInput{
 		SecretName: utils.AwsSecretName,
 		NameSpace:  awsv1alpha1.AccountCrNamespace,
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -148,14 +148,11 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 	}
 
 	// We expect this secret to exist in the same namespace Account CR's are created
-	awsSetupClient, err := r.awsClientBuilder.GetClient(controllerName, r.Client, awsclient.NewAwsClientInput{
+	// We build the input here but the client later because we handle the error differently
+	setupClientInput := awsclient.NewAwsClientInput{
 		SecretName: utils.AwsSecretName,
 		NameSpace:  awsv1alpha1.AccountCrNamespace,
 		AwsRegion:  "us-east-1",
-	})
-	if err != nil {
-		reqLogger.Error(err, "failed building operator AWS client")
-		return reconcile.Result{}, err
 	}
 
 	if currentAcctInstance.IsPendingDeletion() {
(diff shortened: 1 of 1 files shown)


pkg/controller/account/account_controller.go
```diff
@@ -169,6 +166,25 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 			return reconcile.Result{}, err
 		}
 
+		awsSetupClient, err := r.awsClientBuilder.GetClient(controllerName, r.Client, setupClientInput)
+		if err != nil {
+			reqLogger.Error(err, "failed building operator AWS client")
+			if aerr, ok := err.(awserr.Error); ok {
+				switch aerr.Code() {
+				// If it's AccessDenied we want to just delete the finalizer and continue as we assume
```
dkeohane, line 174: Might be useful to put some logging here to help improve visibility on this - we can remove at a later date if found unnecessary

iamkirkbater: the code on line 171 is already logging the error, the `switch` code just allows us to define if it's specifically an AWS error or not.  The "print" version of the AWS error ends up looking something like this in the logs:

`AccessDenied: User: arn:aws:iam::REDACTED:user/REDACTED is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::REDACTED:role/REDACTED\n\tstatus code: 403, request id: 14258866-1552-11ec-82a8-0242ac130003`

Is there something specific you're looking for in the logs you're thinking about that's missing?

dkeohane: Nope that seems fair - just wanted to double check to see if we wanted to have it highlighted in our logs when a customer removes our creds and therefore we can't clean up correctly after ourselves

pkg/controller/account/account_controller.go
```diff
@@ -148,14 +148,11 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 	}
 
 	// We expect this secret to exist in the same namespace Account CR's are created
-	awsSetupClient, err := r.awsClientBuilder.GetClient(controllerName, r.Client, awsclient.NewAwsClientInput{
+	// We build the input here but the client later because we handle the error differently
```
ArielLima, line 151: I think that this comment is misleading. My first thought was, "we handle the errors exactly the same...", but we do build the client in two places so IMO I would just leave it like:
```
// Defining the AWS client builder input
```

pkg/controller/account/account_controller.go
```diff
@@ -203,6 +219,12 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 		return reconcile.Result{}, nil
 	}
 
+	awsSetupClient, err := r.awsClientBuilder.GetClient(controllerName, r.Client, setupClientInput)
```
ArielLima, line 222: I think a comment above here could be nice so we don't "forget" why we only build the client so late. just an idea.

ArielLima (changes requested):
Just a couple of thoughts on comments, looks good though.

pkg/controller/account/account_controller.go
lisa, line 222: This line appears to reuse the `awsSetupClient` identifier which is used above in your PR on line 169. Is this a scoping issue where the earlier use on line 169 is meant to also be used here at this scope?

pkg/controller/account/account_controller.go
```diff
@@ -169,6 +166,25 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
+				switch aerr.Code() {
+				// If it's AccessDenied we want to just delete the finalizer and continue as we assume
+				// the credentials have been deleted by the customer
+				case "AccessDenied":
+					err = r.removeFinalizer(currentAcctInstance, awsv1alpha1.AccountFinalizer)
+					if err != nil {
+						reqLogger.Error(err, "failed removing account finalizer")
+						return reconcile.Result{}, err
+					}
+					return reconcile.Result{}, nil
```
lisa, line 182: In here (or before this code block), I think we'll want to add a check to only remove the finalizer for non-CCS accounts for safety.

For some accounting, purposes, we may want to add a Prometheus counter and/or log message of when this code path is reached, broken down by Account type (CCS vs non-CCS) so we can look back and see some data about the ratio. (At this time I think a log message will suffice as part of a "break/fix" solution and we can revisit the Prometheus metric later.)

lisa (changes requested):

pkg/controller/account/account_controller.go
iamkirkbater: If the awsSetupClient on line 169 is initialized there should be no reason the code execution for that same reconcile iteration should get here.  I can rename the client, though, to prevent this confusion.

PR #629, opened 2021-09-14
We don't need to output the error because it is expected behavior

ArielLima:
Card: https://issues.redhat.com/browse/OSD-7820


dkeohane (approved):
Let's get this merged, it's spamming me when testing locally :joy:

PR #630, opened 2021-09-15
refactor SetAccountCondition func

katherinelc321:
https://issues.redhat.com/browse/OSD-7694


katherinelc321:
Condition.Status is already checked inside the shouldUpdateCondition() func. If we check it again it's gonna override the updateConditionCheck value we pass in

PR #631, opened 2021-09-15
Enable transit gateway support OSD-8418

jharrington22:
Enabling transit gateway for non-CCS via AWS infrastructure access.


jharrington22:
Closing this PR as transit gateway support will not be enabled in non-CCS OSD, just transit gateway attachments.

PR #632, opened 2021-09-17
[OSD-8416] Added ManagedOpenShiftSupportRole to accounts 

dkeohane:
Ref [OSD-8416](https://issues.redhat.com/browse/OSD-8416):
Added ManagedOpenShiftSupportRole to accounts and update SupportRoleARN in accountClaim with Role ARN

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -12,7 +12,6 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/ec2"
-	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/aws/aws-sdk-go/service/organizations"
 	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/go-logr/logr"
@@ -1190,16 +1189,11 @@ func (r *ReconcileAccount) handleCreateAdminAccessRole(
 	currentAcctInstance *awsv1alpha1.Account,
 	awsSetupClient awsclient.Client) (awsclient.Client, *sts.AssumeRoleOutput, error) {
 
+	var err error
 	var awsAssumedRoleClient awsclient.Client
 	var creds *sts.AssumeRoleOutput
 	currentAccInstanceID := currentAcctInstance.Labels[awsv1alpha1.IAMUserIDLabel]
 	roleToAssume := getAssumeRole(currentAcctInstance)
-
-	SREAccessARN, err := r.GetSREAccessARN(reqLogger)
-	if err != nil {
-		reqLogger.Error(err, "An error was encountered retrieving the SRE Access ARN")
-		return nil, nil, err
-	}
 
 	// Build the tags required to create the Admin Access Role
 	tags := awsclient.AWSTags.BuildTags(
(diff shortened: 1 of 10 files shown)


pkg/controller/account/account_controller.go
```diff
@@ -1256,6 +1256,21 @@ func (r *ReconcileAccount) handleCreateAdminAccessRole(
...
+			awsSetupClient,
+			ccsClient,
+			adminAccessArn,
+			currentAccInstanceID,
+			tags,
+			SREAccessARN,
+		)
+
+		if err != nil {
+			reqLogger.Error(err, "createManagedOpenShiftSupportRole err", roleID)
```
dastergon, line 1270: can you add something more descriptive here?

pkg/controller/account/account_controller.go
```diff
@@ -1283,7 +1298,85 @@ func (r *ReconcileAccount) handleCreateAdminAccessRole(
...
+			awsSetupClient,
+			awsAssumedRoleClient,
+			adminAccessArn,
+			currentAccInstanceID,
+			tags,
+			SREAccessARN,
+		)
+
+		if err != nil {
+			reqLogger.Error(err, "createManagedOpenShiftSupportRole err", roleID)
```
dastergon, line 1313: ditto

pkg/controller/accountclaim/accountclaim_controller.go
```diff
@@ -209,7 +209,17 @@ func (r *ReconcileAccountClaim) Reconcile(request reconcile.Request) (reconcile.
 	if accountClaim.Spec.AccountLink == "" {
 		setAccountLinkOnAccountClaim(reqLogger, unclaimedAccount, accountClaim)
 		return reconcile.Result{}, r.specUpdate(reqLogger, accountClaim)
+	}
 
+	if !accountClaim.Spec.ManualSTSMode {
```
dastergon, line 214: This code block seems a duplicate from the code block below, would it make sense to move them to a func if they're doing the same thing?

dastergon (changes requested):

pkg/controller/account/account_controller.go
```diff
@@ -1256,6 +1256,21 @@ func (r *ReconcileAccount) handleCreateAdminAccessRole(
...
 		}
 
+		_, err = createManagedOpenShiftSupportRole(
+			reqLogger,
+			awsSetupClient,
+			ccsClient,
+			adminAccessArn,
+			currentAccInstanceID,
+			tags,
+			SREAccessARN,
```
iamkirkbater, line 1266: In the case of these roles, we won't be able to use the SREAccessARN the same way we're using it now, we'll need to add the STS access ARN to the configmap and use that.

Essentially, right now we're using a role on the payer account called like CCS-Access or something, which is different than the one we're using for STS jump-role access.  We need to get the STS access arn and add that to App Interface as a parameter, and update the deployed configmap to have that ARN, and then use the same STS jump role here instead of the SREAccessARN.

This is probably confusing as hell, so please reach out if it doesn't make sense.

iamkirkbater (changes requested):

pkg/controller/account/account_controller.go
dkeohane: Updated

pkg/controller/account/account_controller.go
dkeohane: Updated

pkg/controller/accountclaim/accountclaim_controller.go
dkeohane: Updated

iamkirkbater (changes requested):
Small little changes to help onboarding.

QQ: Is it supposed to be STS_JUMP_ARN or STS_JUMP_ROLE in the makefile?  My direnv is set to use `STS_JUMP_ROLE` but I'm not sure what everyone else is using.

iamkirkbater (approved):
/lgtm

PR #633, opened 2021-09-20
Update create account script to stop nuking existing account

dkeohane:
Encountered an issue today when calling `make create-accountclaim` when an account CR with a name of `osd-creds-mgmt-osd-staging-1` already exists, the call nukes the existing account and replaces it with a templated account CR. 

Fixed by adding a check to see if account already exists before applying new account CR.

### DIFF
--- a/test/integration/api/create_account.sh
+++ b/test/integration/api/create_account.sh
@@ -1,7 +1,17 @@
 #!/bin/bash
 
 # Load Environment vars
-source hack/scripts/test_envs 
+source hack/scripts/test_envs
+
+ACCOUNT_CR_EXISTS=$(oc get account ${OSD_STAGING_1_ACCOUNT_CR_NAME_OSD} -n ${NAMESPACE})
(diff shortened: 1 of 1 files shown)


dastergon (approved):
/lgtm

iamkirkbater:
I'm confused what the actual problem is here.  The `oc apply` should be idempotent and should update the existing CR if it already exists or create a new one if it doesn't.  What was the error you were seeing?

dkeohane:
The issue I was encountering was when I had the existing Account CR `osd-creds-mgmt-osd-staging-1` with a valid `iamUserSecret` set. The `oc apply` appears to wipe the `iamUserSecret` and therefore when the new accountclaim see there's a 'valid' account, it claims fine but fails when attempting `createIAMSecret` in the `accountclaim_controller.go`

dkeohane:
Confirmed it's part of spec, removing hold

PR #634, opened 2021-09-21
Update timestamp

katherinelc321:
https://issues.redhat.com/browse/OSD-7694


katherinelc321:
The SetAccountCondition func only returns conditions when we first set them but not when we update them. SetAccountStatus in status.go expects something returned in the other side of awsAccount.Status.Conditions

pkg/controller/utils/conditions.go
```diff
@@ -163,6 +163,15 @@ func SetAccountCondition(
...
 		existingCondition.LastProbeTime = now
+		conditions = append(conditions,
+			awsv1alpha1.AccountCondition{
+				Type:               conditionType,
+				Status:             existingCondition.Status,
+				Reason:             existingCondition.Reason,
+				Message:            existingCondition.Message,
+				LastTransitionTime: existingCondition.LastTransitionTime,
+				LastProbeTime:      existingCondition.LastProbeTime,
+			})
```
ArielLima, line 174: I would remove line 165 and change line 173 to:
```
LastProbeTime:      now,
```

dastergon: @katherinelc321 can you update the code with Ari's recommendation?

katherinelc321:
> Were you able to confirm this fixes the issue through testing locally?

Not through CRC and unit tests weren't working because of the way [this](https://github.com/openshift/aws-account-operator/blob/a11fd5997552273b9ee726b1277d98bed143d84a/pkg/apis/aws/v1alpha1/account_types.go#L107) line was written but that's getting refactored in PR 603 and that should make my test work

PR #636, opened 2021-09-24
Fixes issue with BYOCAdminAccess prefix

iamkirkbater:
Fixes an issue where BYOCAdminAccess prefix is being added a second time
to the role name when attempting to delete the existing role.


lisa (approved):
/lgtm
/approve

let's be sure to fix `DeleteBYOCAdminAccessRole` so it's less foot-gun-y.

PR #637, opened 2021-09-27
Adds some new metrics to the accountpool status

iamkirkbater:
Adds additional status fields and calculations for the same in order to start reporting better metrics on accountPool availability.

Satisfies [OSD-7754](https://issues.redhat.com/browse/OSD-7754)

### DIFF
--- a/pkg/apis/aws/v1alpha1/accountpool_types.go
+++ b/pkg/apis/aws/v1alpha1/accountpool_types.go
@@ -16,9 +16,22 @@ type AccountPoolSpec struct {
 // AccountPoolStatus defines the observed state of AccountPool
 // +k8s:openapi-gen=true
 type AccountPoolStatus struct {
-	PoolSize          int `json:"poolSize"`
+	PoolSize int `json:"poolSize"`
+
+	// UnclaimedAccounts is an approximate value representing the amount of non-failed accounts
 	UnclaimedAccounts int `json:"unclaimedAccounts"`
-	ClaimedAccounts   int `json:"claimedAccounts"`
+
+	// ClaimedAccounts is an approximate value representing the amount of accounts that are currently claimed
+	ClaimedAccounts int `json:"claimedAccounts"`
+
+	// AvailableAccounts denotes accounts that HAVE NEVER BEEN CLAIMED, so NOT reused, and are READY to be claimed.  This differs from the UnclaimedAccounts, who similarly HAVE NEVER BEEN CLAIMED, but include ALL non-FAILED states
+	AvailableAccounts int `json:"availableAccounts"`
+
+	// AccountsProgressing shows the approximate value of the number of accounts that are in the creation workflow (Creating, PendingVerification, InitializingRegions)
+	AccountsProgressing int `json:"accountsProgressing"`
+
+	// AWSLimitDelta shows the approximate difference between the number of AWS accounts currently created and the limit. This should be the same across all hive shards in an environment
+	AWSLimitDelta int `json:"awsLimitDelta"`
 }
 
 // +genclient
@@ -30,6 +43,9 @@ type AccountPoolStatus struct {
 // +kubebuilder:printcolumn:name="Pool Size",type="integer",JSONPath=".status.poolSize",description="Desired pool size"
 // +kubebuilder:printcolumn:name="Unclaimed Accounts",type="integer",JSONPath=".status.unclaimedAccounts",description="Number of unclaimed accounts"
 // +kubebuilder:printcolumn:name="Claimed Accounts",type="integer",JSONPath=".status.claimedAccounts",description="Number of claimed accounts"
+// +kubebuilder:printcolumn:name="Available Accounts",type="integer",JSONPath=".status.availableAccounts",description="Number of ready accounts"
+// +kubebuilder:printcolumn:name="Accounts Progressing",type="integer",JSONPath=".status.accountsProgressing",description="Number of accounts progressing towards ready"
+// +kubebuilder:printcolumn:name="AWS Limit Delta",type="integer",JSONPath=".status.awsLimitDelta",description="Difference between accounts created and soft limit"
 // +kubebuilder:resource:path=accountpools,scope=Namespaced
 type AccountPool struct {
 	metav1.TypeMeta   `json:",inline"`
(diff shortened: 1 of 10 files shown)


pkg/controller/accountpool/accountpool_controller.go
```diff
@@ -155,15 +139,54 @@ func (r *ReconcileAccountPool) Reconcile(request reconcile.Request) (reconcile.R
...
+		if account.HasNeverBeenClaimed() {
+			if !account.IsFailed() {
+				unclaimedAccountCount++
+			}
+		} else {
+			claimedAccountCount++
+		}
+
+		// count available accounts
+		if account.HasNeverBeenClaimed() && account.IsReady() {
```
dkeohane, line 171: Could be moved into the if block above

pkg/controller/accountpool/accountpool_controller.go
```diff
@@ -155,15 +139,54 @@ func (r *ReconcileAccountPool) Reconcile(request reconcile.Request) (reconcile.R
...
-	if currentAccountPool.Status.PoolSize != currentAccountPool.Spec.PoolSize {
-		return true
-	} else if currentAccountPool.Status.UnclaimedAccounts != unclaimedAccounts {
-		return true
-	} else if currentAccountPool.Status.ClaimedAccounts != claimedAccounts {
-		return true
-	} else {
-		return false
+// Calculates the unclaimedAccountCount and Claimed Account Counts
+func (r *ReconcileAccountPool) calculateAccountStatus() (awsv1alpha1.AccountPoolStatus, error) {
```
dkeohane, line 143: minor: Could be named `calculateAccountPoolStatus`

dkeohane (commented):
Left 2 super minor comments.
Nice work, will make figuring what state the account pool is in much easier :+1:

pkg/controller/accountpool/accountpool_controller.go
iamkirkbater: Yeah, I was kind of torn on doing that, or keeping it explicitly separate to keep the logic easier to follow if it changes more in the future.

I was even thinking about getting rid of the "else" block above as well, and just keeping each individual calculation separate just for an easier glance as to what's doing what instead of overloading the if statements and making it harder to follow.

Thoughts?

dkeohane: Removing the else might actually make it more difficult to read, `if !account.HasNeverBeenClaimed() {` has not never been claimed == has been claimed before, translation has to happen in your head. 

Personally I don't see a problem with moving this block up to into other once it's paired with a comment to clarify any potential confusion - but I'm happy with either solution tbh :)

iamkirkbater: I ended up just creating a simpler method `HasBeenClaimedAtLeastOnce()` which describes this a bit better.

pkg/apis/aws/v1alpha1/accountpool_types.go
```diff
@@ -16,9 +16,22 @@ type AccountPoolSpec struct {
...
+	PoolSize int `json:"poolSize"`
+
+	// UnclaimedAccounts is an approximate value representing the amount of non-failed accounts
 	UnclaimedAccounts int `json:"unclaimedAccounts"`
-	ClaimedAccounts   int `json:"claimedAccounts"`
+
+	// ClaimedAccounts is an approximate value representing the amount of accounts that are currently claimed
+	ClaimedAccounts int `json:"claimedAccounts"`
+
+	// AvailableAccounts denotes accounts that HAVE NOT BEEN CLAIMED and are NOT reused and are READY to be claimed.  (This differs from the UnclaimedAccounts above which is all non-failed accounts, this number is only READY accounts)
```
dkeohane, line 27: I found the comparison to UnclaimedAccounts accounts was missing the part where they too have never been claimed. Here's an attempt at improving it, but might need more cleanup - wording is hard.
```suggestion
	// AvailableAccounts denotes accounts that HAVE NEVER BEEN CLAIMED, so NOT reused, and are READY to be claimed.  This differs from the UnclaimedAccounts, who similarly HAVE NEVER BEEN CLAIMED, but include ALL non-FAILED states
```

pkg/controller/account/account_controller.go
```diff
@@ -570,14 +570,23 @@ func (r *ReconcileAccount) accountSpecUpdate(reqLogger logr.Logger, account *aws
...
+	switch utils.DetectDevMode {
+	case utils.DevModeProduction:
+		var err error
+		awsAccountID, err = r.BuildAccount(reqLogger, awsSetupClient, currentAcctInstance)
+		if err != nil {
+			return err
+		}
+	default:
+		log.Info("Running in development mode, skipping account creation")
+		awsAccountID = "123456789012"
```
dkeohane, line 584: Nice, should shave some time off testing :)

pkg/totalaccountwatcher/totalaccountwatcher.go
```diff
@@ -24,20 +24,26 @@ import (
 var ErrAwsAccountLimitExceeded = errors.New("AccountLimitExceeded")
 
 // TotalAccountWatcher global var for TotalAccountWatcher
-var TotalAccountWatcher = &totalAccountWatcher{}
+var TotalAccountWatcher = &AccountWatcher{}
 
 var log = logf.Log.WithName("aws-account-operator")
 
-type totalAccountWatcher struct {
+type AccountWatcherIface interface {
```
dkeohane, line 31: Nice :+1:

dkeohane (approved):
Nice changes! :+1: 
Left some minor ideas / suggestions. Please add an update to the [AccountPool documentation](https://github.com/openshift/aws-account-operator/blob/master/docs/3.1-AccountPool.md) as well to reflect these changes, either as part of this PR or cut a ticket.

pkg/controller/account/account_controller.go
iamkirkbater: This doesn't actually skip anything in our testing as our integration tests have account IDs, so they never actually get to this point, but it allows us to kind of test the accountpool functionality without creating an account. (Each account created by the account pool _will_ fail, but we can test that the account CR is created, etc)

dastergon (approved):
/lgtm 

excellent 👍🏻

PR #640, opened 2021-10-02
[OSD-7432] Added STS validation

dkeohane:
Ref: https://issues.redhat.com/browse/OSD-7432

- Renamed `test-sts-accountclaim` to `test-sts`
- Added some validation for STS mode account and account claim


drpaneas:
For what's worth I just finished testing this and /lgtm -- sorry didn't notice it was merged already.

PR #641, opened 2021-10-05
Added unit tests for AccountClaim finalizer

dkeohane:


pkg/controller/accountclaim/accountclaim_finalizer_test.go
```diff
@@ -0,0 +1,179 @@
...
+
+			err := r.addFinalizer(nullLogger, accountClaim)
+			Expect(err).NotTo(HaveOccurred())
+		})
+		It("should not add finalizer as account claim doesn't exist", func() {
+			// Objects to track in the fake client.
+			objs := []runtime.Object{}
+			r.client = fake.NewFakeClient(objs...)
+
+			err := r.addFinalizer(nullLogger, accountClaim)
```
dkeohane, line 92: Would have liked to add validation on the accountClaim finalizer list itself, to ensure the size actually changes. But given these funcs manipulate the accClaim pointer and even if they encounter an error while making a request to k8s, that accClaim object will still have the modified finalizer list. Could potentially work around this by pulling the accClaim object in the controller, but seems excessive - but open to discussion... if any of this made sense

iamkirkbater: If you run a separate "get" on NewFakeClient at the end of the test, that should only return what's "saved" in k8s I believe.

So the test would be something like the pseudocode below:

```
client := fake.NewFakeClient(objs...)
r.client = client

r.addFinalizer(nullLogger, accountClaim)
Expect(err).To(HaveOccurred())

ac := &awsv1alpha1.AccountClaim{}
testErr := client.Get(context.TODO(), ac)

Expect(TestErr).NotTo(HaveOccurred())
Expect(ac.ObjectMeta.Finalizers).To(BeEmpty())
```

pkg/controller/accountclaim/accountclaim_finalizer_test.go
```diff
@@ -0,0 +1,242 @@
+package accountclaim
```
NautiluX, line 1: IMHO unit tests shouldn't cover package-internal functions but only exported functions (the public interface). This can be easily enforced by putting the tests in a different package than the functions we test, like `accountclaim_test`

PR #644, opened 2021-10-18
Fixes Accountpool Metric

iamkirkbater:
I did a bad and need to fix the accountpool size metric.
Re-Satisfies [OSD-7754](https://issues.redhat.com/browse/OSD-7754)


dastergon (approved):
/lgtm

PR #645, opened 2021-10-18
allow network-mgmt to dowload the VPN configs

yithian:
The permissions required to download a VPN config from AWS appear to have changed, which is preventing customers from downloading it themselves using network-mgmt. This adds the missing permissions.

This is generating toil, as documented in OHSS-6964, [OHSS-7075](https://issues.redhat.com/browse/OHSS-7075) and [OHSS-7264](https://issues.redhat.com/browse/OHSS-7264) where we had to manually provide the configs for the customers.

This should resolve OSD-8599.

### DIFF
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -149,8 +149,6 @@ objects:
             - "ec2:ModifyVpcPeeringConnectionOptions"
             - "ec2:RejectVpcPeeringConnection"
             - "ec2:DisableVgwRoutePropagation"
-            - "ec2:GetVpnConnectionDeviceTypes"
-            - "ec2:GetVpnConnectionDeviceSampleConfiguration"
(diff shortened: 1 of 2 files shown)


hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -149,6 +149,8 @@ objects:
             - "ec2:ModifyVpcPeeringConnectionOptions"
             - "ec2:RejectVpcPeeringConnection"
             - "ec2:DisableVgwRoutePropagation"
+            - "ec2:GetVpnConnectionDeviceTypes"
```
fahlmant, line 152: Why are these being added in two places?

jharrington22:
@fahlmant @yithian lets make sure these are added to the STS support roles in MCC and the ROSA CLI

jharrington22:
@fahlmant that makes sense.

/approve

fahlmant (approved):
/lgtm

PR #652, opened 2021-11-08
OSD-7210: Move AccountClaim Validation into the AccountClaim controller

mrWinston:
This PR makes sure, that the accountclaim validation happens before any account is even created by moving the validation code to `pkg/controller/accountclaim/accountclaim_controller.go`.

Additional unit tests are also added to cover the basics of creating a byoc accountclaim.

### DIFF
--- a/pkg/controller/accountclaim/accountclaim_controller_test.go
+++ b/pkg/controller/accountclaim/accountclaim_controller_test.go
@@ -31,6 +31,7 @@ var _ = Describe("AccountClaim", func() {
 		accountClaim *awsv1alpha1.AccountClaim
 		r            *ReconcileAccountClaim
 		ctrl         *gomock.Controller
+		req          reconcile.Request
 	)
 
 	err := apis.AddToScheme(scheme.Scheme)
@@ -60,6 +61,12 @@ var _ = Describe("AccountClaim", func() {
 				},
 			},
 		}
+		req = reconcile.Request{
+			NamespacedName: types.NamespacedName{
+				Name:      name,
+				Namespace: namespace,
+			},
+		}
 
 		// Create the reconciler with a mocking AWS client IBuilder.
 		r = &ReconcileAccountClaim{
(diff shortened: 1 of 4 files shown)


hack/templates/aws.managed.openshift.io_v1alpha1_ccs_accountclaim_cr_invalid.tmpl
```diff
@@ -0,0 +1,30 @@
+apiVersion: v1
```
dkeohane, line 1: Care to leave a comment at the top of this file indicating how it's an invalid accountclaim?

mrWinston: I would just go ahead and remove this file and associated make targets again, as i don't really see the need for this kind of integration test now that we have the unit test properly set up.

dkeohane: Sounds good :+1:

pkg/controller/accountclaim/accountclaim_controller_test.go
```diff
@@ -142,6 +136,57 @@ var _ = Describe("AccountClaim", func() {
...
+		When("Accountclaim is BYOC", func() {
+
+			BeforeEach(func() {
+				accountClaim.SetFinalizers(append(accountClaim.GetFinalizers(), accountClaimFinalizer))
+				accountClaim.Spec.BYOC = true
+				accountClaim.Spec.AccountLink = ""
+			})
+
+			It("should fail validation", func() {
+				// this should cause Validate() to fail
```
drpaneas, line 149: Did I get that right?

```suggestion
				// fail validation if BYOC is not associated with an account
```

drpaneas (approved):
/lgtm it seems legit change, relocating this codeblock to the appropriate place.

pkg/controller/accountclaim/accountclaim_controller_test.go
mrWinston: Yep, that's better!

NautiluX (commented):
/lgtm

PR #653, opened 2021-11-09
add finalizer to non-sts cr

katherinelc321:
https://issues.redhat.com/browse/OSD-8550

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -158,6 +158,14 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 		return reconcile.Result{}, err
 	}
 
+	// Add finalizer to non-sts account cr
+	if !currentAcctInstance.Spec.ManualSTSMode {
+		err := r.addFinalizer(reqLogger, currentAcctInstance)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+
 	if currentAcctInstance.IsPendingDeletion() {
 		if currentAcctInstance.Spec.ManualSTSMode {
 			// if the account is STS, we don't need to do any additional cleanup aside from
--- a/pkg/controller/account/account_finalizer.go
+++ b/pkg/controller/account/account_finalizer.go
@@ -3,9 +3,27 @@ package account
 import (
 	"context"
 
+	"github.com/go-logr/logr"
 	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
 	"github.com/openshift/aws-account-operator/pkg/controller/utils"
+	controllerutils "github.com/openshift/aws-account-operator/pkg/controller/utils"
 )
+
+func (r *ReconcileAccount) addFinalizer(reqLogger logr.Logger, account *awsv1alpha1.Account) error {
+
+	if !controllerutils.Contains(account.GetFinalizers(), awsv1alpha1.AccountFinalizer) {
+		reqLogger.Info("Adding Finalizer for the Account")
+		account.SetFinalizers(append(account.GetFinalizers(), awsv1alpha1.AccountFinalizer))
+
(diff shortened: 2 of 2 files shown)


pkg/controller/account/account_controller.go
```diff
@@ -72,7 +72,8 @@ const (
 	adminAccessArn = "arn:aws:iam::aws:policy/AdministratorAccess"
 	iamUserNameUHC = "osdManagedAdmin"
 
-	controllerName = "account"
+	controllerName   = "account"
+	accountFinalizer = "finalizer.aws.managed.openshift.io"
```
dkeohane, line 76: this already exists, so no need to duplicate -> use `awsv1alpha1.AccountFinalizer`

dkeohane (changes requested):

pkg/controller/account/account_finalizer.go
```diff
@@ -3,10 +3,25 @@ package account
 import (
 	"context"
 
+	"github.com/go-logr/logr"
 	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
 	"github.com/openshift/aws-account-operator/pkg/controller/utils"
 )
 
+// Function to add finalizer
```
NautiluX, line 11: Probably it would be better to either have a comment that explains a bit more context about the function or not have a comment at all? This one doesn't tell me much more than reading the function name :)

pkg/controller/account/account_finalizer.go
```diff
@@ -3,10 +3,25 @@ package account
...
+func (r *ReconcileAccount) addFinalizer(reqLogger logr.Logger, account *awsv1alpha1.Account) error {
+	reqLogger.Info("Adding Finalizer for the Account")
+	account.SetFinalizers(append(account.GetFinalizers(), awsv1alpha1.AccountFinalizer))
+
+	// Update CR
+	err := r.Client.Update(context.TODO(), account)
+	if err != nil {
+		reqLogger.Error(err, "Failed to update Account with finalizer")
```
NautiluX, line 19: Can we actually output the error that happened here? Or probably just wrap the error with this string to add context and return it?

pkg/controller/account/account_controller.go
```diff
@@ -158,6 +158,18 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 		return reconcile.Result{}, err
 	}
 
+	// Add finalizer to non-sts account cr
+	if !currentAcctInstance.Spec.ManualSTSMode {
+		if !controllerutils.Contains(currentAcctInstance.GetFinalizers(), awsv1alpha1.AccountFinalizer) {
```
NautiluX, line 163: could this check actually happen in the addFinalizer function itself? So we could just call it here, and it would ensure the finalizer is only added if not there already?

pkg/controller/account/account_controller.go
```diff
@@ -158,6 +158,15 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 		return reconcile.Result{}, err
 	}
 
+	// Add finalizer to non-sts account cr
+	if !currentAcctInstance.Spec.ManualSTSMode {
+		err := r.addFinalizer(reqLogger, currentAcctInstance)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		return reconcile.Result{}, nil
```
iamkirkbater, line 167: Wouldn't this just loop at this point forever for non-sts clusters?

katherinelc321: no? return reconcile.Result{}, nil should break the loop

iamkirkbater: But it would still never get to the next part of the code where it starts building the account, right?  Because it would always hit that return statement?

katherinelc321: aah gonna delete that last return

PR #654, opened 2021-11-23
OSD-7582: Deploy Account Shredder Service Account in Account Operator

mrWinston:
This PR makes sure that the service account and accompanying roles for the aws-account-*shredder* are deployed alongside the aws-account-*operator*.

Jira: [OSD-7582](https://issues.redhat.com/browse/OSD-7582)

### DIFF
--- a/deploy/aas_role.yaml
+++ b/deploy/aas_role.yaml
@@ -1,0 +1,34 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  namespace: aws-account-operator
+  name: aws-account-shredder
+rules:
+- apiGroups:
+  - route.openshift.io
+  resources:
+  - routes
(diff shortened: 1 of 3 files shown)


fahlmant (commented):
Will this deploy on prod? The shredder is specifically not deployed in prod so the SA + RBAC should also not get deployed there.

fahlmant:
Can we deploy this via MCC instead? That way we can deploy it to select clusters, and package it separately from AAO

dastergon:
Since we're going to shred the shredder, I guess we could close this PR, WDYT @NautiluX?

mrWinston:
> Since we're going to shred the shredder, I guess we could close this PR, WDYT @NautiluX?

I also don't see a point in keeping this PR open. I'm just gonna go ahead and close it.

PR #656, opened 2021-12-02
[WIP] Add fedramp support

sam-nguyen7:
This PR covers: [OSD-9242](https://issues.redhat.com/browse/OSD-9242) (Add FEDRAMP Deploy Parameter) and [OSD-6798](https://issues.redhat.com/browse/OSD-6798) (GovCloud Support) 

- Add a new param called FEDRAMP to OLM template and operator configmap.

- Condition all operator controllers to check when FEDRAMP = true and if true then change all aws clients region from us-east-1 to us-gov-east-1 

- If FEDRAMP = true the IAM ARN changes from arn:aws:iam:: to arn:aws-us-gov:iam::


pkg/controller/account/account_controller.go
```diff
@@ -643,9 +643,23 @@ func (r *ReconcileAccount) assumeRole(
...
+	if ok {
+		fmt.Sprintln("Account is govcloud")
+		roleArn = fmt.Sprintf("arn:aws-us-gov:iam::%s:role/%s", currentAcctInstance.Spec.AwsAccountID, roleToAssume)
+		// log.Error(err, "fedramp key not available in configmap")
+	}
+
 	// The role ARN made up of the account number and the role which is the default role name
 	// created in child accounts
-	var roleArn = fmt.Sprintf("arn:aws:iam::%s:role/%s", currentAcctInstance.Spec.AwsAccountID, roleToAssume)
+	roleArn = fmt.Sprintf("arn:aws:iam::%s:role/%s", currentAcctInstance.Spec.AwsAccountID, roleToAssume)
```
iamkirkbater, line 662: This would need to be in an "else" block otherwise it will overwrite the fedramp role 100% of the time.

pkg/controller/accountclaim/accountclaim_controller.go
```diff
@@ -263,6 +263,16 @@ func (r *ReconcileAccountClaim) Reconcile(request reconcile.Request) (reconcile.
...
+		// get configMap and check if account is fedramp
+		configMap, err := controllerutils.GetOperatorConfigMap(r.client)
+		if err != nil {
+			log.Error(err, "failed retrieving configmap")
+		}
+		_, ok := configMap.Data["fedramp"]
+		if ok {
+			accountClaim.Spec.SupportRoleARN = fmt.Sprintf(awsv1alpha1.FedrampManagedOpenShiftSupportRoleARN, account.Spec.AwsAccountID, instanceID)
+		}
 		accountClaim.Spec.SupportRoleARN = fmt.Sprintf(awsv1alpha1.ManagedOpenShiftSupportRoleARN, account.Spec.AwsAccountID, instanceID)
```
iamkirkbater, line 276: Same here, this would need to be wrapped in an else block or be set beforehand otherwise it will overwrite it every time.

pkg/controller/accountclaim/accountclaim_controller.go
```diff
@@ -263,6 +263,16 @@ func (r *ReconcileAccountClaim) Reconcile(request reconcile.Request) (reconcile.
...
 	if accountClaim.Spec.STSRoleARN == "" {
 		instanceID := account.Labels[awsv1alpha1.IAMUserIDLabel]
+
+		// get configMap and check if account is fedramp
+		configMap, err := controllerutils.GetOperatorConfigMap(r.client)
+		if err != nil {
+			log.Error(err, "failed retrieving configmap")
+		}
+		_, ok := configMap.Data["fedramp"]
+		if ok {
```
iamkirkbater, line 273: Something about getting the configmap every time we need to check this here feels wrong to me, especially for a var that really only needs to be read once at startup and then never again.

Maybe this should go into the `newReconciler` function (or whatever it's called, I can't remember off the top of my head) and we set a new struct property on the Reconciler that is `Fedramp bool` which then allows us to do things like

```golang
accountClaim.Spec.SupportRoleARN = fmt.Sprintf....

if r.FedRamp {
    accountClaim.Spec.SupportRoleARN = fmt.Sprintf....
}
```

iamkirkbater (changes requested):

pkg/controller/accountclaim/accountclaim_controller.go
fahlmant: Alternatives are to get it once during operator startup and make it globally available, since the operator will only be running on a fedramp hive or a regular hive, OR getting it once at the start of the reconcile. I prefer the former tbh since its a global config for the whole operator

pkg/controller/account/account_controller.go
sam-nguyen7: agreed. changed by setting this beforehand

pkg/controller/accountclaim/accountclaim_controller.go
sam-nguyen7: agreed. changed by setting this beforehand

pkg/controller/accountclaim/accountclaim_controller.go
sam-nguyen7: changes pushed. now using the newReconciler func per suggestion!

hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -28,6 +28,8 @@ parameters:
   required: true
 - name: AWS_MANAGED_TAGS
   required: false
+- name: FEDRAMP
+  required: false
```
dustman9000, line 32: We should default the value to "false" to make sure we don't break commercial app-interface/OSD.
```suggestion
  required: false
  value: "false"
```

pkg/controller/account/account_controller.go
```diff
@@ -99,6 +99,13 @@ func newReconciler(mgr manager.Manager) reconcile.Reconciler {
 		log.Error(err, "shard-name key not available in configmap")
 	}
 	reconciler.shardName = hiveName
+
+	fr, ok := configMap.Data["fedramp"]
+	if !ok {
+		log.Error(err, "fedramp key not available in configmap")
+	}
+	frBool, _ := strconv.ParseBool(fr)
```
dustman9000, line 107: We should check for `nil` and add a log.Info message if we can

pkg/controller/accountclaim/accountclaim_controller.go
```diff
@@ -56,6 +57,18 @@ func newReconciler(mgr manager.Manager) reconcile.Reconciler {
...
+	configMap, err := controllerutils.GetOperatorConfigMap(reconciler.client)
+	if err != nil {
+		log.Error(err, "failed retrieving configmap")
+	}
+
+	fr, ok := configMap.Data["fedramp"]
+	if !ok {
+		log.Error(err, "fedramp key not available in configmap")
+	}
+	frBool, _ := strconv.ParseBool(fr)
```
dustman9000, line 70: Can we check for `nil` and add a `log.Info` message for detecting if we are in the fedramp env?

iamkirkbater: I don't think this type can be nil, as the configmap struct is a `map[string]string`, so you'd get "".  I'm not sure if an empty string plays nice in the strconv.ParseBool though.

fahlmant: Yes, strings default to "" and if you use bool parse, it defaults to false

dustman9000: The point was to check for an error condition here (if any).

hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -28,6 +28,9 @@ parameters:
...
```
yithian, line 32:
```suggestion
```
we don't really need to specify `required: false` in the template

yithian:
/hold

@sam-nguyen7 i think we want to close this PR with #664 open now, right?

dastergon:
Any update on this PR? Should we close in favor of the latest one?

yithian:
yes, this is replaced by #664

PR #658, opened 2021-12-08
PoC for Post Manager Init

iamkirkbater:
Proof of concept for adding a Post-Manager-Init function that will allow us to pull data from the ConfigMap and inject it into the reconciler.

I haven't had a chance to actually test this because CRC hates my computer.  This is just a jumping-off point.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -28,6 +28,8 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 
+	"github.com/openshift/aws-account-operator/pkg/controller/account"
+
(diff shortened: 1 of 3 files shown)


jharrington22:
@iamkirkbater you can create a client that quires the API server directly its not recommended to do this because the cache significantly reduces the load on the API server but since this call is being made once at start up its ok. 

```
import (
    "sigs.k8s.io/controller-runtime/pkg/client/config"
    "sigs.k8s.io/controller-runtime/pkg/client"
)

cfg, err := config.GetConfig()
...
c, err := client.New(cfg, client.Options{})
```

This should allow you to fetch the ConfigMap before the managers and controllers have started.

iamkirkbater:
Closing this as the code has been merged in another form in another PR.

PR #659, opened 2021-12-08
Renames var that was improperly named

iamkirkbater:
This renames a var in the cmd.manager file that I found that was improperly named/documented.


dastergon (approved):
/lgtm

PR #660, opened 2021-12-10
[OSD-8367] Remove references to byocadminaccess role

dkeohane:
Ref: [OSD-8367](https://issues.redhat.com/browse/OSD-8367)

- Remove references to byocadminaccess role as it has been replaced by the ManagedOpenShift-Support role


iamkirkbater:
Tested locally.  This should be good.

/lgtm
/approve

PR #662, opened 2021-12-15
Update accountpool docs

iamkirkbater:
Satisfies [OSD-7640](https://issues.redhat.com/browse/OSD-7640) and [OSD-9105](https://issues.redhat.com/browse/OSD-9105)

### DIFF
--- a/docs/3.1-AccountPool.md
+++ b/docs/3.1-AccountPool.md
@@ -44,7 +44,7 @@ Updates accountPool CR
 * `claimedAccounts` are any accounts with the `status.Claimed=true`.
 * `unclaimedAccounts` are any accounts with `status.Claimed=false` and `status.State!="Failed"`.
 * `poolSize` is the poolsize from the `AccountPool` spec.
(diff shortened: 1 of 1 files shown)


docs/3.1-AccountPool.md
```diff
@@ -34,11 +36,17 @@ Updates accountPool CR
...
   unclaimedAccounts: 3
+  availableAccounts: 5
+  accountsProgressing: 2
+  awsLimitDelta: 1
 ```
 
 * `claimedAccounts` are any accounts with the `status.Claimed=true`.
 * `unclaimedAccounts` are any accounts with `status.Claimed=false` and `status.State!="Failed"`.
 * `poolSize` is the poolsize from the `AccountPool` spec.
+* `availableAccounts` is the amount of accounts that have NEVER been claimed AND are READY to be claimed. This does NOT include Ready reused accounts. This differs from UnclaimedAccounts who similary have never been claimed but includes all non-failed states.
```
dastergon, line 47:
```suggestion
* `availableAccounts` is the amount of accounts that have NEVER been claimed AND are READY to be claimed. This does NOT include Ready reused accounts. This differs from UnclaimedAccounts who similarly have never been claimed but includes all non-failed states.
```

PR #663, opened 2022-01-04
add probing of account status

katherinelc321:
This probes accounts that are in Ready and Failed state updating the lastProbeTime if they're still true, and lastTransitionTime if they have changes from one status to another since the last probe. https://issues.redhat.com/browse/OSD-3265


pkg/controller/account/account_controller.go
```diff
@@ -225,6 +226,10 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 		return reconcile.Result{}, nil
 	}
 
+	go func() {
```
dastergon, line 229: Why did you choose to go with a goroutine?

dastergon: What if there's an error? I don't see the error being collected afterward

pkg/controller/account/account_controller.go
```diff
@@ -1290,10 +1313,66 @@ func (r *ReconcileAccount) handleCreateAdminAccessRole(
...
+		if err != nil {
+			reqLogger.Error(err, "Encountered error while creating ManagedOpenShiftSupportRole for non-CCS Account", roleID)
 			return nil, nil, err
 		}
 	}
 
 	return awsAssumedRoleClient, creds, err
 }
+
+func (r *ReconcileAccount) probeAccountState(reqLogger logr.Logger, currentAcctInstance *awsv1alpha1.Account) error {
```
dkeohane, line 1338: This is already being done as part of SetAccountCondition

PR #664, opened 2022-01-04
support govcloud

yithian:
for [OSD-6798](https://issues.redhat.com/browse/OSD-6798)

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -15,6 +15,7 @@ import (
 	"github.com/openshift/aws-account-operator/pkg/apis"
 	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
 
+	aaoconfig "github.com/openshift/aws-account-operator/config"
 	"github.com/openshift/aws-account-operator/pkg/awsclient"
 	"github.com/openshift/aws-account-operator/pkg/controller"
 	"github.com/openshift/aws-account-operator/pkg/controller/utils"
@@ -193,23 +194,38 @@ func main() {
 }
 
 func initOperatorConfigMapVars(kubeClient client.Client) {
+	// Check if config map exists.
+	cm := &corev1.ConfigMap{}
+	err := kubeClient.Get(context.TODO(), types.NamespacedName{Namespace: awsv1alpha1.AccountCrNamespace, Name: awsv1alpha1.DefaultConfigMap}, cm)
+	if err != nil {
+		log.Error(err, "There was an error getting the default configmap.")
+		return
+	}
+
+	// SetIsFedramp determines if operator is running in fedramp mode.
+	err = aaoconfig.SetIsFedramp(cm)
+	if err != nil {
+		log.Error(err, "Failed to set fedramp runtime status")
+		os.Exit(1)
+	}
+
+	// Check if fedramp env
+	if aaoconfig.IsFedramp() {
+		log.Info("Running in fedramp env")
+	}
+
+	awsRegion := aaoconfig.GetDefaultRegion()
+
+	// Get aws client
 	builder := &awsclient.Builder{}
 	awsClient, err := builder.GetClient("", kubeClient, awsclient.NewAwsClientInput{
 		SecretName: utils.AwsSecretName,
 		NameSpace:  awsv1alpha1.AccountCrNamespace,
-		AwsRegion:  "us-east-1",
+		AwsRegion:  awsRegion,
 	})
 
 	if err != nil {
 		log.Error(err, "Failed creating AWS client")
-		return
-	}
-
-	// Check if config map exists.
-	cm := &corev1.ConfigMap{}
-	err = kubeClient.Get(context.TODO(), types.NamespacedName{Namespace: awsv1alpha1.AccountCrNamespace, Name: awsv1alpha1.DefaultConfigMap}, cm)
-	if err != nil {
-		log.Error(err, "There was an error getting the default configmap.")
 		return
 	}
 
--- a/config/config.go
+++ b/config/config.go
@@ -14,10 +14,70 @@
 
 package config
 
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+
+	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
+)
+
 const (
 	// OperatorName stores the name used by this code for the AWS Account Operator
 	OperatorName string = "aws-account-operator"
 
 	// OperatorNamespace stores a string indicating the Kubernetes namespace in which the operator runs
 	OperatorNamespace string = "aws-account-operator"
+
+	// used in constructing ARNs
+	AwsResourceTypeRole                  string = "role"
+	AwsResourceTypePolicy                string = "policy"
+	AwsResourceIDAdministratorAccessRole string = "AdministratorAccess"
 )
+
+var (
+	isFedramp = false
+)
+
+// SetIsFedramp sets the var isFedramp to value in default configmap
+func SetIsFedramp(configMap *corev1.ConfigMap) error {
+	fedramp, ok := configMap.Data["fedramp"]
+	if !ok {
+		// Since fedramp param is not required, if fedramp param does not exist then assume fedramp=false
+		isFedramp = false
+		return nil
+	}
+	frBool, err := strconv.ParseBool(fedramp)
+	if err != nil {
+		return fmt.Errorf("Invalid value for configmap fedramp. %w", err)
+	}
+	isFedramp = frBool
+	return nil
+}
+
+// IsFedramp returns value of isFedramp var
+func IsFedramp() bool {
+	return isFedramp
+}
+
+func GetDefaultRegion() (regionName string) {
+	regionName = awsv1alpha1.AwsUSEastOneRegion
+	if isFedramp {
+		regionName = awsv1alpha1.AwsUSGovEastOneRegion
+	}
+	return
+}
+
+// construct an ARN
+func GetIAMArn(awsAccountID, awsResourceType, awsResourceID string) (arn string) {
+	awsAPI := "aws"
+	if isFedramp {
+		awsAPI = "aws-us-gov"
+	}
+
+	// arn:partition:service:region:account-id:resource-type/resource-id
+	arn = strings.Join([]string{"arn:", awsAPI, ":iam::", awsAccountID, ":", awsResourceType, "/", awsResourceID}, "")
+	return
+}
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -15,6 +15,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/organizations"
 	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/go-logr/logr"
+	"github.com/openshift/aws-account-operator/config"
 	"github.com/openshift/aws-account-operator/pkg/controller/utils"
 	"github.com/rkt/rkt/tests/testutils/logger"
 	corev1 "k8s.io/api/core/v1"
@@ -29,10 +30,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
-	"github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
 	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
 	"github.com/openshift/aws-account-operator/pkg/awsclient"
-	controllerutils "github.com/openshift/aws-account-operator/pkg/controller/utils"
 	totalaccountwatcher "github.com/openshift/aws-account-operator/pkg/totalaccountwatcher"
 )
 
@@ -67,8 +66,9 @@ const (
 	// AccountPendingVerification indicates verification (of AWS limits and Enterprise Support) is pending
 	AccountPendingVerification = "PendingVerification"
 
-	adminAccessArn = "arn:aws:iam::aws:policy/AdministratorAccess"
-	iamUserNameUHC = "osdManagedAdmin"
+	standardAdminAccessArnPrefix = "arn:aws:iam"
+	adminAccessArnSuffix         = "::aws:policy/AdministratorAccess"
+	iamUserNameUHC               = "osdManagedAdmin"
 
 	controllerName = "account"
 )
@@ -87,7 +87,7 @@ func newReconciler(mgr manager.Manager) reconcile.Reconciler {
 		awsClientBuilder: &awsclient.Builder{},
 	}
 
-	configMap, err := controllerutils.GetOperatorConfigMap(reconciler.Client)
+	configMap, err := utils.GetOperatorConfigMap(reconciler.Client)
 	if err != nil {
 		log.Error(err, "failed retrieving configmap")
 	}
@@ -97,6 +97,7 @@ func newReconciler(mgr manager.Manager) reconcile.Reconciler {
 		log.Error(err, "shard-name key not available in configmap")
 	}
 	reconciler.shardName = hiveName
+
 	return utils.NewReconcilerWithMetrics(reconciler, controllerName)
 }
 
@@ -145,11 +146,18 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 		return reconcile.Result{}, err
 	}
 
+	configMap, err := utils.GetOperatorConfigMap(r.Client)
+	if err != nil {
+		log.Error(err, "Failed retrieving configmap")
+		return reconcile.Result{}, err
+	}
+
+	awsRegion := config.GetDefaultRegion()
 	// We expect this secret to exist in the same namespace Account CR's are created
 	awsSetupClient, err := r.awsClientBuilder.GetClient(controllerName, r.Client, awsclient.NewAwsClientInput{
 		SecretName: utils.AwsSecretName,
 		NameSpace:  awsv1alpha1.AccountCrNamespace,
-		AwsRegion:  "us-east-1",
+		AwsRegion:  awsRegion,
 	})
 	if err != nil {
 		reqLogger.Error(err, "failed building operator AWS client")
@@ -285,7 +293,7 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 			_, stateErr := r.setAccountFailed(
 				reqLogger,
 				currentAcctInstance,
-				v1alpha1.AccountCreationFailed,
+				awsv1alpha1.AccountCreationFailed,
 				"CreationTimeout",
 				errMsg,
 				AccountFailed,
@@ -301,9 +309,12 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 			if !currentAcctInstance.HasAwsAccountID() {
 				// before doing anything make sure we are not over the limit if we are just error
 				if !totalaccountwatcher.TotalAccountWatcher.AccountsCanBeCreated() {
-					reqLogger.Error(awsv1alpha1.ErrAwsAccountLimitExceeded, "AWS Account limit reached")
-					// We don't expect the limit to change very frequently, so wait a while before requeueing to avoid hot lopping.
-					return reconcile.Result{Requeue: true, RequeueAfter: time.Duration(5) * time.Minute}, nil
+					// fedramp clusters are all CCS, so the account limit is irrelevant there
+					if !config.IsFedramp() {
+						reqLogger.Error(awsv1alpha1.ErrAwsAccountLimitExceeded, "AWS Account limit reached")
+						// We don't expect the limit to change very frequently, so wait a while before requeueing to avoid hot lopping.
+						return reconcile.Result{Requeue: true, RequeueAfter: time.Duration(5) * time.Minute}, nil
+					}
 				}
 
 				if err := r.nonCCSAssignAccountID(reqLogger, currentAcctInstance, awsSetupClient); err != nil {
@@ -321,7 +332,7 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 		}
 	}
 
-	configMap, err := controllerutils.GetOperatorConfigMap(r.Client)
+	// Get regions from configmap data
 	stringRegions, ok := configMap.Data["regions"]
 	if !ok {
 		err = awsv1alpha1.ErrInvalidConfigMap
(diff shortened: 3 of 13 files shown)


pkg/controller/account/account_controller.go
```diff
@@ -145,11 +148,25 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 		return reconcile.Result{}, err
 	}
 
+	// Determine if in fedramp env
+	configMap, err := controllerutils.GetOperatorConfigMap(r.Client)
+	if err != nil {
+		log.Error(err, "Failed retrieving configmap")
+	}
+	ifFedramp, err := controllerutils.IsFedramp(configMap)
```
NautiluX, line 156:
```suggestion
	isFedramp, err := controllerutils.IsFedramp(configMap)
```

pkg/controller/account/account_controller.go
```diff
@@ -145,11 +148,25 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 		return reconcile.Result{}, err
 	}
 
+	// Determine if in fedramp env
+	configMap, err := controllerutils.GetOperatorConfigMap(r.Client)
+	if err != nil {
+		log.Error(err, "Failed retrieving configmap")
```
NautiluX, line 154: Wouldn't we want to cancel the reconcilation here if we can't read the configmap?

pkg/controller/account/account_controller.go
```diff
@@ -145,11 +148,25 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
...
+	}
+	ifFedramp, err := controllerutils.IsFedramp(configMap)
+	if err != nil {
+		log.Error(err, "Unable to verify if cluster is fedramp")
+	}
+
+	awsRegion := "us-east-1"
+	if ifFedramp {
+		awsRegion = "us-gov-east-1"
+	}
```
NautiluX, line 164: would it make sense to use the new constants here?

pkg/controller/account/account_controller.go
```diff
@@ -321,7 +341,7 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 		}
 	}
 
-	configMap, err := controllerutils.GetOperatorConfigMap(r.Client)
+	// Get region from configmap data
```
NautiluX, line 344:
```suggestion
	// Get regions from configmap data
```

pkg/controller/account/account_controller.go
```diff
@@ -643,7 +663,23 @@ func (r *ReconcileAccount) assumeRole(
...
+
+	// Determine if in fedramp env
+	configMap, err := controllerutils.GetOperatorConfigMap(r.Client)
+	if err != nil {
+		log.Error(err, "Failed retrieving configmap")
+	}
+	ifFedramp, err := controllerutils.IsFedramp(configMap)
+	if err != nil {
+		log.Error(err, "Unable to verify if cluster is fedramp")
+	}
```
NautiluX, line 677: do we really need to read the configmap again?
If so, wouldn't we want to return an error if we can't read it?

pkg/controller/account/account_controller.go
```diff
@@ -688,11 +724,16 @@ func (r *ReconcileAccount) assumeRole(
 			break
 		}
 	}
+
+	awsRegion := "us-east-1"
+	if ifFedramp {
+		awsRegion = "us-gov-east-1"
+	}
```
NautiluX, line 731: should we use the new constants here?

pkg/controller/account/account_controller.go
```diff
@@ -716,15 +757,28 @@ func (r *ReconcileAccount) assumeRole(
...
 func (r *ReconcileAccount) initializeRegions(reqLogger logr.Logger, currentAcctInstance *awsv1alpha1.Account, creds *sts.AssumeRoleOutput, regionAMIs map[string]awsv1alpha1.AmiSpec) error {
+	// Determine if in fedramp env
+	configMap, err := controllerutils.GetOperatorConfigMap(r.Client)
+	if err != nil {
+		log.Error(err, "Failed retrieving configmap")
+	}
+	ifFedramp, err := controllerutils.IsFedramp(configMap)
+	if err != nil {
+		log.Error(err, "Unable to verify if cluster is fedramp")
+	}
```
NautiluX, line 768: do we really need to read the configmap again?
If so, wouldn't we want to return an error if we can't read it?
I see these lines duplicated multiple times in the PR, probably worth encapsulating in their own function?

pkg/controller/accountclaim/accountclaim_controller.go
```diff
@@ -221,11 +222,25 @@ func (r *ReconcileAccountClaim) Reconcile(request reconcile.Request) (reconcile.
...
+			log.Error(err, "failed retrieving configmap")
+		}
+		ifFedramp, err := utils.IsFedramp(configMap)
+		if err != nil {
+			log.Error(err, "Unable to verify if cluster is fedramp")
+		}
+		awsRegion := "us-east-1"
+		if ifFedramp {
+			awsRegion = "us-gov-east-1"
+		}
```
NautiluX, line 237: Wouldn't we want to return an error if we can't read the config map and cancel reconcilation?
I see these lines duplicated multiple times in the PR, probably worth encapsulating in their own function?

pkg/controller/accountclaim/accountclaim_controller.go
```diff
@@ -264,6 +279,19 @@ func (r *ReconcileAccountClaim) setSupportRoleARNManagedOpenshift(reqLogger logr
...
+
+		// if account if fedramp use the appropriate iam arn
+		configMap, err := utils.GetOperatorConfigMap(r.client)
+		if err != nil {
+			log.Error(err, "failed retrieving configmap")
+		}
+		ifFedramp, err := utils.IsFedramp(configMap)
+		if err != nil {
+			log.Error(err, "Unable to verify if cluster is fedramp")
+		}
```
NautiluX, line 291: Do we really need to read the config map here again?

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -581,11 +594,24 @@ func (r *ReconcileAWSFederatedAccountAccess) cleanFederatedRoles(reqLogger logr.
...
 	roleName := currentFAA.Spec.AWSFederatedRole.Name + "-" + uidLabel
 
+	configMap, err := controllerutils.GetOperatorConfigMap(r.client)
+	if err != nil {
+		log.Error(err, "failed retrieving configmap")
+	}
+	ifFedramp, err := controllerutils.IsFedramp(configMap)
+	if err != nil {
+		log.Error(err, "Unable to verify if cluster is fedramp")
+	}
```
NautiluX, line 604: Wouldn't we want to return an error if we can't read the config map and cancel reconcilation?
I see these lines duplicated multiple times in the PR, probably worth encapsulating in their own function?

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -581,11 +594,24 @@ func (r *ReconcileAWSFederatedAccountAccess) cleanFederatedRoles(reqLogger logr.
...
+
 	// Build AWS client from root secret
+	awsRegion := "us-east-1"
+	if ifFedramp {
+		awsRegion = "us-gov-east-1"
+	}
```
NautiluX, line 610: should we use the new constants here?

pkg/controller/awsfederatedrole/awsfederatedrole_controller.go
```diff
@@ -87,9 +87,23 @@ type ReconcileAWSFederatedRole struct {
...
 	reqLogger := log.WithValues("Controller", controllerName, "Request.Namespace", request.Namespace, "Request.Name", request.Name)
 
+	configMap, err := utils.GetOperatorConfigMap(r.client)
+	if err != nil {
+		log.Error(err, "failed retrieving configmap")
+	}
+	ifFedramp, err := utils.IsFedramp(configMap)
+	if err != nil {
+		log.Error(err, "Unable to verify if cluster is fedramp")
+	}
```
NautiluX, line 97: Wouldn't we want to return an error if we can't read the config map and cancel reconcilation?
I see these lines duplicated multiple times in the PR, probably worth encapsulating in their own function?

pkg/controller/awsfederatedrole/awsfederatedrole_controller.go
```diff
@@ -134,10 +148,14 @@ func (r *ReconcileAWSFederatedRole) Reconcile(request reconcile.Request) (reconc
 		return reconcile.Result{}, nil
 	}
 	// Setup AWS client
+	awsRegion := "us-east-1"
+	if ifFedramp {
+		awsRegion = "us-gov-east-1"
+	}
```
NautiluX, line 154: should we use the new constants here?

pkg/totalaccountwatcher/totalaccountwatcher.go
```diff
@@ -46,14 +46,29 @@ type AccountWatcher struct {
 func initialize(client client.Client, watchInterval time.Duration) *AccountWatcher {
 	log.Info("Initializing the totalAccountWatcher")
 
+	cm := &corev1.ConfigMap{}
+	err := client.Get(context.TODO(), types.NamespacedName{Namespace: awsv1alpha1.AccountCrNamespace, Name: awsv1alpha1.DefaultConfigMap}, cm)
+	if err != nil {
+		log.Error(err, "Failed to get configmap")
+	}
```
NautiluX, line 53: isn't this what the `GetOperatorConfigMap` function does? What's the reason we can't use it here?

pkg/totalaccountwatcher/totalaccountwatcher.go
```diff
@@ -46,14 +46,29 @@ type AccountWatcher struct {
...
+
+	awsRegion := awsv1alpha1.AwsUSEastOneRegion
+	fr, ok := cm.Data["fedramp"]
+	if ok {
+		frBool, _ := strconv.ParseBool(fr)
```
NautiluX, line 58: is it ok to ignore the error here?

NautiluX (changes requested):
Hi, thanks for this PR!
I see that it requires reading the configmap a lot to read the new bool value.
I wonder if this value could ever change while running the operator or if we could just read it once when starting the operator?
If not I think it would make sense to at least encapsulate reading the configmap and the value into a function and/or ensuring it's only read once during each reconciliation.

cmd/manager/main.go
```diff
@@ -193,26 +194,38 @@ func main() {
...
+	// Check if config map exists.
+	cm := &corev1.ConfigMap{}
+	err := kubeClient.Get(context.TODO(), types.NamespacedName{Namespace: awsv1alpha1.AccountCrNamespace, Name: awsv1alpha1.DefaultConfigMap}, cm)
+	if err != nil {
+		log.Error(err, "There was an error getting the default configmap.")
+		return
+	}
+
+	// Check if fedramp env
+	awsRegion := "us-east-1"
```
dkeohane, line 206: Let's use AwsUSEastOneRegion here instead

cmd/manager/main.go
```diff
@@ -193,26 +194,38 @@ func main() {
...
+	fr, ok := cm.Data["fedramp"]
+	if ok {
+		frBool, _ := strconv.ParseBool(fr)
+		if frBool {
+			awsRegion = "us-gov-east-1"
```
dkeohane, line 211: Similar to above, let's use AwsUSGovEastOneRegion

pkg/totalaccountwatcher/totalaccountwatcher.go
yithian: good catch. i was assuming that successfully reading `cm.Data["fedramp"]` would ensure that `fr` is a bool but we shouldn't make that assumption

yithian: actually, if `cm.Data["fedramp"]` returns any value in `fr`, then `strconf.ParseBool()` _should_ return either true or false in `frBool`

but there's no reason to tempt fate. i'll check the error code anyway :)

pkg/controller/account/account_controller.go
yithian: @NautiluX i think b0755f238ec7e5ac0feea39ba9bbf665baa8e4e4 should address this

i'm not sure i like the "only check at operator startup" model, since to update the value you'd have to restart the operator. i'd much rather it be able to adjust to a change in the configmap while running, even if that means a few more cycles are spent checking the configmap

what do you think?

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -192,11 +194,24 @@ func (r *ReconcileAWSFederatedAccountAccess) Reconcile(request reconcile.Request
 		return reconcile.Result{}, err
 	}
 
+	configMap, err := controllerutils.GetOperatorConfigMap(r.client)
+	if err != nil {
+		log.Error(err, "failed retrieving configmap")
+	}
+	config.IsFedramp, err = controllerutils.IsFedramp(configMap)
```
NautiluX, line 201: so, you're setting a global variable here, but update it on each reconcile, which seems like a dangerous pattern, can you think of a different way to solve this?
At least, I think you could hide that complexity inside the config package, so the global var is not set outside.

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -192,11 +194,24 @@ func (r *ReconcileAWSFederatedAccountAccess) Reconcile(request reconcile.Request
...
+	if err != nil {
+		log.Error(err, "Unable to verify if cluster is fedramp")
+	}
+
 	// Get aws client
+	awsRegion := v1alpha1.AwsUSEastOneRegion
+	if config.IsFedramp {
+		awsRegion = v1alpha1.AwsUSGovEastOneRegion
+	}
```
NautiluX, line 210: Can we reduce the amount of `if config.IsFedramp` change region occurances?
E.g. we could make a function `config.GetDefaultRegion()` which would return `v1alpha1.AwsUSGovEastOneRegion` if fedramp and `v1alpha1.AwsUSEastOneRegion` if not.

NautiluX (changes requested):
Thanks for your changes! Looks much cleaner now. However, if possible making that global setting not leave the config package would be neat, at the same time, we could reduce code duplication.

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
yithian: done

pkg/controller/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
yithian: i've moved checking for fedramp into operator startup

NB this will mean the operator will need to be restarted if the fedramp value in the configmap changes

cmd/manager/main.go
```diff
@@ -193,26 +195,39 @@ func main() {
...
+		log.Error(err, "There was an error getting the default configmap.")
+		return
+	}
+
+	// Check if fedramp env
+	fr, ok := cm.Data["fedramp"]
+	if ok {
+		frBool, _ := strconv.ParseBool(fr)
+		if frBool {
+			aaoconfig.IsFedramp = frBool
```
sam-nguyen7, line 211: +1 with setting this once at start up instead of each reconcile. However, I still think we should hide the complexity in the config package like https://github.com/sam-nguyen7/aws-account-operator/commit/32652f2c7c2436025de2ae9aa06a29177a2d04a4#diff-435eecb1d2af40ee96747f88ab71eb2758da989c92f76dcb1f714fc1b300c633R186 . What do you think?

sam-nguyen7 (commented):
Default region logic looks good to me.

cmd/manager/main.go
yithian: i've cherry-picked this commit into this PR

thanks!

config/config.go
```diff
@@ -14,10 +14,60 @@
...
+	return isFedramp
+}
+
+func GetDefaultRegion() (regionName string) {
+	regionName = awsv1alpha1.AwsUSEastOneRegion
+	if isFedramp {
+		regionName = awsv1alpha1.AwsUSGovEastOneRegion
+	}
+	return
+}
```
NautiluX, line 73: nice encapsulation of the defaultregion! :+1:

pkg/controller/account/account_controller.go
```diff
@@ -643,7 +655,13 @@ func (r *ReconcileAccount) assumeRole(
...
 	// The role ARN made up of the account number and the role which is the default role name
 	// created in child accounts
-	var roleArn = fmt.Sprintf("arn:aws:iam::%s:role/%s", currentAcctInstance.Spec.AwsAccountID, roleToAssume)
+	var roleArn string
+	roleArn = fmt.Sprintf("arn:aws:iam::%s:role/%s", currentAcctInstance.Spec.AwsAccountID, roleToAssume)
+
+	// if Fedramp change the role ARN
+	if config.IsFedramp() {
+		roleArn = fmt.Sprintf("arn:aws-us-gov:iam::%s:role/%s", currentAcctInstance.Spec.AwsAccountID, roleToAssume)
+	}
```
NautiluX, line 664: would it be possible to encapsulate the roleArn(s) in a similar way as the default regions?

pkg/controller/accountclaim/accountclaim_controller.go
```diff
@@ -264,6 +267,11 @@ func (r *ReconcileAccountClaim) setSupportRoleARNManagedOpenshift(reqLogger logr
 	if accountClaim.Spec.STSRoleARN == "" {
 		instanceID := account.Labels[awsv1alpha1.IAMUserIDLabel]
 		accountClaim.Spec.SupportRoleARN = fmt.Sprintf(awsv1alpha1.ManagedOpenShiftSupportRoleARN, account.Spec.AwsAccountID, instanceID)
+
+		// if account if fedramp use the appropriate iam arn
+		if config.IsFedramp() {
+			accountClaim.Spec.SupportRoleARN = fmt.Sprintf(awsv1alpha1.FedrampManagedOpenShiftSupportRoleARN, account.Spec.AwsAccountID, instanceID)
+		}
```
NautiluX, line 274: jprobably you could make something like `config.GetSupportRoleArn()` similar to `getDefaultRegion()`

NautiluX (commented):
I like that a lot of the complexity added by the fedramp switches is hidden in the config package, I'm just wondering if we could do that even more, I highlighted a few spots where I think it would make sense.

pkg/controller/account/iam.go
```diff
@@ -265,14 +265,16 @@ func CreateIAMUser(reqLogger logr.Logger, client awsclient.Client, userName stri
 // AttachAdminUserPolicy attaches the AdministratorAccess policy to a target user
 // Takes a logger, an AWS client for the target account, and the target IAM user's username
 func AttachAdminUserPolicy(client awsclient.Client, iamUser *iam.User) (*iam.AttachUserPolicyOutput, error) {
+	// grab the ARN base to support govcloud without passing in another argument
+	awsARNBase := strings.Split(*iamUser.Arn, "::")[0]
```
dkeohane, line 269: Could be handled like the others with `if config.IsFedramp() ` instead of a string split?

pkg/controller/account/account_controller.go
```diff
@@ -1213,6 +1234,13 @@ func (r *ReconcileAccount) handleCreateAdminAccessRole(
 	currentAccInstanceID := currentAcctInstance.Labels[awsv1alpha1.IAMUserIDLabel]
 	roleToAssume := getAssumeRole(currentAcctInstance)
 
+	var adminAccessArn string
+	if config.IsFedramp() {
```
dkeohane, line 1238: Can we encapsulate this and other such actions into standalone functions?

pkg/controller/utils/utils.go
```diff
@@ -219,3 +220,15 @@ func GetOperatorConfigMap(kubeClient client.Client) (*corev1.ConfigMap, error) {
...
+func IsFedramp(configMap *corev1.ConfigMap) (bool, error) {
+	// Check if fedramp env
+	fr, ok := configMap.Data["fedramp"]
+	if !ok {
+		// Since fedramp param is not required, if fedramp param does not exist then assume fedramp=false
+		return false, nil
+	}
+
+	return strconv.ParseBool(fr)
+}
```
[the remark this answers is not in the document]
yithian: no it's not, i'll remove it

pkg/totalaccountwatcher/totalaccountwatcher.go
```diff
@@ -46,14 +47,31 @@ type AccountWatcher struct {
 func initialize(client client.Client, watchInterval time.Duration) *AccountWatcher {
 	log.Info("Initializing the totalAccountWatcher")
 
+	cm, err := utils.GetOperatorConfigMap(client)
+	if err != nil {
+		log.Error(err, "Failed to get configmap")
+	}
+
+	awsRegion := awsv1alpha1.AwsUSEastOneRegion
+	fr, ok := cm.Data["fedramp"]
```
dkeohane, line 56: This can all be replaced with `if isFedramp()` ? this would avoid getting the configmap again

dkeohane (changes requested):

pkg/totalaccountwatcher/totalaccountwatcher.go
yithian: actually, we can replace that whole section with `config.GetDefaultRegion()` now

pkg/controller/account/account_controller.go
```diff
@@ -1213,6 +1228,8 @@ func (r *ReconcileAccount) handleCreateAdminAccessRole(
 	currentAccInstanceID := currentAcctInstance.Labels[awsv1alpha1.IAMUserIDLabel]
 	roleToAssume := getAssumeRole(currentAcctInstance)
 
+	adminAccessArn := config.GetIAMArn("", config.AwsResourceTypePolicy, config.AwsResourceIDAdministratorAccessRole)
```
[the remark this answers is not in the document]
yithian: i couldn't quite make sense of the `arn:aws:iam::aws:policy/AdministratorAccess` arn, is the second `aws` an account id? https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html

regardless, i've added a commit that should get the right arn for this now

iamkirkbater:
This all looks sane to me, and it's been reviewed heavily.  Dan mentioned it was tested locally using CRC and it didn't break any existing functionality within AAO.  The new govcloud functionality will need to be tested in govcloud accounts, so let's move this forward and merge it so that it can be tested in govcloud accounts.

/lgtm

PR #665, opened 2022-01-06
Add namespace to test accountpool CR

sam-nguyen7:
When running  "make deploy-cluster" the accountpool CR in hack/files/aws.managed.openshift.io_v1alpha1_zero_size_accountpool.yaml is getting deployed to the default namespace. This PR fixes that by adding in the namespace field in the metadata


NautiluX (commented):
/lgtm

PR #670, opened 2022-02-09
Scripts out Deployment Validation

iamkirkbater:
Currently tests both `non-ccs` and `ccs` cluster creation using OCM.

This script assumes you're already logged into whatever OCM environment you are attempting to test.

This script also assumes you have a valid set of credentials for an IAM User: `osdCCSAdmin` in your `~/.aws/credentials` file (you pass the profile name in as the parameter).

Example:

```
$ cat ~/.aws/credentials
[osd-staging-2-ccsadmin]
aws_access_key_id = myaccesskeyid
aws_secret_access_key_id = mysecretaccesskeyid
```

Given the above file, you would call the script as such: `./hack/scripts/validate_deployment.sh -p osd-staging-2-ccsadmin`

This script runs the creation in parallel, allowing you to run the script and just wait for the output to finish.  The script also only waits for the cluster to move from `pending` to `installing` as once the cluster is `installing` the work of AWS Account Operator is complete.

Future work: Script out STS creation.

### DIFF
--- a/hack/scripts/validate_deployment.sh
+++ b/hack/scripts/validate_deployment.sh
@@ -1,0 +1,280 @@
+#!/bin/bash
+
+usage() {
+  cat <<EOF
+  usage: $0 [ OPTIONS ]
+
+  validate_deployment.sh will create a cluster using OCM in each of 'non-ccs' and 'ccs' types. This script should be used to test aws-account-operator deployments, and should be run after new pods are deployed to the different environments.
+
+  This script assumes that you are already logged into an OCM environment, and that you have a set of osdCcsAdmin credentials in your ~/.aws/credentials directory for the account you want to create a CCS cluster in.
+  Options
+  -e  --environment  Which OCM environment to run tests against
+  -h  --help         Show this message and exit
+  -p  --profile      Which AWS Profile should we use?
+EOF
+}
+
+# Ensure there's a parameter passed in
+if [ $# -lt 2 ]; then
+  usage
+  exit 1
+fi
+
+while [ "$1" != "" ]; do
+  case "$1" in
+    -e | --env | --environment) OCM_ENVIRONMENT=$2
+                                shift
+                                ;;
+    -h | --help )           usage
+                            exit 0
+                            ;;
+    -p | --profile )        PROFILE=$2
+                            shift
+                            ;;
+
+    * ) echo "Unexpected parameter $1"
+        usage
+        exit 1
+  esac
+  shift
+done
+
+export YELLOW="\033[33m"
+export RED="\033[31m"
+export GREEN="\033[32m"
+export RESET="\033[0m"
+
+export BULLET="\xe2\x97\x89"
+export ARROW="\xe2\x96\xb8"
+export PENDING="${YELLOW}${ARROW}${RESET}"
+export ERROR="${RED}!!${RESET}"
+export SUCCESS="${GREEN}${BULLET}${RESET}"
+
+## Define all functions, scroll to the bottom for the entrypoint of the script
+
+run_test()
+{
+  local TEST_NAME=$1
+  local TEST_CASE=$2
+  local DATETIME=$(date +%m%d)
+
+  case $TEST_NAME in
+    non-ccs)
+      local CLUSTER_NAME="aao-$DATETIME"
+      echo -e "$PENDING Creating Non-CCS Cluster $CLUSTER_NAME"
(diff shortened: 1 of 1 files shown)


hack/scripts/validate_deployment.sh
```diff
@@ -0,0 +1,276 @@
...
+  # Set the timeout to be 900 seconds, or 15m
+  local CREATE_TIMEOUT=900
+  local DELETE_TIMEOUT=900
+  
+  # Check Interval is how long between checks we go, in seconds
+  local CHECK_INTERVAL=5
+  
+  install_wait_loop $CLUSTER_ID $CREATE_TIMEOUT $CHECK_INTERVAL
+
+  ocm delete cluster $CLUSTER_ID
```
macgregor, line 188: how paranoid do we want to be about ensuring tests clean up after themselves? You may want to consider using a trap (https://www.linuxjournal.com/content/bash-trap-command) to ensure clean up happens even in script failure scenarios. That may require more engineering than is necessary though.

macgregor (changes requested):
looks pretty good so far. few questions/asks

hack/scripts/validate_deployment.sh
iamkirkbater: Yeah, I was thinking about that, but then like you mentioned the engineering into that started to yak-shave.

Like, I could add the trap, but then I'd need to know whether or not the clusters are initiating already or not and then clean it up.  And then catch it in the subprocesses, and ignore it in the delete wait loop, so I just opted out of doing it at all.

It might be worth adding a simple trap though that just lets you know that depending on the state of each thread you might have to manually clean something up.

hack/scripts/validate_deployment.sh
```diff
@@ -0,0 +1,280 @@
...
+      local CLUSTER_NAME="aao-$DATETIME"
+      echo -e "$PENDING Creating Non-CCS Cluster $CLUSTER_NAME"
+      OCM_CREATE=$(ocm create cluster --region us-east-1 $CLUSTER_NAME)
+      if [[ $? -ne 0 ]]; then
+        echo -e "$ERROR $TEST_NAME failed.  Unable to create cluster."
+        exit 1
+      fi
+
+      # get the cluster ID from the OCM output
+      local CLUSTER_ID=$(echo $OCM_CREATE | grep '^ID:' | awk '{print $2}')
```
georgettica, line 72: created an ocm-cli issue to make this less janky https://github.com/openshift-online/ocm-cli/issues/383

hack/scripts/validate_deployment.sh
```diff
@@ -0,0 +1,280 @@
...
+    ccs)
+      local CLUSTER_NAME="aao-ccs-$DATETIME"
+      echo -e "$PENDING Creating CCS Cluster $CLUSTER_NAME"
+      eval $(get_aws_credentials)
+
+      echo -e "$PENDING Getting AWS Account Info from Profile"
+      ACCOUNT=$(aws sts get-caller-identity | jq -r .Account)
+      echo -e "$PENDING AWS Account $ACCOUNT info collected."
+
+      OCM_CREATE=$(ocm create cluster --region us-east-1 --ccs --aws-access-key-id $AWS_ACCESS_KEY_ID --aws-secret-access-key $AWS_SECRET_ACCESS_KEY --aws-account-id $ACCOUNT $CLUSTER_NAME)
```
georgettica, line 87: that is a very nice way to create a CCS cluster. I hope there is something similar to GCP as well @rafael-azevedo

hack/scripts/validate_deployment.sh
```diff
@@ -0,0 +1,280 @@
...
+  local CLUSTER_ID=$1
+  local TIMEOUT=$2
+  local CHECK_INTERVAL=$3
+  # Loop 
+  while true; do
+    sleep $CHECK_INTERVAL
+    TIMEOUT=$(( TIMEOUT - CHECK_INTERVAL ))
+    local STATUS=$(ocm get /api/clusters_mgmt/v1/clusters/$CLUSTER_ID/status | jq -r .state)
+    if [[ $STATUS == "installing" ]]; then
+      echo -e "$SUCCESS $TEST_NAME has passed. Beginning Cleanup..."
```
georgettica, line 123: maybe split this into 2:
so the echo here will just note the test has passed

and the begin cleanup will be higher in the stack?

dastergon:
@iamkirkbater is there anything blocking this PR?

PR #671, opened 2022-02-10
naming the ec2 test instance that is created in each aws region

macgregor:
Closes [OSD-7011](https://issues.redhat.com/browse/OSD-7011) - provide a name for the AWS ec2 region test instances.

### DIFF
--- a/pkg/awsclient/tags.go
+++ b/pkg/awsclient/tags.go
@@ -42,6 +42,10 @@ func (t *AWSAccountOperatorTags) GetEC2Tags() []*ec2.Tag {
 	for _, tag := range t.Tags {
 		tags = append(tags, &ec2.Tag{Key: aws.String(tag.Key), Value: aws.String(tag.Value)})
 	}
+
+	//make sure the ec2 instance has a descriptive name to avoid customer confusion
+	tags = append(tags, &ec2.Tag{Key: aws.String("Name"), Value: aws.String("red-hat-region-init")})
+
 	return tags
 }
 


fahlmant:
Please provide a description and a jira ticket reference

pkg/awsclient/tags.go
```diff
@@ -42,6 +42,10 @@ func (t *AWSAccountOperatorTags) GetEC2Tags() []*ec2.Tag {
 	for _, tag := range t.Tags {
 		tags = append(tags, &ec2.Tag{Key: aws.String(tag.Key), Value: aws.String(tag.Value)})
 	}
+
+	//make sure the ec2 instance has a descriptive name to avoid customer confusion
+	tags = append(tags, &ec2.Tag{Key: aws.String("Name"), Value: aws.String("red-hat-region-init")})
```
fahlmant, line 47: The following tag also needs to be added: `red-hat-managed: true`

gregsheremeta: what does `red-hat-managed` have to do with OSD-7011? Should that be a separate card and PR, or is there some relation to this change?

fahlmant: I discussed with the v1alpha1 team aligning all tags across the verifier and the aws-account-operator with OSD-7011 as the card I saw adding tags to the operator. I assumed OSD-7011 had been updated with the need for this tag, but it wasn't. As it stands now, you're right, 7011 doesn't mention this tag.

gregsheremeta: suggest opening a new PR for `red-hat-managed: true`. I opened:
[OSD-9949] Add the `red-hat-managed: true` tag to AAO-created EC2 instances

gregsheremeta:
lgtm, as long as y'all are ok with hardcoding magic strings wherever :)

macgregor:
> Please provide a description and a jira ticket reference
The jira number is in the branch name and the description is the same as the branch name, the MR title and the code comment with the change. Figured the description was pretty redundant, but I added it.

iamkirkbater:
> jira number is in the branch name

I wish the bot that links JIRA tickets to these PRs made that connection :(  Until then, though, we have to keep adding it to the description.

fahlmant:
The PRs here are public, however the jira cards not always so. Providing a description is necessary because not everyone knows or can see the context behind the change. There's also not always going to be a jira behind a change, so adding a description is a good regular practice

PR #675, opened 2022-02-21
[OSD-9246] Block gaps in our AccountClaim finalizer handling

dkeohane:
While investigating dangling accounts I looked through our accountclaim finalize functionality to try identify areas where we may not be returning errors

Previous behaviours which have been addressed as part of this PR: 
- If the accountclaim `finalizeAccountClaim` function encountered an error, we would still proceed to remove the finalizer.
- If the `cleanUpAwsAccount` func encountered an error, we simply logged it instead of returning it.

Additions:
- Added some more unit tests. 
- Made errors from AWS visible in our logging, as debugging why we can't clean up an AWS resource without the context from AWS was impossible.

Ticket ref: [OSD-9246](https://issues.redhat.com/browse/OSD-9246)

### DIFF
--- a/pkg/controller/accountclaim/accountclaim_controller.go
+++ b/pkg/controller/accountclaim/accountclaim_controller.go
@@ -306,14 +306,16 @@ func (r *ReconcileAccountClaim) handleAccountClaimDeletion(reqLogger logr.Logger
 			failedReusedAccount, accountErr := r.getClaimedAccount(accountClaim.Spec.AccountLink, awsv1alpha1.AccountCrNamespace)
 			if accountErr != nil {
 				reqLogger.Error(accountErr, "Failed to get claimed account")
-				return err
+				return errors.Wrap(err, "failed to get claimed account")
 			}
 			// Update account status and add "Reuse Failed" condition
 			accountErr = r.resetAccountSpecStatus(reqLogger, failedReusedAccount, accountClaim, awsv1alpha1.AccountFailed, "Failed")
 			if accountErr != nil {
 				reqLogger.Error(accountErr, "Failed updating account status for failed reuse")
-				return err
+				return errors.Wrap(err, "failed updating account status for failed reuse")
 			}
+
+			return err
 		}
 	}
 
--- a/pkg/controller/accountclaim/reuse.go
+++ b/pkg/controller/accountclaim/reuse.go
@@ -2,9 +2,11 @@ package accountclaim
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"time"
+
+	"github.com/pkg/errors"
+	"github.com/rkt/rkt/tests/testutils/logger"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
@@ -36,12 +38,14 @@ func (r *ReconcileAccountClaim) finalizeAccountClaim(reqLogger logr.Logger, acco
 			return err
 		}
 		// Cleanup BYOC secret
-		err = r.removeBYOCSecretFinalizer(accountClaim)
-		if err != nil {
+		secretErr := r.removeBYOCSecretFinalizer(accountClaim)
+		if secretErr != nil {
 			reqLogger.Error(err, "Failed to remove BYOC iamsecret finalizer")
-			return err
+			return secretErr
 		}
 
+		// Here we are returning nil, instead of a potential err,
+		// as we only want to block if it's non-CCS where we can't cleanup.
 		return nil
 	}
 
@@ -79,7 +83,6 @@ func (r *ReconcileAccountClaim) finalizeAccountClaim(reqLogger logr.Logger, acco
 	}
 
 	awsClient, err := r.awsClientBuilder.GetClient(controllerName, r.client, awsClientInput)
-
 	if err != nil {
 		connErr := fmt.Sprintf("Unable to create aws client for region %s", clusterAwsRegion)
 		reqLogger.Error(err, connErr)
@@ -90,6 +93,7 @@ func (r *ReconcileAccountClaim) finalizeAccountClaim(reqLogger logr.Logger, acco
 		err := r.client.Delete(context.TODO(), reusedAccount)
 		if err != nil {
 			reqLogger.Error(err, "Failed to delete BYOC account from accountclaim cleanup")
+			return err
 		}
 
 		// Cleanup BYOC secret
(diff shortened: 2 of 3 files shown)


mrWinston (commented):
Looks good to me and the changes make sense, though i'm not  100% confident i can grasp the scope of it. Are the functions you've changed covered by the unit tests? even these small changes lead to slightly different behaviour during the reconcile, which might be what we want.

pkg/controller/accountclaim/reuse.go
```diff
@@ -36,13 +36,13 @@ func (r *ReconcileAccountClaim) finalizeAccountClaim(reqLogger logr.Logger, acco
...
-		if err != nil {
+		secretErr := r.removeBYOCSecretFinalizer(accountClaim)
+		if secretErr != nil {
 			reqLogger.Error(err, "Failed to remove BYOC iamsecret finalizer")
-			return err
+			return secretErr
 		}
 
-		return nil
+		return err
```
iamkirkbater, line 45: I think we still want to `return nil` here, as we only want to block if it's not CCS where we can't cleanup.

iamkirkbater (changes requested):

pkg/controller/accountclaim/reuse.go
dkeohane: Fair point - I'll update it with a comment then explaining why we're returning nil instead of the err

pkg/controller/accountclaim/accountclaim_controller.go
```diff
@@ -304,6 +304,8 @@ func (r *ReconcileAccountClaim) handleAccountClaimDeletion(reqLogger logr.Logger
 				reqLogger.Error(accountErr, "Failed updating account status for failed reuse")
 				return err
 			}
+
+			return err
```
dkeohane, line 308: Now that we're _actually_ returning an error here, if `resetAccountSpecStatus` worked as expected but the claim wasn't cleaned up, we'll attempt to call this function again but the accountClaim.Spec.AccountLink will now be "", so we will never attempt to clean the account again and simply remove the finalizer

dkeohane: Actually, scratch that, resetAccountSpecStatus doesn't affect the accountclaim, just the account.

pkg/controller/accountclaim/accountclaim_controller.go
```diff
@@ -304,6 +304,8 @@ func (r *ReconcileAccountClaim) handleAccountClaimDeletion(reqLogger logr.Logger
 				reqLogger.Error(accountErr, "Failed updating account status for failed reuse")
```
iamkirkbater, line 304: Looking at this a bit more, do we want to return err here and in the function above, or do we want to wrap err with a more descriptive error like we're doing for the isConflict error?

iamkirkbater: Like, we still want the original error, but we also want to know what happened inside of this error check.

pkg/controller/accountclaim/reuse.go
```diff
@@ -185,23 +189,24 @@ func (r *ReconcileAccountClaim) cleanUpAwsAccount(reqLogger logr.Logger, awsClie
...
 		}
 	}
 
 	// Return an error if we saw any errors on the awsErrors channel so we can make the reused account as failed
 	if cleanUpStatusFailed {
-		cleanUpStatusFailedMsg := "Failed to clean up AWS account"
-		err := errors.New(cleanUpStatusFailedMsg)
+		cleanUpStatusFailedMsg := "failed to clean up AWS account"
 		reqLogger.Error(err, cleanUpStatusFailedMsg)
+		return err
```
dkeohane, line 209: As a result of moving the err var creation to outside the loop on line 192, this returned err will be the last err we retrieved from the cleangoFunc which failed

dkeohane, line 209: As a result of moving the err var creation to outside the loop on line 192, this returned err will be the last err we retrieved from the cleangoFunc which failed

pkg/controller/accountclaim/reuse.go
```diff
@@ -238,7 +243,7 @@ func (r *ReconcileAccountClaim) cleanUpAwsAccountSnapshots(reqLogger logr.Logger
 
 		_, err = awsClient.DeleteSnapshot(&deleteSnapshotInput)
 		if err != nil {
-			delError := fmt.Sprintf("Failed deleting EBS snapshot: %s", *snapshot.SnapshotId)
+			delError := errors.Wrap(err, fmt.Sprintf("Failed deleting EBS snapshot: %s", *snapshot.SnapshotId)).Error()
```
dkeohane, line 246: Decided to wrap these in order to gain visibility into what was coming from the AWS side of things

dkeohane, line 246: Decided to wrap these in order to gain visibility into what was coming from the AWS side of things

pkg/controller/accountclaim/reuse.go
dkeohane: Adding the return err here is possibly the most important aspect of this PR - previously if we encountered an error while cleaning up AWS resources, we would still return nil and remove the finalizer

dkeohane: Adding the return err here is possibly the most important aspect of this PR - previously if we encountered an error while cleaning up AWS resources, we would still return nil and remove the finalizer

NautiluX (commented):
/lgtm
/hold merge when ready

pkg/controller/accountclaim/reuse.go
```diff
@@ -185,23 +189,24 @@ func (r *ReconcileAccountClaim) cleanUpAwsAccount(reqLogger logr.Logger, awsClie
...
 		select {
 		case msg := <-awsNotifications:
 			reqLogger.Info(msg)
 		case errMsg := <-awsErrors:
-			err := errors.New(errMsg)
+			err = errors.New(errMsg)
 			reqLogger.Error(err, errMsg)
 			cleanUpStatusFailed = true
 		}
 	}
```
iamkirkbater, line 203: I think instead of `var err error` here we need something like `var cleanupErrors[error]{}` and append to that list, and then check if that list is failed.

Otherwise if there are multiple errors we will only see the one error message.

Functionally, this will still get what we're looking for though it might just make future debugging harder in those situations, so I won't block the PR for this request.

iamkirkbater (approved):
Overall LGTM.  One small comment we can address in a different PR in order to not block this one.

iamkirkbater:
Can we squash this before we merge, please.

PR #668, opened 2022-01-31
add function/tests for probing secret

katherinelc321:
This PR adds the function to validate account secrets with its respective unit test. [OSD-9298](https://issues.redhat.com/browse/OSD-9298)

PR #678, opened 2022-02-28
probe secret

katherinelc321:
This PR probes the osd-admin secret in accounts and heal them if they're invalid https://issues.redhat.com/browse/OSD-9298

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -118,6 +118,23 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Define a kubeClient for any processes that need to run during operator startup or independent routines to use
+	// We should avoid using this kubeClient except for when necessary and utilize the operator-sdk provided client as much as possible.
+	// The operator-sdk kube client provides a level of caching that we don't get with building our own this way.
+	kubeClient, err := client.New(cfg, client.Options{})
+	if err != nil {
+		log.Error(err, "Failed to create a kubernetes client")
+		os.Exit(1)
+	}
+
+	errors := utils.InitControllerMaxReconciles(kubeClient)
+	if len(errors) > 0 {
+		log.Info("There was at least one error initializing controller max reconcile values.")
+		for _, err := range errors {
+			log.Error(err, "")
+		}
+	}
+
 	log.Info("Registering Components.")
 
 	// Setup Scheme for all resources
@@ -170,13 +187,6 @@ func main() {
 	// to stop work. This channel can also be used to signal routines to complete any cleanup
 	// work
 	stopCh := signals.SetupSignalHandler()
-
-	// Define a kubeClient for any processes that need to run during operator startup or independent routines to use
-	kubeClient, err := client.New(cfg, client.Options{})
-	if err != nil {
-		log.Error(err, "Failed to create a kubernetes client")
-		os.Exit(1)
-	}
 
 	// Initialize our ConfigMap with default values if necessary.
 	initOperatorConfigMapVars(kubeClient)
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -23,7 +23,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/controller"
+
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -104,7 +104,7 @@ func newReconciler(mgr manager.Manager) reconcile.Reconciler {
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
 func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	// Create a new controller
-	c, err := controller.New("account-controller", mgr, controller.Options{Reconciler: r})
+	c, err := utils.NewControllerWithMaxReconciles(log, controllerName, mgr, r)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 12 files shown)


config/config.go
```diff
@@ -81,3 +82,16 @@ func GetIAMArn(awsAccountID, awsResourceType, awsResourceID string) (arn string)
 	arn = strings.Join([]string{"arn:", awsAPI, ":iam::", awsAccountID, ":", awsResourceType, "/", awsResourceID}, "")
 	return
 }
+
+func SetHiveSHardName(configMap *corev1.ConfigMap) error {
```
dkeohane, line 86:
```suggestion
func SetHiveShardName(configMap *corev1.ConfigMap) error {
```

cmd/manager/main.go
```diff
@@ -118,6 +118,21 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Define a kubeClient for any processes that need to run during operator startup or independent routines to use
```
dkeohane, line 121: From: https://github.com/openshift/aws-account-operator/pull/668#discussion_r810034883
```suggestion
	// Define a kubeClient for any processes that need to run during operator startup or independent routines to use
	// We should avoid using this kubeClient except for when necessary and utilize the operator-sdk provided client as much as possible.
	// The operator-sdk kube client provides a level of caching that we don't get with building our own this way.
```

config/config.go
```diff
@@ -81,3 +82,16 @@ func GetIAMArn(awsAccountID, awsResourceType, awsResourceID string) (arn string)
 	arn = strings.Join([]string{"arn:", awsAPI, ":iam::", awsAccountID, ":", awsResourceType, "/", awsResourceID}, "")
 	return
 }
+
+func SetHiveShardName(configMap *corev1.ConfigMap) error {
```
iamkirkbater, line 86: I don't see where we're calling this.  I think we need to explicitly call this in manager.go first.

iamkirkbater (changes requested):

pkg/controller/account/account_controller.go
```diff
@@ -344,6 +344,11 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 	}
 	regionAMIs := processConfigMapRegions(stringRegions)
 
+	awsAssumedRoleClient, creds, err := r.handleCreateAdminAccessRole(reqLogger, currentAcctInstance, awsSetupClient)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
```
iamkirkbater, line 351: Is there a reason this got moved up?  This seems to be breaking STS cluster initialization.

iamkirkbater: Ahh, figured it out.  Patch included in review comment.

iamkirkbater (changes requested):
This breaks during STS testing. `make test-sts` was able to catch this.  `make test-integration` runs through all of our integration tests.

Here's a patch of the changes I made to get this working:
```
diff --git a/pkg/controller/account/account_controller.go b/pkg/controller/account/account_controller.go
index 3b1220f4..1975f77b 100644
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -344,11 +344,6 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 	}
 	regionAMIs := processConfigMapRegions(stringRegions)
 
-	awsAssumedRoleClient, creds, err := r.handleCreateAdminAccessRole(reqLogger, currentAcctInstance, awsSetupClient)
-	if err != nil {
-		return reconcile.Result{}, err
-	}
-
 	// Account init for both BYOC and Non-BYOC
 	if currentAcctInstance.ReadyForInitialization() {
 		reqLogger.Info("initializing account", "awsAccountID", currentAcctInstance.Spec.AwsAccountID)
@@ -397,6 +392,11 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 			return reconcile.Result{Requeue: true}, r.Client.Update(context.TODO(), currentAcctInstance)
 		}
 
+		awsAssumedRoleClient, creds, err := r.handleCreateAdminAccessRole(reqLogger, currentAcctInstance, awsSetupClient)
+		if err != nil {
+		}
+
 		// Use the same ID applied to the account name for IAM usernames
 		iamUserUHC := fmt.Sprintf("%s-%s", iamUserNameUHC, currentAcctInstance.Labels[awsv1alpha1.IAMUserIDLabel])
 		secretName, err := r.BuildIAMUser(reqLogger, awsAssumedRoleClient, currentAcctInstance, iamUserUHC, request.Namespace)
@@ -430,14 +430,27 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 	}
 
 	if currentAcctInstance.IsReady() {
+		roleToAssume := getAssumeRole(currentAcctInstance)
+		awsClient, _, err := r.assumeRole(reqLogger, currentAcctInstance, awsSetupClient, roleToAssume, "")
+		if err != nil {
+			reqLogger.Error(err, "failed building BYOC client from assume_role")
+			// We don't want to error here as erroring will requeue and we will end in an
+			// infinite loop.  So we just log the error and exit.
+			// TODO maybe there's a better way to handle this?
+		}
 
 		// Check that secret is valid and reheal it if not
 		iamUserUHC := fmt.Sprintf("%s-%s", iamUserNameUHC, currentAcctInstance.Labels[awsv1alpha1.IAMUserIDLabel])
 
 		reqLogger.Info("probing account secret")
-		err := r.ProbeSecret(reqLogger, currentAcctInstance, awsAssumedRoleClient, iamUserUHC, request.Namespace)
+		err = r.ProbeSecret(reqLogger, currentAcctInstance, awsClient, iamUserUHC, request.Namespace)
 		if err != nil {
-			return reconcile.Result{}, err
+			reqLogger.Error(err, "failed to probe secret")
+			// We don't want to error here as erroring will requeue and we will end in an
+			// infinite loop.  So we just log the error and exit.
+			// TODO maybe there's a better way to handle this?
 		}
 	}
 
```

PR #679, opened 2022-03-02
updating documentation around AccountClaim customTags

macgregor:
updating docs to capture some past and recent decisions around `customTags` which are passed to AAO via the `AccountClaim` that we add to AWS resources that we create.

Related to:
* [OSD-9949](https://issues.redhat.com/browse/OSD-9949)
* [OSD-10191](https://issues.redhat.com/browse/OSD-10191)


dkeohane (approved):
Thanks for the additions to the docs - much appreciated! :t-rex:

PR #688, opened 2022-04-05
Fix broken link to the manual STS setup.

bergmannf:
The link to the manual STS setup was pointing to non-existent file.


dkeohane (approved):
/lgtm

PR #691, opened 2022-04-12
[OSD-9894] Add condition when account is claimed.

bergmannf:
The condition is added when the Status.Claimed is set to true.

This happens for regular and byoc accounts.

Fixes OSD-9894.

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -286,8 +286,7 @@ func (r *ReconcileAccount) Reconcile(request reconcile.Request) (reconcile.Resul
 
 		// Update account Status.Claimed to true if the account is ready and the claim link is not empty
 		if currentAcctInstance.IsReadyUnclaimedAndHasClaimLink() {
-			currentAcctInstance.Status.Claimed = true
-			return reconcile.Result{}, r.statusUpdate(currentAcctInstance)
+			return reconcile.Result{}, ClaimAccount(r, currentAcctInstance)
 		}
 
 		// see if in creating for longer than default wait time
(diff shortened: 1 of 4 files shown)


pkg/controller/account/account_controller.go
```diff
@@ -978,6 +978,24 @@ func CreateAccount(reqLogger logr.Logger, client awsclient.Client, accountName,
 	return accountStatus, nil
 }
 
+func ClaimAccount(currentAcctInstance *awsv1alpha1.Account) {
```
dkeohane, line 981: Could you add the function call to `statusUpdate` into this function as this is generally the next call made after claiming the account. The we could just return the result from here

dkeohane (commented):
Awesome first ticket - thanks for adding and improving the documentation, left one minor comment! :)

fahlmant (commented):
Been a while since I worked on this controller, but I know the logic in the Account controller and the AccountPool controller are highly dependent on the status of the Account CR. The Pool for example looks at Accounts in specific states (or Accounts NOT in specific states) to keep the count. So make sure the controller logic for the Pool, Account and Claim handle this new state

bergmannf:
I checked the `AccountPoolController`, and it seems that none of the checks are looping through the `conditions` (which is all that is changed in this PR) - even though something like `HasBeenClaimedAtLeastOnce` could probably use it as a fallback.

I also didn't see any logic in the `ClaimController` that uses the `conditions`, so I think it should be fine.

pkg/controller/account/account_controller.go
bergmannf: Ok updated it, so the ClaimAccount will also do the statusUpdate and return the error (if any).

bergmannf:
Reran the integration tests and everything still passed, anything else that blocks this PR from proceeding?

dastergon:
@fahlmant WDYT?

iamkirkbater:
This LGTM.  @bergmannf is correct, we shouldn't be relying on the conditions behind the status but the status itself, so this should be relatively safe to add in.  I'm going to LGTM and /hold, please feel free to remove the hold and deploy this to stage/int at your convenience.

/hold

PR #692, opened 2022-04-20
Disable secret probe

jharrington22:
This PR disables the secret probe as it significantly increases the reconcile time in production. We should investigate doing this kind of operation outside of the reconcile loop.

For the moment this PR effectively disables changes in PR #678

### DIFF
--- a/pkg/controller/account/account_controller.go
+++ b/pkg/controller/account/account_controller.go
@@ -71,6 +71,9 @@ const (
 	iamUserNameUHC               = "osdManagedAdmin"
 
 	controllerName = "account"
+	// probeSecretEnabled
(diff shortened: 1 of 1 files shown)


pkg/controller/account/account_controller.go
```diff
@@ -71,6 +71,9 @@ const (
 	iamUserNameUHC               = "osdManagedAdmin"
 
 	controllerName = "account"
+	// probeSecretEnabled
+	// Currently disabled becuase it significantly increases reconcile time in production
```
iamkirkbater, line 75:
```suggestion
	// Currently disabled because it significantly increases reconcile time in production
```

PR #693, opened 2022-04-22
Update setup aws script

bergmannf:
The script did miss the step to create the correct policy for the `AccessRole` that is created.

Without the policy the Role can not `describeRegions` in `ec2` and will fail in the `STS` scenario.

### DIFF
--- a/hack/scripts/aws/setup_aws_accounts.sh
+++ b/hack/scripts/aws/setup_aws_accounts.sh
@@ -3,7 +3,7 @@
 command -v aws >/dev/null 2>&1 || { echo >&2 "Script requires aws but it's not installed.  Aborting."; exit 1; }
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
-cd $DIR
+cd "$DIR" || exit
 
 usage() {
     cat <<EOF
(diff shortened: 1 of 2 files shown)


hack/scripts/aws/setup_aws_accounts.sh
```diff
@@ -141,10 +145,13 @@ do
   ((j=2**i))
   echo "Access role Arn not ready yet."
   echo "sleeping $j"
-  sleep $j
+  sleep "$j"
   ((i=i+1))
 done
 
+POLICY_ARN=$(aws iam create-policy --policy-name "minimum-permissions-access-role${ID}" --policy-document "file://${PWD}/setup-aws-policies/AccessRolePolicy.json" | jq -r ".Policy.Arn")
+aws iam attach-role-policy --policy-arn "${POLICY_ARN}" --role-name "AccessRole${ID}"
```
dkeohane, line 153: Might need a sleep or something before this? Is there any delay between the policy being created and being able to use it?

dkeohane, line 153: Might need a sleep or something before this? Is there any delay between the policy being created and being able to use it?

bergmannf: I only tested it once via CLI, and it *felt* like it wasn't asynchronous, but actually synchronous.

I could check to see if there is a simple call to `get` the policy before proceeding though.

dkeohane (approved):
/lgtm

Thanks for these updates! :tractor:

PR #694, opened 2022-04-22
How to find correct OSD_STAGING_1_OU_BASE_ID

bergmannf:
This is needed to have the AAO running and it can differ, depending
where osdctl assigns the account.


dkeohane (approved):
/lgtm

Thanks for updating the docs

PR #701, opened 2022-05-27
[OSD-11085] [WIP] Add KMS key to AccountClaim.

bergmannf:
Add a new field to the account claim that holds a customer supplied KMS key to encrypt the region init instances with.


test/integration/api/create_account.sh
```diff
@@ -20,6 +20,8 @@ while true
 do STATUS=$(oc get account ${OSD_STAGING_1_ACCOUNT_CR_NAME_OSD} -n ${NAMESPACE} -o json | jq -r '.status.state');
     if [ "$STATUS" == "Ready" ]; then
         break
+    elif [ "$STATUS" == "InitializingRegions" ]; then
```
bergmannf, line 23: I had to add this, as otherwise the accountclaim would not be linked to the account.

So I have to create the claim first, then the account, which seems to immediately force the account into `InitializingRegions`.

Maybe someone knows a better way to do this.

pkg/controller/account/ec2.go
```diff
@@ -46,12 +46,23 @@ func (r *ReconcileAccount) InitializeSupportedRegions(reqLogger logr.Logger, acc
 	vCPUQuota, _ := r.getDesiredVCPUValue(reqLogger)
 	reqLogger.Info("retrieved desired vCPU quota value from configMap", "quota.vcpu", vCPUQuota)
 
+	// FIXME: Is this the right place to get it - in the reconcile loop for accounts there are two call to this initialize, so should reduce weird copy-paste errors in the future.
+	// Will however hit the cache for the claim, as non-sts already retrieved the claim before.
+	var kmsKeyId string
+	accountClaim, accountClaimError := r.getAccountClaim(account)
+	if accountClaimError != nil {
+		reqLogger.Error(accountClaimError, "Could not retrieve account claim for account.", "account", account.Name)
```
dkeohane, line 54: We may not want to be logging an error here? Not all account crs created will have an accountclaim attached at this point, we could just be filling the accountpool, so logging an error may not be necessary

pkg/controller/account/ec2.go
```diff
@@ -46,12 +46,23 @@ func (r *ReconcileAccount) InitializeSupportedRegions(reqLogger logr.Logger, acc
 	vCPUQuota, _ := r.getDesiredVCPUValue(reqLogger)
 	reqLogger.Info("retrieved desired vCPU quota value from configMap", "quota.vcpu", vCPUQuota)
 
+	// FIXME: Is this the right place to get it - in the reconcile loop for accounts there are two call to this initialize, so should reduce weird copy-paste errors in the future.
```
dkeohane, line 49: I don't like seeing FIXME in docstrings, we should ideally track this in a ticket elsewhere, or look at resolving it before the PR is merged

asyncRegionInit is only called in initializeRegions which hits 2 flows, but they never overlap, so I don't think we'll be calling this twice in a loop

dkeohane (commented):
Gave the kms code changes a quick lookover but didn't read through the integration + unit tests yet, will hopefully check the rest at some point in the next day or so

pkg/controller/account/ec2.go
bergmannf: Oh yea that was more like a note for reviewers, if this seems to be something we would like to do elsewhere, so I think I would just remove the FIXME + comment.

pkg/controller/account/ec2_test.go
```diff
@@ -0,0 +1,301 @@
...
+			},
+		},
+		ImageId:      aws.String("fakeami"),
+		InstanceType: aws.String("t2.micro"),
+		MaxCount:     aws.Int64(1),
+		MinCount:     aws.Int64(1),
+		TagSpecifications: []*ec2.TagSpecification{
+			{
+				ResourceType: &awsv1alpha1.InstanceResourceType,
+				Tags: []*ec2.Tag{
```
dkeohane, line 49: minor: Could declare the tags array outside this assignment to cut down on length of the builder

pkg/controller/account/ec2_test.go
```diff
@@ -0,0 +1,301 @@
...
+		customerTags        []awsclient.AWSTag
+		customerKmsKeyId    string
+		instanceInput       *ec2.RunInstancesInput
+		instanceOutput      *ec2.Reservation
+		instanceOutputError error
+	}
+	tests := []struct {
+		name    string
+		args    args
+		want    string
```
dkeohane, line 133: minor: We generally use expected instead of want

pkg/controller/account/ec2_test.go
```diff
@@ -0,0 +1,301 @@
...
+		regions    []awsv1alpha1.AwsRegions
+		creds      *sts.AssumeRoleOutput
+		regionAMIs map[string]awsv1alpha1.AmiSpec
+	}
+	tests := []struct {
+		name   string
+		fields fields
+		args   args
+	}{
+		// TODO: Add test cases.
```
dkeohane, line 262: remove or add more ;)

test/integration/tests/validate_kms_key.sh
```diff
@@ -0,0 +1,91 @@
...
+}
+
+accClaim=$(oc get accountclaim "$KMS_CLAIM_NAME" -n "$KMS_NAMESPACE_NAME" -o json)
+
+if [[ $(jq -r '.spec.kmsKeyId | length' <<< "$accClaim") -lt 1 ]]; then
+    echo "KmsKeyID is empty"
+    exit 1
+fi
+
+# TODO: Get the ec2 instances that are started and verify they are encrypted
```
dkeohane, line 31: remove

test/integration/tests/validate_kms_key.sh
```diff
@@ -0,0 +1,91 @@
...
+    exit 1
+fi
+
+assume_account
+
+TARGET_KEY_ID=$(aws kms list-aliases --region "${AWS_REGION}" | jq -r '.Aliases[] | select(.AliasName == "alias/aao-test-key").TargetKeyId')
+
+# Check for 60 seconds if an encrypted volume with the correct tag occurs in the region we are checking.
+# If this happens the tests is a success, otherwise something likely went wrong.
+echo "Waiting for encrypted volumes to appear using key ${TARGET_KEY_ID}"
```
dkeohane, line 46: Maybe indicate in the echo that you'll wait up to 60 seconds for better user visibility

dkeohane (commented):
Dug a bit deeper this time round, but still haven't ran the integration tests :)

iamkirkbater:
This works well on my local instance!  Thanks for updating with the new patch.  I got it working by deleting the CRC instance and starting a new one, starting from scratch, so something was wrong with my CRC, as it wasn't picking up the change even after running `make predeploy` again.

Let's watch the rollout for stage to see if there's any issues here that are similar.  Hopefully OLM won't have a fit about it.

/lgtm

PR #704, opened 2022-07-18
Osd 10116 verify account tags

AlexVulaj:
This PR includes changes from https://github.com/openshift/aws-account-operator/pull/702 and should not be merged until it's taken care of first for an accurate diff.

### DIFF
--- a/pkg/controller/validation/account_validation_controller.go
+++ b/pkg/controller/validation/account_validation_controller.go
@@ -23,7 +23,8 @@ import (
 
 var log = logf.Log.WithName("controller_accountvalidation")
 
-var account_move_enabled = false
+var accountMoveEnabled = false
+var accountTagEnabled = false
 
 const (
 	controllerName = "accountvalidation"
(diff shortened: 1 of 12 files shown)


pkg/controller/validation/account_validation_controller.go
```diff
@@ -150,6 +166,35 @@ func MoveAccount(awsAccountId string, client awsclient.Client, targetOU string,
...
+	}
+
+	resp, err := client.ListTagsForResource(listTagsForResourceInput)
+	if err != nil {
+		return err
+	}
+
+	for _, tag := range resp.Tags {
+		if ownerKey == *tag.Key {
+			if shardName != *tag.Value {
```
macgregor, line 182: seems possible for shardName to be unset based on the logic above which gets it from the AAO config map:
	hiveName, ok := configMap.Data["shard-name"]
	if !ok {
		log.Error(err, "shard-name key not available in configmap")
	}
	reconciler.shardName = hiveName

The most likely reason for this would be misconfiguration, however, that misconfiguration would also cause a bunch of noise here. That could be considered a feature, but im wondering if we want to skip tag validation based on shard name if its nil/empty. It doesnt look like we are attempting to update the tag if its different (which would result in information loss if things are misconfigured) so this is more of a what do you think comment than a request.

PR #708, opened 2022-07-23
osd-12227 update prow ci operator deploy script

rajdeepc2792:
Jira Story :- https://issues.redhat.com/browse/OSD-12227

Motivation around this script is to build and deploy AAO operator with supporting artifacts on a cluster claimed by Prow CI pipeline.

The script is also intend to perform similar build and deployment on the local cluster be it CRC or staging demo cluster.

At the end the script also perform cleanup of the deployed operator.
Before this cleanup the trigger for testing existing integration tests will be introduced later.


hack/scripts/prow-ci-operator-deploy.sh
```diff
@@ -1,4 +1,205 @@
...
+            export STS_ROLE_ARN=$(cat /tmp/secret/aao-aws-creds/STS_ROLE_ARN)
+            export STS_JUMP_ARN=$(cat /tmp/secret/aao-aws-creds/STS_JUMP_ARN)
+            export OSD_STAGING_1_OU_ROOT_ID=$(cat /tmp/secret/aao-aws-creds/OSD_STAGING_1_OU_ROOT_ID)
+            export OSD_STAGING_1_OU_BASE_ID=$(cat /tmp/secret/aao-aws-creds/OSD_STAGING_1_OU_BASE_ID)
+            export SUPPORT_JUMP_ROLE=$(cat /tmp/secret/aao-aws-creds/SUPPORT_JUMP_ROLE)
+            export STS_JUMP_ROLE=$(cat /tmp/secret/aao-aws-creds/STS_JUMP_ROLE)
+            export OPERATOR_ACCESS_KEY_ID=$(cat /tmp/secret/aao-aws-creds/OPERATOR_ACCESS_KEY_ID)
+            export OPERATOR_SECRET_ACCESS_KEY=$(cat /tmp/secret/aao-aws-creds/OPERATOR_SECRET_ACCESS_KEY)
+        else
+            source .envrc
```
bergmannf, line 94: Might want to check the file actually exists, before attempting to source it and error out if it does not.

hack/scripts/prow-ci-operator-deploy.sh
```diff
@@ -1,4 +1,205 @@
...
+}
+
+function cleanupPre {
+    cleanup    
+}
+
+function getEnvVariables {
+    if [[ -z $USE_ENVRC ]];
+        then
+            export FAKE_KEY_VALUE=$(cat /tmp/secret/aao-aws-creds/FAKE_KEY_VALUE)
```
bergmannf, line 81: Where do all these files in `/tmp` come from - are they created somewhere?

Otherwise might be worth adding some documentation about this case.

hack/scripts/prow-ci-operator-deploy.sh
```diff
@@ -1,4 +1,205 @@
...
+    fi
+}
+
+function deployOperator {
+  echo -e "\nDEPLOYING OPERATOR\n"
+
+  make prow-ci-deploy-image OPERATOR_IMAGE_URI=$BUILD_CONFIG
+}
+
+function waitForDeployment {
```
bergmannf, line 146: Might be worth to check if just using `kubectl wait` works here - might make the code a lot shorter.

hack/scripts/prow-ci-operator-deploy.sh
rajdeepc2792: These files in `/tmp` are the secrets we created on vault UI, using following doc:-
https://docs.ci.openshift.org/docs/how-tos/adding-a-new-secret-to-ci/

hack/scripts/prow-ci-operator-deploy.sh
rajdeepc2792: `wait` works well as a replacement. Please review the below modified code, please also find the logs for successful run here:-  https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_aws-account-operator/708/pull-ci-openshift-aws-account-operator-master-integration-test/1551988141563318272/artifacts/integration-test/test/build-log.txt

hack/scripts/prow-ci-operator-deploy.sh
rajdeepc2792: Added the above document link as a comment before using secret values.

hack/scripts/prow-ci-operator-deploy.sh
rajdeepc2792: Added a check for .envrc before usage.

hack/scripts/prow-ci-operator-deploy.sh
bergmannf: That's great - that makes it a lot clearer when and why those are used.

Makefile
```diff
@@ -473,6 +473,21 @@ clean-operator: ## Clean Operator
...
 prow-ci-deploy: ## Triggers prow-ci build and deploy operator bash script
 	hack/scripts/prow-ci-operator-deploy.sh
+
+.PHONY: cluster-ci-entrypoint
+cluster-ci-entrypoint: ## Triggers prow-ci build and deploy operator bash script
+	hack/scripts/prow-ci-operator-deploy.sh --use-envrc --skip-cleanup --is-staging -n $(OPERATOR_NAMESPACE)
+
+.PHONY: local-ci-entrypoint
+local-ci-entrypoint: ## Triggers prow-ci build and deploy operator bash script
+	hack/scripts/prow-ci-operator-deploy.sh --use-envrc --skip-cleanup --is-local -n $(OPERATOR_NAMESPACE)
```
macgregor, line 490: consider renaming `hack/scripts/prow-ci-operator-deploy.sh` to something like  `hack/scripts/integration-test-bootstrap.sh`

Makefile
```diff
@@ -473,6 +473,21 @@ clean-operator: ## Clean Operator
 	oc delete awsfederatedaccountaccess --all -n ${NAMESPACE}
 	oc delete awsfederatedrole --all -n ${NAMESPACE}
 
+.PHONY: prow-ci-predeploy
+prow-ci-predeploy: predeploy-aws-account-operator deploy-aws-account-operator-credentials create-ou-map
+	@ls deploy/*.yaml | grep -v operator.yaml | xargs -L1 oc apply -f
+
 .PHONY: prow-ci-deploy
 prow-ci-deploy: ## Triggers prow-ci build and deploy operator bash script
```
macgregor, line 481: is this still needed or are you just keeping it around until you can update the make target that our prow pipeline calls?

macgregor (changes requested):
Looking pretty good. Some naming suggestions and like one relatively big refactor about using "profiles" rather than combinations of cli flags to organize the different use cases we have for the script (which i could probably be talked down from pretty easily if you dont like the idea)

hack/scripts/prow-ci-operator-deploy.sh
```diff
@@ -1,4 +1,283 @@
...
+    removeDockerfileSoftLink
+}
+
+function cleanupPre {
+    cleanup    
+}
+
+function getEnvVariables {
+    if [[ -z $USE_ENVRC ]]; then
+        ## Prow CI uses existing secrets:- https://docs.ci.openshift.org/docs/how-tos/adding-a-new-secret-to-ci/
```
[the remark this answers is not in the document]
rajdeepc2792: @bergmannf pointed the need of comment here!!

Makefile
rajdeepc2792: Yes, keeping it so that prow pipeline can trigger pipeline.
I will add a similar operating target with more appropriate name. And when this PR is merged, change the target in prow configuration.
Target name that I will use:- `prow-ci-entrypoint`

hack/scripts/prow-ci-operator-deploy.sh
```diff
@@ -1,4 +1,283 @@
...
+        waitForDeployment
+        if [ -z $IS_STAGING ]; then
+            installProwCIDependencies
+        fi
+        make ci-int-tests
+    else
+        make deploy-local &
+        localOperator=$!
+        make ci-int-tests
+        kill $localOperator
```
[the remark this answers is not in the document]
rajdeepc2792: So, this kill is only for the operator process, which is started by the script itself. So IMO, whatever is started or created by the script should be removed. When running locally, currently script assumes that there's no existing operator process running.
In local, operator runs as a os process.

hack/scripts/prow-ci-operator-deploy.sh
```diff
@@ -1,4 +1,283 @@
...
+        then
+            echo -e "\nno $NAMESPACE namespace found\n"
+            return
+	fi
+
+    $OC delete namespace $NAMESPACE || true
+    removeDockerfileSoftLink
+}
+
+function cleanupPre {
```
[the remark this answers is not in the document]
rajdeepc2792: I started maintaining the flow of cleanup as `pre` and `post` considering there can be things which need to be done for one and not for other.
For Eg. - printing of logs is done as post-cleanup step but not pre-cleanup.

Just to make the flow visible kept it till the end. Not any other reason.

I can remove it if you think it should be removed.

hack/scripts/prow-ci-operator-deploy.sh
```diff
@@ -1,4 +1,283 @@
...
+function cleanup {
+    echo -e "\nPERFORMING CLEANUP\n"
+    ##  Add step to remove CRDs
+    if [[ $($OC get namespace $NAMESPACE --no-headers | wc -l) == 0 ]];
+        then
+            echo -e "\nno $NAMESPACE namespace found\n"
+            return
+	fi
+
+    $OC delete namespace $NAMESPACE || true
```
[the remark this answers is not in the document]
rajdeepc2792: I will need to wait for it, as I can't move ahead to the deployments if not certain that namespace is deleted.

hack/scripts/prow-ci-operator-deploy.sh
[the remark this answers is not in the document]
macgregor: Oh, I see, cleanup is called by both cleanupPre and cleanupPost calls. Ok, i think thats fine, maybe just add some comments to telegraph to casual readers that this is intentional

hack/scripts/prow-ci-operator-deploy.sh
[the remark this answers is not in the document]
macgregor: I think its a little confusing/unexpected to have explicit pre/post hooks for resource cleanup but then have some resource cleanup happen outside those functions. So, if we have/want to ensure the process exits explicitly by killing it, I think that should happen in the cleanup. cleanupPost being a trap would also ensure the child is killed if `make ci-int-tests` exits in a weird way like a user hitting ctrl+c: https://stackoverflow.com/questions/360201/how-do-i-kill-background-processes-jobs-when-my-shell-script-exits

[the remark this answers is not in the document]
macgregor: the trap part is the more important bit than the $SKIP_CLEANUP flag

PR #709, opened 2022-07-25
[OSD-9106] add SLI for capturing accountclaim pending duration

dastergon:
This PR introduces two histograms that capture the pending duration of account claims in ccs and non-ccs environments.

The goal is to use these histograms as SLIs and create an SLO upon our needs. The histograms have buckets spanning from 60 to 600s. There's a native +Inf bucket that gets created automatically to capture the values > our upper bound (600) .

Card: https://issues.redhat.com/browse/OSD-9106


mrWinston:
+1 on the implementation, but i do have a conceptual question:
Is there a reson behind implementing this as two separate metrics for ccs and non-ccs accountclaims vs a ccs label on a single metrics?

Having a single metric with a ccs label will make queries across both ccs and non-ccs accountclaim significantly easier. Also, from a semantic point of view, it's just a property of the same thing, so having it under the same name makes more sense to me.

dastergon:
@mrWinston I follow the same pattern as our previously implemented SLIs for consistency.

mrWinston:
@dastergon Yes, i see. That makes sense then :)

dastergon:
@NautiluX can you PTAL?

pkg/controller/accountclaim/accountclaim_controller.go
```diff
@@ -162,6 +163,17 @@ func (r *ReconcileAccountClaim) Reconcile(request reconcile.Request) (reconcile.
...
+
+	if accountClaim.Status.State == awsv1alpha1.ClaimStatusPending {
+		now := metav1.Now()
+		unclaimedCondition := controllerutils.FindAccountClaimCondition(accountClaim.Status.Conditions, awsv1alpha1.AccountUnclaimed)
+		if unclaimedCondition != nil {
+			pendingDuration := now.Sub(unclaimedCondition.LastTransitionTime.Time)
+			localmetrics.Collector.SetAccountClaimPendingDuration(isCCS, pendingDuration.Seconds())
+		}
+	}
+
```
iamkirkbater, line 176: I don't think this is getting called for CCS accounts, see line 156 above, we're returning if it's a BYOC account.

I think we should abstract this to a function and utilize it both here on this line as well as around line 410 where we're requeueing while waiting.

pkg/controller/accountclaim/accountclaim_controller.go
```diff
@@ -162,6 +163,17 @@ func (r *ReconcileAccountClaim) Reconcile(request reconcile.Request) (reconcile.
 		return reconcile.Result{}, nil
 	}
 
+	isCCS := accountClaim.Spec.BYOCAWSAccountID != ""
+
+	if accountClaim.Status.State == awsv1alpha1.ClaimStatusPending {
+		now := metav1.Now()
+		unclaimedCondition := controllerutils.FindAccountClaimCondition(accountClaim.Status.Conditions, awsv1alpha1.AccountUnclaimed)
```
iamkirkbater, line 170: This AccountClaimCondition part is not working for CCS accounts.  I checked my `make create-ccs` accountclaim and the only status condition it has at all is `type: Claimed`.  I think we could simplify this to something like:

```go
if accountClaim.Status.State == awsv1alpha1.ClaimStatusPending {
  now := metav1.Now()
  pendingDuration := now.Sub(accountClaim.ObjectMeta.CreationTimestamp)
  localmetrics.Collector.SetAccountClaimPendingDuration(isCCS, pendingDuration.Seconds())
}
```

 I'm not sure if that's the correct way to use it offhand, but lets use the creation timestamp.  The SLO is from Creation Time to Ready, so it's our best indicator, as the accountClaim is created when the customer wants a cluster for both CCS and non-CCS and is Ready when it's done either way as well.

pkg/localmetrics/localmetrics.go
```diff
@@ -144,7 +146,18 @@ func NewMetricsCollector(store cache.Cache) *MetricsCollector {
...
 			Buckets:     []float64{5, 10, 20, 30, 60, 120, 240, 300, 480, 600},
 		}),
-
+		accountClaimPendingDuration: prometheus.NewHistogram(prometheus.HistogramOpts{
+			Name:        "aws_account_operator_account_claim_pending_duration_seconds",
+			Help:        "The duration for account claim cr in pending mode",
+			ConstLabels: prometheus.Labels{"name": operatorName},
+			Buckets:     []float64{60, 120, 240, 300, 600},
+		}),
+		ccsAccountClaimPendingDuration: prometheus.NewHistogram(prometheus.HistogramOpts{
```
iamkirkbater, line 155: Discussion: I'm not sure if cardinality will be an issue, but can we include labels in these metrics for the specific account and how long it's taking?  Labels per account would include which shard we're on as well as the accountclaim name for troubleshooting purposes.  

If cardinality is an issue when including the labels we could potentially work around it, but I wonder if the metric can clear after it's been claimed.

For example, if there's an account that for whatever reason sits in pending, we'd get the alert at 10 minutes, but let's say it's just AWS API slowness that day and it clears up at 12 minutes.  Would the metric clear itself, or do we need to adjust the alert to ensure that it's only creating alerts when they're pending and the time pending is incrementing?  I'm not sure if that makes sense or not.

iamkirkbater (changes requested):
Some small functionality changes are required to get this working with CCS accounts.  There's also one item open for discussion about metric implementation.

Looks good so far!

PR #712, opened 2022-07-27
Only validate accounts that are in a ready state

AlexVulaj:


pkg/controller/validation/account_validation_controller.go
```diff
@@ -239,12 +239,19 @@ func ValidateAccountOrigin(account awsv1alpha1.Account) error {
...
 	if !account.IsOwnedByAccountPool() {
-		log.Info("Will not validate account not owned by account pool", account)
+		log.Info("Will not validate account not owned by account pool", "account", account)
 		return &AccountValidationError{
 			Type: InvalidAccount,
 			Err:  errors.New("Account is not in an account pool"),
 		}
 	}
+	if !account.IsReady() {
+		log.Info("Will not validate account not in a ready state", "account", account)
```
iamkirkbater, line 249: Looking at these logs while testing #709, I'm not sure we need to print out the account here. The logger will already have a link to the account itself to reference, so I think we can save ourselves the log lines by removing the account object from these calls.  Can you do that for this whole controller since you're already working on this?

iamkirkbater (changes requested):
Overall looks great.  Thanks for taking this so quickly.

Can you please see the one comment I have and address that through the code, and also add a pull request description?

Thanks again!

pkg/controller/validation/account_validation_controller.go
AlexVulaj: Of course, I'll update that anywhere I see it;

PR #719, opened 2022-08-08
[run-int-test] [OSD-12630] Add AWS Cleanup handling for prow CI

rajdeepc2792:
EC2 instances and VPCs cleanup script after prow CI integration-test step is completed.
Jira Story:- https://issues.redhat.com/browse/OSD-12630

### DIFF
--- a/hack/scripts/cleanup-aws-resources.sh
+++ b/hack/scripts/cleanup-aws-resources.sh
@@ -54,8 +54,6 @@ function deleteVPCs {
     fi
 }
 
-echo -e "\nSTARTING AWS RESOURCES CLEANUP FOR CI\n"
-
 ## Currently the int tests only interact with us-east-1 region (specified in the AccountClaim CRs the tests create)
 ## for now we only clean up this region, but a more thorough implementation should probably check all regions or
 ## better yet just use the Account Shredder (https://github.com/openshift/aws-account-shredder) or whatever replaces it.
@@ -71,5 +69,3 @@ terminateEC2Instances
 waitForInstancesTermination
 getVPCs
 deleteVPCs
-
-echo -e "\nAWS RESOURCES CLEANUP COMPLETED\n"
--- a/hack/scripts/cleanup-aws-resources.sh
+++ b/hack/scripts/cleanup-aws-resources.sh
@@ -1,0 +1,75 @@
+# Generate creds based on Access Role & Jump Role for OSD_STAGING_1 & OSD_STAGING_2 respectively
+function generateRoleCreds {
+    if [ -z $1 -a -z $2 ]; then
+        echo "ERROR: Either Role ARN or AWS Account ID not provided as function argument."
+        return 1
+    fi
+    ROLE_ARN=$1
+    AWS_ACCOUNT_ID=$2
+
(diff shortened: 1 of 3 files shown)


hack/scripts/cleanup-aws-resources.sh
```diff
@@ -0,0 +1,71 @@
+# Generate creds based on Access Role & Jump Role for OSD_STAGING_1 & OSD_STAGING_2 respectively
+function generateRoleCreds {
+    if [ -z $1 -a -z $2 ]; then
+        echo "ERROR: Either Role ARN or AWS Account ID not provided as function argument."
+        return 1
+    fi
+    ROLE_ARN=$1
+    AWS_ACCOUNT_ID=$2
+
+    # TODO: remove profile
```
bergmannf, line 10: Is that comment outdated or which profile is it referring to?

hack/scripts/cleanup-aws-resources.sh
```diff
@@ -0,0 +1,71 @@
...
+REGION=us-east-1
+
+## aws-cli uses below credentials instead of profile for access
+export AWS_ACCESS_KEY_ID=$OPERATOR_ACCESS_KEY_ID
+export AWS_SECRET_ACCESS_KEY=$OPERATOR_SECRET_ACCESS_KEY
+
+generateRoleCreds $1 $2
+getEC2Instances
+terminateEC2Instances
+waitForInstancesTermination
```
bergmannf, line 69: I would probably just merge this into the `terminateEC2Instances` function - I don't think anyone will run the script in an `async` manner.

hack/scripts/cleanup-aws-resources.sh
```diff
@@ -0,0 +1,71 @@
...
+## for now we only clean up this region, but a more thorough implementation should probably check all regions or
+## better yet just use the Account Shredder (https://github.com/openshift/aws-account-shredder) or whatever replaces it.
+REGION=us-east-1
+
+## aws-cli uses below credentials instead of profile for access
+export AWS_ACCESS_KEY_ID=$OPERATOR_ACCESS_KEY_ID
+export AWS_SECRET_ACCESS_KEY=$OPERATOR_SECRET_ACCESS_KEY
+
+generateRoleCreds $1 $2
+getEC2Instances
```
bergmannf, line 67: Could probably also be moved into the `terminateEC2instances` function.

That would prevent using the `instanceIds` as a global variable - which is pretty much the only way to do it in bash, so I'd rather have the functions do all at once instead of relying on the correct call-order to set the variables.

hack/scripts/cleanup-aws-resources.sh
```diff
@@ -0,0 +1,71 @@
...
+terminateEC2Instances
+waitForInstancesTermination
+getVPCs
```
bergmannf, line 70: In the same vein as the get -> terminate -> wait requiring the correct call-order to have the variable set, I would just merge the get and delete functions as well.

rajdeepc2792: Merged the `getVPCs`  function to `deleteVPCs`

PR #721, opened 2022-08-10
[run-int-test][OSD-12227] update int-test bootstrap script for local & staging

rajdeepc2792:
Integration test bootstrap script updates for local and cluster flows.

- Sanity Check added for checking:-
   (1) Current Directory == Root Directory
   (2) AWS credential variables are set
 
 - backplane-cluster-admin access for staging flow.

### DIFF
--- a/hack/scripts/integration-test-bootstrap.sh
+++ b/hack/scripts/integration-test-bootstrap.sh
@@ -52,7 +52,9 @@ function parseArgs {
                 ;;
         esac
     done
-
+    if [ -z $NAMESPACE ]; then
+        NAMESPACE=aws-account-operator
+    fi
     echo "PROFILE=${PROFILE}"
     echo "NAMESPACE=${NAMESPACE}"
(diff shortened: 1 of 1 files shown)


hack/scripts/integration-test-bootstrap.sh
```diff
@@ -274,8 +296,13 @@ function profileProw {
 }
 
 function profileStage {
+    clusterUserName=$($OC whoami)
+    $OC --as backplane-cluster-admin adm policy add-cluster-role-to-user cluster-admin $clusterUserName
```
macgregor, line 300: looks good. just add a comment here about why youre doing this.

macgregor (changes requested):
minor request

hack/scripts/integration-test-bootstrap.sh
rajdeepc2792: Added the comment.

PR #724, opened 2022-08-24
Adds check if account is claimed before moving OU

iamkirkbater:
Adds a check to switch which OU the account should be in based on the claimed state of the account.

Work towards completing [OSD-9910](https://issues.redhat.com//browse/OSD-9910)

### DIFF
--- a/pkg/controller/validation/account_validation_controller.go
+++ b/pkg/controller/validation/account_validation_controller.go
@@ -115,7 +115,7 @@ func ParentsTillPredicate(awsId string, client awsclient.Client, p func(s string
 
 // Verify if the account is already in the root OU
 // The predicate indicates if the parent considered the desired root was found.
-func IsAccountInPoolOU(account awsv1alpha1.Account, client awsclient.Client, isPoolOU func(s string) bool) bool {
+func IsAccountInCorrectOU(account awsv1alpha1.Account, client awsclient.Client, isPoolOU func(s string) bool) bool {
 	if account.Spec.AwsAccountID == "" {
 		return false
 	}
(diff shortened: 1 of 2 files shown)


iamkirkbater:
/hold 

I still need to test this locally

pkg/controller/validation/account_validation_controller.go
```diff
@@ -267,15 +267,21 @@ func ValidateAwsAccountId(account awsv1alpha1.Account) error {
...
-func (r *ValidateAccount) ValidateAccountOU(awsClient awsclient.Client, account awsv1alpha1.Account, poolOU string) error {
+func (r *ValidateAccount) ValidateAccountOU(awsClient awsclient.Client, account awsv1alpha1.Account, poolOU string, claimedAccountOU string) error {
 	// Perform all checks on the account we want.
-	inPool := IsAccountInPoolOU(account, awsClient, func(s string) bool {
-		return s == poolOU
+	correctOU := poolOU
+
+	if account.IsClaimed() {
+		correctOU = claimedAccountOU
+	}
```
macgregor, line 276: I dont think this is right, the claimed OU structure looks like this doesnt it: `base ou -> some ou based on the org that owns the cluster i think -> account`

I dont know the exact logic for how claimed accounts get grouped under the base OU, but my point is that base OU isnt the direct parent.

iamkirkbater: Hmm, good catch. I'll have to go brush up on the OU code to double check that then.

iamkirkbater:
Updated to make sure that we find the correct OU.  

One thing I want to double check though is whether we want _ALL_ unclaimed accounts to be in the pool OU, even if they're reused, or if we want only those that have _never_ been claimed to be in the pool OU.

pkg/controller/validation/account_validation_controller.go
```diff
@@ -202,10 +203,8 @@ func ValidateAccountTags(client awsclient.Client, accountId *string, shardName s
 
 					return nil
 				} else {
-					return &AccountValidationError{
-						Type: IncorrectOwnerTag,
-						Err:  fmt.Errorf("Account is not tagged with the correct owner, has %s; want %s", *tag.Value, shardName),
-					}
+					log.Info(fmt.Sprintf("Account is not tagged with the correct owner, has %s; want %s", *tag.Value, shardName))
```
iamkirkbater, line 206: I removed the error here as it was really just polluting the logs and causing the validation controller to continually retry when the flag was turned off, which is just wasting compute cycles when the flag is off.

macgregor (changes requested):
looking good but I would like to see a unit test

pkg/controller/validation/account_validation_controller.go
AlexVulaj: I'm surprised this didn't break any tests, as I thought tests here were validating for a specific error type with a correct message.  Either way I'm fine with it.

PR #725, opened 2022-08-29
Migrate aws-account-operator to OSDKv1

mrbarge:
### What type of PR is this?

Feature

### What this PR does / why we need it?

This PR migrates `aws-account-operator` to be compatible with OperatorSDK v1.21.

This has required involved a number of changes, which are summarized below.

#### Major changes

- API package relocation per OSDKv1 standards (`pkg/apis/v1alpha1` -> `api/v1alpha1`) [0]
- Controller package relocation per OSDKv1 standards (`pkg/controller` -> `controllers`)
- Golang minimum version bumped to 1.17
- `controller-runtime` version bumped from 0.4.0 to 0.21.1
- Removed version pin to kubernetes-1.16.2. `k8s/api` dependencies bumped to `0.24.0`
- The dedicated `go.mod` for the API was also updated to align its dependencies with those of the underlying operator.

#### Minor changes

- Boilerplate update to latest release.
- Controllers and associated reconciler structs were renamed from `<Reconcile>API` to `<API>Reconciler` per OSDKv1 conventions (eg. `ReconcileAccount` -> `AccountReconciler`)
- Many tests broke after the migration. A majority  broke due to the bump in the `golang/mock` dependency. The migration uncovered several underlying issues in the tests, which were addressed:
  - Tests which did not define Name/Namespaces for their test objects; the mocked client now rejects calls relating to objects that are missing these. A correct name/namespace was set in impacted test objects.
  - Tests which included mocked client `EXPECT()` calls that were never actually being called. This was ignored in the older `golang/mock` library but now triggers a test error if they are not called when the test ends. (verification of this was conducted by both analysing the old AAO tests and new AAO tests side-by-side in step debuggers to confirm matching execution flow). These erroneous `EXPECT()` calls were removed.
  - The `logr` version bump from 0.x to 1.x changes its interfaces in a way that broke the existing null test logger, a replacement was included.
  - Several other more unique edge cases of broken tests were uncovered and fixed. See the doc linked below for more details.
- It is no longer possible to `operator-sdk run` an operator locally, the `deploy-local` Make target was changed to an alternate `go run` command which should work.

A more detailed summary of why certain changes were performed is available [here](https://docs.google.com/document/d/1Dn0FbAUmBEFOY8YUSuExWaiUHoyXI3WogA8Ktb6kYYY/edit#).

All commits up to and including 78d762c8c0b9f0580c5794fe67a12a03198f08c9 (the most recent on master branch) are included in this change.

[0] This is a change that will impact the following repositories when they next update their AAO dependency version:
- ocm-aws-account-operator-api
- [certman-operator](https://github.com/openshift/certman-operator) (will likely be addressed in [OSD-11798](https://issues.redhat.com//browse/OSD-11798))
- [osdctl](https://github.com/openshift/osdctl) 
- https://github.com/ramonbutter/velero-plugin-status-restorer 
- https://github.com/NautiluX/managed-velero-plugin

### Which Jira/Github issue(s) this PR fixes?

[OSD-11799](https://issues.redhat.com//browse/OSD-11799) / [OSD-13030](https://issues.redhat.com//browse/OSD-13030)

### Testing notes

Some pre-testing of this PR has occurred:
- Testing that OLM successfully updates AAO from the current version to this version
- Testing that AAO creates a metrics service / route if one does not exist
- Testing that metrics are still accessible from the route and contain AAO-related local metrics
- All tests in the `make test` integration suite ran successfully (except for one, `test-aws-ou-logic`, which failed to clean up some resources at the very end - but I had issues with this with the original AAO too, so it may be something related to my testing accounts)

### Pre-checks (if applicable):
- [X] Tested latest changes against a cluster
- [X] Included documentation changes with PR


mrbarge:
/hold
whilst it is reviewed and tested

controllers/validation/account_validation_controller.go
```diff
@@ -464,3 +442,19 @@ func (r *ValidateAccount) Reconcile(request reconcile.Request) (reconcile.Result
 	}
 	return utils.DoNotRequeue()
 }
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *AccountValidationReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	r.awsClientBuilder = &awsclient.Builder{}
```
iamkirkbater, line 448:
```suggestion
	r.awsClientBuilder = &awsclient.Builder{}
	r.OUNameIDMap = map[string]string{}
```

This is required otherwise it will panic from attempting to assign to a nil map

iamkirkbater (changes requested):
Overall this LGTM.  Need to get that suggestion in and then we can get this merged on Monday.

iamkirkbater:
I'm going to approve this.  I don't think the `approve` label will come off when a change is made (where an LGTM label would), so I think anyone in EMEA on the team will be able to LGTM this on Monday morning.

/approve

controllers/validation/account_validation_controller.go
mrbarge: Oh, great catch - thanks for spotting it! I must have missed it during the rebase. Added back in now.

PR #727, opened 2022-08-29
Adds check if account is claimed before moving OU - Take Two

iamkirkbater:
Reverts openshift/aws-account-operator#726, which removed the functionality that another PR added.

This re-adds the original functionality of trying to allow correct movement of the OU, however instead of using the accountclaim controller's naive implementation we're utilizing an in-memory map of the OUs that we've seen before to sidestep AWS Rate limits.

### DIFF
--- a/pkg/controller/validation/account_validation_controller.go
+++ b/pkg/controller/validation/account_validation_controller.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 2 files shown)


iamkirkbater:
TODO: Fix or remove unit tests to see if this works in stage before spending too much more time trying to fix unit tests.

bergmannf:
/lgtm

Bit strange that there is no way to filter organizational-units by name, but I looked through the API-calls and there seems to be no faster way to do it, than iterating unfortunately.

PR #729, opened 2022-09-20
[OSD-13258] Do not validate accounts being deleted.

bergmannf:
As we already do not validate accounts that don't exist, this will also skip validation for accounts that are in the process of being deleted.

### DIFF
--- a/controllers/validation/account_validation_controller.go
+++ b/controllers/validation/account_validation_controller.go
@@ -4,9 +4,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"sigs.k8s.io/controller-runtime/pkg/controller"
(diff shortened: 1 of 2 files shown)


controllers/validation/account_validation_controller.go
```diff
@@ -360,6 +361,10 @@ func (r *AccountValidationReconciler) Reconcile(ctx context.Context, request ctr
 		log.Error(err, "Could not retrieve account to validate", "account-request", request.NamespacedName)
```
iamkirkbater, line 361: Can we also change this to a `log.Info` here?
```suggestion
		log.Info(err, "Account does not exist", "account-request", request.NamespacedName)
```

PR #733, opened 2022-09-27
Removes accountpool validation check

iamkirkbater:
Removes the extra check in the validation controller making sure that the account is owned by the accountpool.  This gives us the ability to validate migrated accounts (migrated from v3 - v4 shards)


AlexVulaj:
I'm a little wary of dropping this check outright.  Disregarding the migration from v3 to v4, does this check have any value for v4 clusters?  As a follow up, what circumstances would need to occur for an account to _not_ be owned by the account pool?

bergmannf:
I think we can put this in, but I would run in non-enabled mode again to check we don't move something we don't want to.

PR #735, opened 2022-11-18
[OSD-13615] AAO accountpool Service Quota handling 

dkeohane:
# What is being added?
AWS Account Operator supports increasing support quota limits for a specific accountpool on a per-region basis.

Example configmap formatting:
```
zero-size-accountpool: 
  default: true
hs-zero-size-accountpool:
  servicequotas:
    default:
      L-1216C47A: '750'
      L-0EA8095F: '200'
      L-69A177A2: '255'
      L-0263D0A3: '6'
    us-east-1:
      L-1216C47A: '11'"
```
The above information gets loaded into the spec of an Account CR on account creation in the accountpool_controller (therefore this code change will not affect existing account CRs for a given pool, even if this pool has a SQ defined - those changes will be added in a future PR). These newly created Account CRs will attempt to request a service quota increase during their 'PendingValidation' state (this process can take a few hours per Account), and will eventually become Ready on completion. 

Ref [OSD-13615](https://issues.redhat.com//browse/OSD-13615)
Ref [OSD-15001](https://issues.redhat.com/browse/OSD-15001)


controllers/accountclaim/accountclaim_controller.go
```diff
@@ -146,20 +147,34 @@ func (r *AccountClaimReconciler) Reconcile(ctx context.Context, request ctrl.Req
...
+		}
+
+		if err = r.Client.List(context.TODO(), accountList, listOpts...); err != nil {
+			reqLogger.Error(err, "Unable to get accountList")
+			return reconcile.Result{}, err
+		}
+
+		// If the AccountClaim is targetting a specific AccountPool, we want to claim an account from it
+
+		filteredAccountList := &awsv1alpha1.AccountList{}
```
iamkirkbater, line 168: I think we'll need to filter _out_ accounts from the non-default accountpool if it's not defined, too, right?

Otherwise there's a chance that a normal OSD account will attempt to claim one from the other pool.

iamkirkbater (changes requested):
Mostly looks good, definitely on the right track.  One change that needs to be addressed.

api/v1alpha1/account_types.go
```diff
@@ -176,6 +196,26 @@ func (a *Account) HasSupportCaseID() bool {
 	return a.Status.SupportCaseID != ""
 }
 
+//HasOpenQuotaIncreaseRequests returns true if an account has any open quota increase requests
+func (a *Account) HasOpenQuotaIncreaseRequests() bool {
+	for _, v := range a.Status.QuotaIncreaseRequests {
```
iamkirkbater, line 201: Was the linter complaining about using `_, v` in general, or was it complaining about using it when the list is a list of pointers?  I don't remember, but we should double check the linter to make sure it doesn't complain about this, where it was pushing us in the past to use `for i := range myList { mylist[i].something }`

controllers/accountclaim/accountclaim_controller.go
```diff
@@ -427,49 +418,92 @@ func (r *AccountClaimReconciler) getClaimedAccount(accountLink string, namespace
...
 	if unclaimedAccountFound {
 		reqLogger.Info(fmt.Sprintf("Claiming account: %s", unclaimedAccount.ObjectMeta.Name))
 		return &unclaimedAccount, nil
 	}
-
 	// Neither unclaimed nor reused accounts found
 	return nil, fmt.Errorf("can't find a ready account to claim")
 }
 
+func (r *AccountClaimReconciler) getDefaultAccountPoolName() (string, error) {
```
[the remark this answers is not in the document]
iamkirkbater: Survey from Slack says that we should put this in the configmap.

controllers/accountclaim/accountclaim_controller_test.go
```diff
@@ -313,3 +314,347 @@ func (p *possiblyErroringFakeCtrlRuntimeClient) Update(
...
+
+	err := apis.AddToScheme(scheme.Scheme)
+	if err != nil {
+		fmt.Printf("failed adding apis to scheme in account controller tests")
+	}
+	localmetrics.Collector = localmetrics.NewMetricsCollector(nil)
+
+	BeforeEach(func() {
+		ctrl = gomock.NewController(GinkgoT())
+		myYaml := `
```
bergmannf, line 348: Does this work? It seems to be missing any of the 'multiline' markers for yaml like `|`.

controllers/account/account_controller_test.go
```diff
@@ -1415,4 +1420,167 @@ var _ = Describe("Account Controller", func() {
...
+						Namespace: TestAccountNamespace,
+						Name:      TestAccountName,
+					}, &k8sAccount)
+					Expect(k8sAccount.Status.State).To(Equal(AccountReady))
+				})
+			})
+		})
+	})
+
+	AfterEach(func() {
```
dkeohane, line 1583: Probably move this back up to where it was to be more in line with the rest of our tests

controllers/account/cases.go
```diff
@@ -111,6 +111,5 @@ func checkCaseResolution(reqLogger logr.Logger, caseID string, client awsclient.
 		return true, nil
 	}
 
-	// reqLogger.Info(fmt.Sprintf("Case [%s] not yet Resolved, waiting. Current Status: %s", caseID, *caseResult.Cases[0].Status))
```
dkeohane, line 114: Add back in?

controllers/account/ec2.go
```diff
@@ -43,7 +39,7 @@ func (r *AccountReconciler) InitializeSupportedRegions(reqLogger logr.Logger, ac
 	// We should not bomb out just because we can't retrieve the vCPU value
 	// and we'll just continue with a "0"
 	// Errors are logged already in getDesiredVCPUValue
-	vCPUQuota, _ := r.getDesiredVCPUValue(reqLogger)
+	vCPUQuota, _ := r.getDesiredServiceQuotaValue(reqLogger, "vcpu")
```
dkeohane, line 42: We should validate which types we want to use where, either string or the new servicequota types

controllers/account/service_quota.go
```diff
@@ -0,0 +1,382 @@
...
+	"github.com/go-logr/logr"
+	awsv1alpha1 "github.com/openshift/aws-account-operator/api/v1alpha1"
+	"github.com/openshift/aws-account-operator/pkg/awsclient"
+	controllerutils "github.com/openshift/aws-account-operator/pkg/utils"
+	"github.com/openshift/aws-account-operator/test/fixtures"
+)
+
+const (
+/*
+   limits required are in this accepted adr's "how" section: SD-ADR-0075: OSD Fleet Manager AWS Account Management
```
dkeohane, line 21: remove?

controllers/account/service_quota.go
```diff
@@ -0,0 +1,382 @@
...
+		}
+	}
+
+	// Otherwise return false (doesn't need increase) and any errors
+	return false, err
+}
+
+// setRegionVCPUQuota sets the AWS quota limit for vCPUs in the region
+// This just sends the request, and checks that it was submitted, and does not wait
+func setServiceQuota(client awsclient.Client, quotaCode string, serviceCode string, desiredQuota float64) (string, error) {
```
dkeohane, line 175: Add reqLogger?

controllers/account/service_quota_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/openshift/aws-account-operator/pkg/awsclient/mock"
+	"github.com/openshift/aws-account-operator/pkg/testutils"
+	"k8s.io/client-go/kubernetes/scheme"
+)
+
+func TestAccountReconciler_HandleServiceQuotaRequests(t *testing.T) {
+
+	err := apis.AddToScheme(scheme.Scheme)
+	if err != nil {
+		fmt.Printf("failed adding to scheme in account_controller_test.go")
```
dkeohane, line 23: service_quota_test.go

controllers/accountclaim/accountclaim_controller.go
```diff
@@ -427,45 +419,72 @@ func (r *AccountClaimReconciler) getClaimedAccount(accountLink string, namespace
...
+	}
+
+	defaultAccountPoolName, err := config.GetDefaultAccountPoolName(reqLogger, r.Client)
+
+	if err != nil {
+		return nil, err
+	}
+
+	if defaultAccountPoolName == "" {
+		return nil, fmt.Errorf("can't find a ready account to claim") // TODO CHANGE WORDING
```
dkeohane, line 442: TODO

controllers/accountpool/accountpool_controller.go
```diff
@@ -103,8 +114,52 @@ func (r *AccountPoolReconciler) Reconcile(ctx context.Context, request ctrl.Requ
...
+func (r *AccountPoolReconciler) handleServiceQuotas(reqLogger logr.Logger, account *awsv1alpha1.Account) error {
+	reqLogger.Info("Loading Service Quotas")
+
+	cm, err := utils.GetOperatorConfigMap(r.Client)
+	if err != nil {
+		reqLogger.Error(err, "failed retrieving configmap")
+		return err
+	}
+
+	accountpoolString := cm.Data["accountpool"]
```
dkeohane, line 126: err handling

api/v1alpha1/account_types.go
```diff
@@ -35,23 +35,61 @@ type AccountSpec struct {
...
+type ServiceRequestStatus string
+
+const (
+	ServiceRequestTodo       ServiceRequestStatus = "TODO"
+	ServiceRequestInProgress ServiceRequestStatus = "IN_PROGRESS"
+	ServiceRequestCompleted  ServiceRequestStatus = "COMPLETED"
+	ServiceRequestDenied     ServiceRequestStatus = "DENIED"
+)
+
+type SupportedServiceQuotas string
```
AlexVulaj, line 64: It seems odd to create custom aliases for primitive types, but I think it makes sense in this case for clarity.

controllers/account/ec2.go
AlexVulaj, line 42: Let's handle this error here in some way, even if all we do is log it.

controllers/account/service_quota.go
```diff
@@ -0,0 +1,352 @@
+package account
+
+import (
+	"fmt"
+	"strconv"
+	"time"
+
+	retry "github.com/avast/retry-go"
```
AlexVulaj, line 8: My IDE tells me this alias isn't needed because the package itself is named "retry." Non functional but maybe worth the clean up.

controllers/account/service_quota.go
```diff
@@ -0,0 +1,352 @@
...
+			reqLogger.Info(fmt.Sprintf("Requiring a servicequota increase: current [%.1f] wanted [%.1f]\n", *result.Quota.Value, desiredQuota))
+			return true, err
+		}
+	}
+
+	// Otherwise return false (doesn't need increase) and any errors
+	return false, err
+}
+
+func setServiceQuota(reqLogger logr.Logger, client awsclient.Client, quotaCode string, serviceCode string, desiredQuota float64) (bool, error) {
```
AlexVulaj, line 183: I think we can drop the logger here - I don't see anywhere it's being used.

controllers/account/service_quota.go
```diff
@@ -0,0 +1,352 @@
...
+		return false, err
+	}
+
+	if err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+func checkQuotaRequestStatus(reqLogger logr.Logger, awsClient awsclient.Client, quotaCode string, serviceCode string, expectedQuota float64) (awsv1alpha1.ServiceRequestStatus, error) {
```
AlexVulaj, line 262: Same comment about this logger

controllers/account/service_quota.go
```diff
@@ -0,0 +1,352 @@
...
+				)
+				return err
+			},
+
+			// Retry if we receive some specific errors: access denied, rate limit or server-side error
+			retry.RetryIf(func(err error) bool {
+				if aerr, ok := err.(awserr.Error); ok {
+					switch aerr.Code() {
+					// AccessDenied may indicate the BYOCAdminAccess role has not yet propagated
+					case "AccessDeniedException":
```
AlexVulaj, line 293: Last place I see this - just a duplicate comment on the above.

controllers/accountpool/accountpool_controller.go
```diff
@@ -106,6 +114,61 @@ func (r *AccountPoolReconciler) Reconcile(ctx context.Context, request ctrl.Requ
...
+		return err
+	}
+
+	accountpoolString, found := cm.Data["accountpool"]
+	if !found {
+		reqLogger.Error(fixtures.NotFound, "failed getting accountpool data from configmap")
+		return fixtures.NotFound
+	}
+
+	type Servicequotas map[string]string
```
AlexVulaj, line 132: Could this reuse the `ServiceQuotas` type from account_types?

AlexVulaj:
This all looks good to me functionally. I left a few comments that are more stylistic than anything. I'd feel more comfortable with more test cases in service_quota_test.

controllers/accountpool/accountpool_controller.go
bergmannf: This one is different because the configmap holds a nested YAML for the servicequota definition we want to load. 
Because of the nested YAML we can't load it in one step in go, but instead have to load a string first and then convert it to the actual type later in this function:

This is a configmap example we use to test:
```yaml
accountpool: "\nzero-size-accountpool: \n  default: true\nhs-zero-size-accountpool:\n
    \ servicequotas:\n    default:\n      L-1216C47A: '10'\n      L-0263D0A3: '6'\n
    \   us-east-1:\n      L-1216C47A: '11'"
```

And the value of the accountpool must first be loaded as string as it's still yaml, after the first Unmarshal ...

api/v1alpha1/account_types.go
bergmannf: It helps to limit the values it can take at least - kind of like a poor man's enum.

PR #736, opened 2022-11-29
Add a consumer ServiceAccount, ClusterRole, and ClusterRoleBinding

tiwillia:
# What is being added?
The existing `aws-account-operator-client` service account contains most of the permissions required for the uhc specific use case. For a more generic use case where the AWS account credentials are simply consumed from the created secret, access to those secrets and the credentials within is missing.

This change adds a new `aws-account-operator-consumer` ServiceAccount, ClusterRole, and ClusterRoleBinding that grants just the necessary permissions for this basic use-case

### DIFF
--- a/deploy/consumer_cluster_role.yaml
+++ b/deploy/consumer_cluster_role.yaml
@@ -1,0 +1,23 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: aws-account-operator-consumer
+rules:
+- apiGroups:
+  - "aws.managed.openshift.io"
+  resources:
+  - accountclaims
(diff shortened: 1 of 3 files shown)


patjlm:
Note: this would grant the serviceaccount access to all secrets on the cluster.

bergmannf:
As @patjlm already noted - the permissions should at most be scoped to the aws-account-operator namespace only.

But I also don't really see what this serviceaccount is used for - is there some ticket as a reference as to why the new account is needed?

iamkirkbater:
As discussed in Slack this has been done within app-interface and we will not be creating the ServiceAccount within AAO.  I'm going to close this PR.

/close

PR #740, opened 2023-01-19
[OSD-13614] Add multiple accountpools to AAO

dkeohane:
- Add handling to AAO for multiple account pools.
- Account claims can now specify which account pool they would like to claim from, if none specified, will claim from default account pool.

 
Ticket ref: [OSD-13614](https://issues.redhat.com//browse/OSD-13614)

### DIFF
--- a/api/v1alpha1/account_types.go
+++ b/api/v1alpha1/account_types.go
@@ -38,6 +38,7 @@ type AccountSpec struct {
 	ClaimLinkNamespace string      `json:"claimLinkNamespace,omitempty"`
 	LegalEntity        LegalEntity `json:"legalEntity,omitempty"`
 	ManualSTSMode      bool        `json:"manualSTSMode,omitempty"`
+	AccountPool        string      `json:"accountPool,omitempty"`
 }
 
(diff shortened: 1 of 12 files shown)


dkeohane:
/hold
Need to confirm everything from other PR has been copied over correctly

iamkirkbater (commented):
LGTM

Do we need to set up the configmaps in app-interface before this is deployed?  Without a "named default" will this work?

dkeohane:
> LGTM
> 
> Do we need to set up the configmaps in app-interface before this is deployed? Without a "named default" will this work?

It'll require an configmap update, yes - although I can update the code to not error if it can't find a default pool if you'd prefer

iamkirkbater:
There's no problem with requiring a configmap update - there's no need to hold up this PR, we could potentially check to see if there's only one account pool and then call that the default if there's no configmap entry explicitly saying which one to use in another PR, or some other functionality to make it so that we don't block on this if the configmap isn't present.

With that said, do we have the configmap change queued?  I worry if we don't have that in place before we merge this then we will end up locking out stage/int envs.

PR #744, opened 2023-02-08
Reverts non-crd changes involving multiple accountpools

iamkirkbater:
This reverts the code changes for adding multiple accountpools.  CRD changes were left in to avoid any potential backwards compatibility pitfalls.


iamkirkbater:
/hold

we might not need this....

iamkirkbater:
/hold cancel

I think we still need this for the near future.

PR #753, opened 2023-04-10
OSD-14723 enabling ap-southeast-4 for cloud provider operators

MitaliBhalla:
# What is being added?
Feature : Adding AMI for enabling ap-southeast-4 for cloud provider operators

## Steps To Manually Test
1. Start the operator
1. Run the thing
1. Clean up the thing

Ref OSD-0000


iamkirkbater:
We found the issue to be that Boilerplate now enforces that the olm-registry is built to the latest version of ubi-micro.

That's what's getting caught in the validate job.

When running `make validate` locally, if `skopeo` isn't installed, it just returns an empty string, which is what we're seeing with the empty `FROM` in the dockerfile.

We're running the boilerplate update in a separate PR, then you should be able to just rebase on the latest from `master` here and remove the `boilerplate-update` comit from this PR, and this should validate properly.

PR #751, opened 2023-04-06
When an AWSFederatedRole CR is updated, automatically update all associated AWSFederatedAccountAccess CRs

AlexVulaj:
# What is being added?
When an `awsfederatedaccountrole` CR is updated, this change will pick up that change and sync it to all `awsfederatedaccountaccess` IAM policies for that role.

## Checklist before requesting review

- [x] I have tested this locally
- [x] I have included unit tests
- [ ] I have updated any corresponding documentation

## Steps To Manually Test
1. Start the operator
1. Create a new `awsfederatedrole` CR and `awsfederatedaccountaccess` CR for that role.
1. Check the AWS console for the account you're working with - validate that IAM policy created matches the federated role CR.
1. `oc edit` the `awsfederatedrole` you created to change the permissions in `.spec.awsCustomPolicy.awsStatements.action`.
1. Refresh your AWS console. Observe that the change made to the CR in the previous step is reflected in the IAM policy.

Ref [OSD-15564](https://issues.redhat.com//browse/OSD-15564)

PR #757, opened 2023-04-12
Make sure to add federated role label to existing accounts if not present

AlexVulaj:
Small miss from https://github.com/openshift/aws-account-operator/pull/751.

Catches an edge case where the `awsFederatedRole` label would not be applied to existing `awsfederatedaccountaccess` CRs.

### DIFF
--- a/controllers/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
+++ b/controllers/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
@@ -148,11 +148,7 @@ func (r *AWSFederatedAccountAccessReconciler) Reconcile(_ context.Context, reque
 			newLabel := map[string]string{awsv1alpha1.FederatedRoleNameLabel: requestedRole.Name}
 
 			// Join the new UID label with any current labels
-			if currentFAA.Labels != nil {
-				currentFAA.Labels = controllerutils.JoinLabelMaps(currentFAA.Labels, newLabel)
-			} else {
-				currentFAA.Labels = newLabel
-			}
+			currentFAA.Labels = controllerutils.JoinLabelMaps(currentFAA.Labels, newLabel)
 
 			err = r.Client.Update(context.TODO(), currentFAA)
 			if err != nil {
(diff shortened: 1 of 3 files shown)


controllers/awsfederatedaccountaccess/awsfederatedaccountaccess_controller.go
```diff
@@ -141,6 +141,26 @@ func (r *AWSFederatedAccountAccessReconciler) Reconcile(_ context.Context, reque
...
+			reqLogger.Info(fmt.Sprintf("Adding %s label with value %s to AccountAccess %s", awsv1alpha1.FederatedRoleNameLabel, requestedRole.Name, currentFAA.Name))
+
+			newLabel := map[string]string{awsv1alpha1.FederatedRoleNameLabel: requestedRole.Name}
+
+			// Join the new UID label with any current labels
+			if currentFAA.Labels != nil {
+				currentFAA.Labels = controllerutils.JoinLabelMaps(currentFAA.Labels, newLabel)
+			} else {
+				currentFAA.Labels = newLabel
+			}
```
fahlmant, line 155:
```suggestion
			// Join the new UID label with any current labels
			currentFAA.Labels = newLabel
			if currentFAA.Labels != nil {
				currentFAA.Labels = controllerutils.JoinLabelMaps(currentFAA.Labels, newLabel)
			} 
```

Minor nit, this is more "Go-like" to try and avoid an else

AlexVulaj: This would override the existing map on the CR, however I think we could update `JoinLabelMaps` to explicitly handle null maps. Let me take a look...

AlexVulaj:
/hold

I need to investigate the impact of my changes to JoinLabelMaps more.

iamkirkbater:
Was able to test this by doing the following:

* Create the FederatedRoles.
* Create an account.
* Create a FederatedAccountAccess
* Delete the `awsFederatedRoleName` label out of the FederatedAccountAccess
* Checkout this PR
* Start the operator
* Validate that the label is applied to the FederatedAccountAccess CR.

/lgtm
/hold
Unhold this in the morning.

PR #765, opened 2023-06-20
Add ServiceQuotasReadOnlyAccess managed policy to Read only users

BumbleFeng:
related to https://issues.redhat.com/browse/OSD-17109

OCM users use the AWS Infrastructure Access feature to gain access to the cluster's AWS account. They can either choose read-only or network-mgmt roles.

To enable OSD users to get proper usage/quota metrics for AWS accounts, it is expected that users can get service quota from there. When trying to do that currently, a user gets a permission error (not authorized to perform: servicequotas:GetServiceQuota).

This PR adds the ServiceQuotasReadOnlyAccess managed policy to the Read only access policy.


fahlmant:
/hold

This will expose Service Quotas to external customers using this feature as well, which needs review.

PR #767, opened 2023-06-29
Prefer scheduling registry pod onto infra nodes

anispate:
This commit specifies .spec.grpcPodConfig.nodeSelector and .spec.grpdPodConfig.tolerations to schedule this operator's registry pod to infra nodes when possible.

[OSD-6629](https://issues.redhat.com//browse/OSD-6629)


iamkirkbater:
/hold

I'm super excited for this feature now that it's available, however I don't think this applies to AWS Account Operator.

PR #772, opened 2023-08-01
[OSD-17714] Decouple Functions from account controller

reedcort:
# What is being added?
Decouple the "[AssumeRoleFunc](https://issues.redhat.com/secure/RapidBoard.jspa?rapidView=5364&projectKey=OSD&view=detail&selectedIssue=[OSD-17714](https://issues.redhat.com//browse/OSD-17714)#L779])" and "[assumeRole](https://issues.redhat.com/secure/RapidBoard.jspa?rapidView=5364&projectKey=OSD&view=detail&selectedIssue=[OSD-17714](https://issues.redhat.com//browse/OSD-17714)#L789])" functions from the account_controller.go file and move them to the pkg directory. This will enable other controllers within the project to utilize these functions

## Steps To Manually Test
1. Start the operator
1. Run the thing
1. Clean up the thing

Ref OSD-0000


pkg/awsclient/sts_assume_role.go
reedcort, line 1: Should I change the name of this file and should I move it outside the "pkg/awsclient"

api/v1alpha1/account_types.go
```diff
@@ -385,6 +389,15 @@ func (a *Account) IsOwnedByAccountPool() bool {
 	return false
 }
 
+func (a *Account) GetAssumeRole() string {
```
reedcort, line 392: Should I leave this function in account_types or move it to the new sts_assume_role pkg?

controllers/account/account_controller.go
```diff
@@ -36,7 +34,8 @@ import (
 )
 
 var log = logf.Log.WithName("controller_account")
-var AssumeRole = AssumeRoleFunc
+var AssumeRole = awsclient.AssumeRoleFunc
+var assumeRole = awsclient.AssumeRole
```
iamkirkbater, line 38: This is confusing - what is the difference between these two things? Can these be named better?

controllers/account/account_controller.go
```diff
@@ -776,106 +829,106 @@ func TagAccount(awsSetupClient awsclient.Client, awsAccountID string, shardName
...
-		)
-		if stateErr != nil {
-			reqLogger.Error(err, "failed setting account state", "desiredState", AccountFailed)
-		}
-		return nil, nil, err
-	}
-
-	return awsAssumedRoleClient, creds, nil
-}
+//func AssumeRoleFunc(r *AccountReconciler,
```
iamkirkbater, line 832: Can we remove these lines instead of comment them out? Git history will keep these for us.

controllers/account/account_controller.go
```diff
@@ -1263,15 +1316,15 @@ func matchSubstring(roleID, role string) (bool, error) {
...
-func getAssumeRole(c *awsv1alpha1.Account) string {
-	// If the account is a CCS account, return the ManagedOpenShiftSupport role
-	if c.IsBYOC() {
-		return fmt.Sprintf("%s-%s", awsv1alpha1.ManagedOpenShiftSupportRole, c.Labels[awsv1alpha1.IAMUserIDLabel])
-	}
-
-	// Else return the default role
-	return awsv1alpha1.AccountOperatorIAMRole
-}
+//func getAssumeRole(c *awsv1alpha1.Account) string {
```
iamkirkbater, line 1319: Same for the removal

iamkirkbater (changes requested):
Looks mostly good - Just a small nit about the naming of the assumerole vs AssumeRole variables - they're going to cause confusion in the future if they're not either named better or explicitly commented as to why they're named the way they are, but I think we can find better names for them that are more descriptive.

Also, let's just remove the commented out lines.

Great work so far!

controllers/account/account_controller.go
reedcort: In the original code one function was "AssumeRoleFunc" and the other just "assumeRole".  "AssumeRoleFunc"  is basically a wrapper function for `assumeRole`. `assumeRole` handles the logic for assuming an AWS role and obtaining temporary credentials. I can either do 1 or 2

1. update line 37 and add "Func" to the end of "AssumeRole" to keep it somewhat consistent with the original code and add comments like you suggested to describing what each ones does.

2. change "AssumeRoleFunc" to "AssumeRoleAndCreateClient" and "assumeRole" to "handleRoleAssumption" or any other names that you think better encapsulate there purpose.

ref: https://github.com/reedcort/aws-account-operator/blob/OSD-17714/pkg/awsclient/sts_assume_role.go#L78-L153

iamkirkbater: Let's go with Option two here and rename them :)

pkg/awsclient/sts_assume_role.go
```diff
@@ -0,0 +1,153 @@
...
+		// Log AWS error
+		if aerr, ok := err.(awserr.Error); ok {
+			reqLogger.Error(aerr,
+				fmt.Sprintf(`New AWS Error while getting STS credentials,
+					AWS Error Code: %s,
+					AWS Error Message: %s`,
+					aerr.Code(),
+					aerr.Message()))
+		}
+		return &sts.AssumeRoleOutput{}, err
```
mrWinston, line 72: This line can actually be removed. 
When `err != nil`, the var `assumeRoleOutput` will be nil, and the other way around.
So you can just write 
```
return assumeRoleOutput, err
```
at the end of the function and have the same behaviour.

pkg/awsclient/sts_assume_role.go
```diff
@@ -0,0 +1,153 @@
...
+	// changed if we want it to be longer than 3600 seconds
+	var roleSessionDuration int64 = 3600
+	reqLogger.Info(fmt.Sprintf("Creating STS credentials for AWS ARN: %s", roleArn))
+	// Build input for AssumeRole
+	assumeRoleInput := sts.AssumeRoleInput{
+		DurationSeconds: &roleSessionDuration,
+		RoleArn:         &roleArn,
+		RoleSessionName: &roleSessionName,
+	}
+	if externalID != "" {
```
mrWinston, line 46: Thinking out loud here:
I'm wondering if there's even a difference in the aws api between 
```
sts.AssumeRoleInput{
  ExternalId: nil
}
```
and 
```
sts.AssumeRoleInput{
  ExternalId: aws.String("")
}
```
. Cause for string itself, there's no difference between a nil string and "". If there were no difference in the api between nil and a pointer to `""`, then we could just get rid of this check and always just pass the parameter to the AssumeRoleInput.

controllers/account/iam_test.go
```diff
@@ -91,7 +91,7 @@ func TestGetSTSCredentials(t *testing.T) {
 		nil, // no error
 	)
 
-	creds, err := getSTSCredentials(
+	creds, err := awsclient.GetSTSCredentials(
```
mrWinston, line 94: I think it makes sense to move the tests to a new file `sts_assume_role_test.go`, so that they are next to what they are actually testing.

api/v1alpha1/account_types.go
mrWinston: I would leave it here, as it's not interacting with the aws api in any way. While at it, i would rename the function to `GetRoleARNToAssume` to make it clear that it just constructs the arn and nothing more.

pkg/awsclient/sts_assume_role.go
```diff
@@ -0,0 +1,153 @@
...
+		if err == nil {
+			break
+		}
+		if i == 99 {
+			reqLogger.Info(fmt.Sprintf("Timed out while assuming role %s", roleArn))
+		}
+	}
+	if err != nil {
+		// Log AWS error
+		if aerr, ok := err.(awserr.Error); ok {
```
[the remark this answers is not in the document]
reedcort: yeah that how it was in the original code, but I will add the else statement since it doesn't hurt to have it.

controllers/account/iam_test.go
reedcort: I have no issue with that, but we would have to make the `[setupDefaultMocks](https://github.com/openshift/aws-account-operator/blob/master/controllers/account/account_controller_test.go#L54)` in the account_controller_test.go public. @iamkirkbater should I go ahead and make the change

pkg/awsclient/sts_assume_role.go
reedcort: when I remove that the return statement the [TestGetSTSCredentials ](https://github.com/openshift/aws-account-operator/blob/5be3cb3945eb31bf35836985bae93af675664177/controllers/account/iam_test.go#L130) fails

controllers/account/iam_test.go
mrWinston: I wouldn't neccessarily make the function public but just instantiate the mocks directly. From what `setupDefaultMocks` returns, the only thing we're using in these tests is actually `mocks.mockCtrl`, which we can also just initialize with a single line. So, instead of making `setupDefaultMocks` public, id just change the first few lines of the Test Like so:
```
	mockCtrl := gomock.NewController(t)	
        nullLogger := testutils.NewTestLogger().Logger()
        mockAWSClient := mock.NewMockClient(mockCtrl)
	defer mockCtrl.Finish()

```

pkg/awsclient/sts_assume_role.go
mrWinston: I see, that's because of the way the test is set up. Because the test is both returning an error and sts credentials in the [AssumeRole call](https://github.com/openshift/aws-account-operator/blob/5be3cb3945eb31bf35836985bae93af675664177/controllers/account/iam_test.go#L110C7-L110C7) and then expects the returned credentials to actually by empty. Imho, we should remove that check from the test, as i don't really see the value in that. But I also don't think it's that important, so i leave this up to you :)

controllers/account/iam_test.go
reedcort: I'm not able to call the mock pkg, from my test. I'm running into a "circular import dependency" error since the mock (pkg/awsclient/mock) is within the awsclient pkg where the sts_assume_role_test is located

controllers/account/account_controller.go
```diff
@@ -36,7 +35,8 @@ import (
 )
 
 var log = logf.Log.WithName("controller_account")
-var AssumeRole = AssumeRoleFunc
+var AssumeRoleAndCreateClient = stsclient.AssumeRoleAndCreateClient
+var handleRoleAssumption = stsclient.HandleRoleAssumption
```
bergmannf, line 39: I understand making `AssumeRuleAndCreateClient` a public var so it can be overwritten in tests, but why assign to a private variable for `handleRoleAssumption` - it doesn't seem to require a global variable as it only gets used inside this file?

controllers/account/account_controller.go
```diff
@@ -143,10 +143,28 @@ func (r *AccountReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 
 		var awsClient awsclient.Client
 		if currentAcctInstance.IsBYOC() {
-			roleToAssume := getAssumeRole(currentAcctInstance)
-			awsClient, _, err = r.assumeRole(reqLogger, currentAcctInstance, awsSetupClient, "", roleToAssume, "")
+			roleToAssume := currentAcctInstance.GetAssumeRole()
+			awsClient, _, err = handleRoleAssumption(reqLogger, r.awsClientBuilder, currentAcctInstance, r.Client, awsSetupClient, "", roleToAssume, "")
 			if err != nil {
 				reqLogger.Error(err, "failed building BYOC client from assume_role")
+				// Get custom failure reason to update account status
```
bergmannf, line 150: This error handling block is duplicated 3 times in this function alone - might be worth putting it into one place, so it is easier to update it in the future.

controllers/account/account_controller.go
reedcort: For OSD-14200 it will use the `stsclient.HandleRoleAssumption` function in the account_validation_controller

PR #781, opened 2023-09-11
[OSD-16603] Add 5m timeout to all aws api calls

mrWinston:
# What is being added?
This PR makes sure that calls to the aws api use an `http.Client` with a timeout. This is added as a preventative measure and to hopefully fix the intermittent account creation errors that have been happening for 2 years.

- [ ] I have included unit tests (Current unit tests don't use that code path)

## Steps To Manually Test
- todo for me: figure out if there's a way to test this in the integration tests.
- 
Ref OSD-16603

### DIFF
--- a/pkg/awsclient/client.go
+++ b/pkg/awsclient/client.go
@@ -16,6 +16,7 @@ package awsclient
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws/endpoints"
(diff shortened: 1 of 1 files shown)


pkg/awsclient/client_test.go
```diff
@@ -0,0 +1,52 @@
+package awsclient
```
bergmannf, line 1: Indentation in this file looks very strange - probably should throw `gofmt` at it to get it consistent.

mrWinston: Consider it done :)

PR #788, opened 2023-10-18
[OSD-14200-remove-accountreconciler-dependency] 

reedcort:
# What is being added?
Removing  function accountreconciler dependency

## Steps To Manually Test
1. Start the operator
1. Run the thing
1. Clean up the thing

Ref [OSD-14200](https://issues.redhat.com//browse/OSD-14200)


controllers/account/account_controller.go
```diff
@@ -601,7 +601,14 @@ func (r *AccountReconciler) HandleNonCCSPendingVerification(reqLogger logr.Logge
 				}
 			}
 			reqLogger.Info("Handling quotarequets", "current-in-flight-count", currentInFlightCount)
-			err := r.updateServiceQuotaRequests(reqLogger, awsSetupClient, currentAcctInstance, inFlightQuotaRequests, currentInFlightCount)
+			err := UpdateServiceQuotaRequests(reqLogger, r.awsClientBuilder, awsSetupClient, currentAcctInstance, r.Client, inFlightQuotaRequests, currentInFlightCount)
```
bergmannf, line 604: This obviously is not because of this MR, but I noticed that the order of arguments could be streamlined for `UpdateServiceQuotaRequests` and `SetCurrentAccountServiceQuotas` - e.g. the `awsSetupClient` is a completely different positions in the argument list.

Also - but not sure if this would make it better or worse: given that we are passing a `awsClientBuilder` in - I wonder if the functions could derive their own `awsSetupClient` instead of passing this as well.

I will leave deciding if this should be changed or not to the other reviewers though, as I feel both issues aren't big enough to hold this up.

PR #789, opened 2023-10-19
[OSD-19291] update service quota documentation

reedcort:
# What is being added?
Updated documentation

Ref OSD-19291


docs/8.0-ServiceQuotas.md
```diff
@@ -102,6 +102,11 @@ status:
         status: TODO
         value: 750
 ```
+## How to add support for additional service quotas?
+#### Step 1: Define Service and Quota Codes
+In `account_types.go` you add the quota code if it's not already supported. Below is a list of supported service quotas with their respective codes: [https://github.com/reedcort/aws-account-operator/blob/99f0207d0e3c3704b9d30334079e402d268405f9/api/v1alpha1/account_types.go#L67-L74]. Additionally, in the same file, you must also define the service code for the supported service quota services: [https://github.com/reedcort/aws-account-operator/blob/99f0207d0e3c3704b9d30334079e402d268405f9/api/v1alpha1/account_types.go#L78-L82].
```
AlexVulaj, line 107: Noting the braces around these links - were you intending to give them a title as well?

AlexVulaj: Resolved per slack conversation.

PR #791, opened 2023-10-30
[OSD-19093] Don't run account claim cleanup when there are additional finalizers

mrWinston:
# What is being added?
Add a check to the account cr deletion step to ensure that we only run the account cleanup when there are no additional finalizers present on the accountclaim.

- [X] I have tested this locally
- [X] I have included unit tests
- [X] I have updated any corresponding documentation

Ref [OSD-19093](https://issues.redhat.com//browse/OSD-19093)

### DIFF
--- a/controllers/accountclaim/accountclaim_controller.go
+++ b/controllers/accountclaim/accountclaim_controller.go
@@ -246,6 +246,13 @@ func (r *AccountClaimReconciler) handleAccountClaimDeletion(reqLogger logr.Logge
 		return nil
 	}
 
+  // Workaround for FleetManagers special account handling, see
+  // https://issues.redhat.com/browse/OSD-19093
+  if len(accountClaim.GetFinalizers()) > 1 {
+    reqLogger.Info("Found additional finalizers on AccountClaim. Not attempting cleanup.")
+    return nil
+  }
+
 	// Only do AWS cleanup and account reset if accountLink is not empty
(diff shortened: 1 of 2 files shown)


controllers/accountclaim/accountclaim_controller.go
```diff
@@ -246,6 +246,13 @@ func (r *AccountClaimReconciler) handleAccountClaimDeletion(reqLogger logr.Logge
 		return nil
 	}
 
+  // Workaround for FleetManagers special account handling, see
+  // https://issues.redhat.com/browse/OSD-19093
+  if len(accountClaim.GetFinalizers()) > 1 {
```
iamkirkbater, line 251: Should we also check here that when there _is_ only one finalizer that it is also the AAO finalizer? (it wouldn't have to be in this if block, could be a separate block)

iamkirkbater (changes requested):

controllers/accountclaim/accountclaim_controller.go
mrWinston: This is actually already checked as the first statement in this function. So when we reach this part, we are certain that the accountclaim has the AAO finalizer.

iamkirkbater: Ahh, perfect then. That's what I get for not looking at the whole function. Thanks for keeping me honest :D

iamkirkbater (approved):
/lgtm

PR #796, opened 2023-11-16
OSD-14200 Add Max Concurrent Reconcile limit

reedcort:
# What is being added?
Adds Max Concurrent Reconcile limit field in the configmap for the accountpool validation controller

Ref [OSD-14200](https://issues.redhat.com//browse/OSD-14200)

### DIFF
--- a/controllers/validation/accountpool_validation_controller.go
+++ b/controllers/validation/accountpool_validation_controller.go
@@ -26,6 +26,10 @@ var (
 	logs              = logf.Log.WithName("controller_accountpoolvalidation")
 )
 
+const (
+	validationControllerName = "accountpoolvalidation"
+)
+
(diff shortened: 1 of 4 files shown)


controllers/validation/accountpool_validation_controller.go
AlexVulaj, line 1: Can we revert the changes in this file? I think it's better to have this string defined in one consistent place instead of repeating it in multiple places.

rafael-azevedo: +1

reedcort: Added a the var back but changed the name, because the name 'controllerName' is already being used by the 'accountValidation' controller

PR #819, opened 2024-03-25
[OSD-21095] Implements a new feature to allow supported opt-in region enablement during account creation process

reedcort:
# What is being added?
Implements a new feature to allow supported opt-in region enablement during account creation process

## Steps To Manually Test

1. Start the operator
2. Run "make predeploy"
3. Set "feature.opt_in_regions" to true in hack/templates/aws.managed.openshift.io_v1alpha1_configmap.tmpl
4. Add  "opt-in-regions" entry to hack/templates/aws.managed.openshift.io_v1alpha1_configmap.tmpl
5. Run "make create-account"
6. Set account.Spec.accountPool to "hs-zero-size-accountpool"
7. Verify account CR state changes to READY
8. Verify in the AWS console that the opt-in regions specified in the configMap are enabled

Ref [OSD-21095](https://issues.redhat.com//browse/OSD-21095)


api/v1alpha1/account_types.go
```diff
@@ -82,6 +82,35 @@ const (
...
+	MelbourneRegion SupportedOptInRegions = "ap-southeast-4"
+	HyderabadRegion SupportedOptInRegions = "ap-south-2"
+	MilanRegion     SupportedOptInRegions = "eu-south-1"
+	ZurichRegion    SupportedOptInRegions = "eu-central-2"
+	HongKongRegion  SupportedOptInRegions = "ap-east-1"
+	UAERegion       SupportedOptInRegions = "me-central-1"
+	SpainRegion     SupportedOptInRegions = "eu-south-2"
+	BahrainRegion   SupportedOptInRegions = "me-south-1"
+	JakartaRegion   SupportedOptInRegions = "ap-southeast-3"
+)
```
iamkirkbater, line 112: I'd advise against having this in the code, especially in the API for a few reasons:

1. Every change to add a new AWS region will now require a code change and promotion
2. Being in the API package, which other services (hive, ocm, etc) consume - we could run into issues with downstream consumers not updating, and if we ever make changes for any reason we'll have to run through deprecation procedures, etc.

Mostly - it's a large maintenance cost to add these to the code, and we'll _also_ need subsequent SOP updates to track these changes for our future selves.

controllers/account/account_controller.go
```diff
@@ -106,12 +114,47 @@ func (r *AccountReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+	}
+
+	if err = r.Client.List(context.TODO(), accountList, listOpts...); err != nil {
+		if k8serr.IsNotFound(err) {
+			// Request object not found, could have been deleted after reconcile request.
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
```
iamkirkbater, line 130: Probably want to add log statements before these two return statements

controllers/account/account_controller.go
```diff
@@ -106,12 +114,47 @@ func (r *AccountReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+		}
+	}
+
 	configMap, err := utils.GetOperatorConfigMap(r.Client)
 	if err != nil {
 		log.Error(err, "Failed retrieving configmap")
 		return reconcile.Result{}, err
 	}
 
+	var isEnabled = false
```
iamkirkbater, line 149: Can we rename this var to something more descriptive?

Maybe this whole function could be abstracted?

controllers/account/account_controller.go
```diff
@@ -106,12 +114,47 @@ func (r *AccountReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+
+	// since it's not possible to filter on custom field values when listing using the golang client
+	// manual filtering of accounts opting-in is required to ensure the account limit is not reached
+	numberOfAccountsOptingIn := 0
+	for _, account := range accountList.Items {
+		if account.Status.State == "OptingInRegions" {
+			numberOfAccountsOptingIn += 1
+		}
+	}
+
```
iamkirkbater, line 142: I recommend abstracting this block of code into its own function.

Additionally, I think maybe this code may be better executed only when we need to opt into the region, otherwise we'll be calling this function every time an AccountCR reconciles - which on staging/int hives will be thousands of iterations. So maybe move the invocation of this functionality towards the block at line 413?

controllers/account/account_controller.go
```diff
@@ -367,6 +410,12 @@ func (r *AccountReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 
 		}
 
+		optInRegions, ok := configMap.Data["opt-in-regions"]
+		// handles region enablement for supported Opt-In regions
+		if (ok && optInRegions != "" && isEnabled && !currentAcctInstance.Spec.BYOC) && (currentAcctInstance.Status.State == AccountCreating || currentAcctInstance.Status.State == AccountOptingInRegions) {
+			return r.handleOptInRegionEnablement(reqLogger, currentAcctInstance, awsSetupClient, optInRegions, numberOfAccountsOptingIn)
+		}
+
```
iamkirkbater, line 418: I think we need to rethink how we get to this point - I think at the end of the AccountCreating step we should decide what the next state is and put the CR into that state directly, and then return the reconciler - then when it gets back to this point in the code it skips the Account Creation stuff and gets here and goes "oh, I'm AccountOptingInRegions - I'll do that handler" and returns from that, then when we're done handling the opt-in region enablement we tell that to put it to the next state, instead of relying on it to just fall through as we're doing here for the InitializeRegions.

If that doesn't make sense, let me know I'll be happy to explain.

iamkirkbater (changes requested):

controllers/account/region_enablement.go
```diff
@@ -0,0 +1,349 @@
...
+	// This also defaults to an exponential backoff, so we only need to try ~5 times, default is 10
+	retry.DefaultDelay = 3 * time.Second
+	retry.DefaultAttempts = uint(5)
+	err := retry.Do(
+		func() (err error) {
+			result, err = client.EnableRegion(&account.EnableRegionInput{
+				RegionName: aws.String(regionCode),
+			})
+			if err != nil {
+				if aerr, ok := err.(awserr.Error); ok {
```
AlexVulaj, line 152: Can we replace this with an `errors.As` ? I know we use this structure elsewhere in the codebase but it can lead to mistakes with wrapped errors. `errors.As` traverses all of the wrapped errors while checking for a specific error type.

There are also a few other places that do this in this file as well. I'll leave it to your discretion.

AlexVulaj:
/approve

One comment that's more about semantics than function - it's a change I'd like to see but wouldn't hold up the PR for it.

PR #820, opened 2024-04-04
First Draft AAO changes to service quota optin region issue

luis-falcon:
# What is being added?
This PR will resolve https://issues.redhat.com/browse/OSD-21098

This issue will add the logic to handle the errors on regions that do not support Quota increases and Update the account yaml to put an status of "Manual_action" which will let us know to go into the AWS account and solve the issue manually.

TESTS:
 Deploy a cluster to test. 
 Start the operator.
 Change the account configmap to request a quota increase. (make sure you have enabled an opt-in region that does not support this kind of request).
 Observe the accounts yaml to make sure the service for which you requested the increase on the particular region gets a status Update of "Manual_Action"

### DIFF
--- a/api/v1alpha1/account_types.go
+++ b/api/v1alpha1/account_types.go
@@ -272,7 +272,7 @@ func (a *Account) GetOptInRequestsByStatus(stati OptInRequestStatus) (int, OptIn
 func (a *Account) HasOpenQuotaIncreaseRequests() bool {
 	for _, accountServiceQuotas := range a.Status.RegionalServiceQuotas {
 		for _, v := range accountServiceQuotas {
-			if v.Status == ServiceRequestUnknown {
+			if v.Status != ServiceRequestCompleted || v.Status != ServiceRequestUnknown {
 				return true
 			}
 		}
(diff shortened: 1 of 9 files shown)


api/v1alpha1/account_types.go
```diff
@@ -270,7 +272,7 @@ func (a *Account) GetOptInRequestsByStatus(stati OptInRequestStatus) (int, OptIn
 func (a *Account) HasOpenQuotaIncreaseRequests() bool {
 	for _, accountServiceQuotas := range a.Status.RegionalServiceQuotas {
 		for _, v := range accountServiceQuotas {
-			if v.Status != ServiceRequestCompleted {
+			if v.Status != ServiceRequestCompleted || v.Status == ServiceRequestUnknown {
```
AlexVulaj, line 275: We can leave this line alone, as the second clause is already caught by the first one. That is, if `v.Status` is `ServiceRequestUnknown`, that will already be caught by `v.Status != ServiceRequestCompleted`

reedcort: yeah that should actually be `v.Status != ServiceRequestUnknown`   bc we want return false whenever all service request are Completed or Unknown

luis-falcon: Fixes

PR #821, opened 2024-04-10
[OSD-21095-fix-olm-template] fix olm template

reedcort:
# What is being added?
Fix field value for OLM template

Ref [OSD-21095](https://issues.redhat.com//browse/OSD-21095)


mjlshen (approved):
/lgtm

PR #831, opened 2024-09-04
[OSD-17066] Update Total Account Watcher to count both active and creating accounts

abyrne55:
# What is being added?
This PR addresses [OSD-17066](https://issues.redhat.com/browse/OSD-17066), which calls for AAO's Total Account Watcher to keep track of both active AWS accounts _and_ AWS accounts that are still being created. Currently, the Watcher only counts active AWS accounts, which has led to race conditions where too many accounts were allowed to be created.

The approach this PR takes is quite simple: in addition to counting the number of active accounts returned by `ListAccounts()`, AccountWatcher's `getTotalAwsAccounts()` now also counts the number of accounts returned by `ListCreateAccountStatus(States: ["IN_PROGRESS"])`. Both counts are added together and returned as one integer. We rely on AWS to only return "IN_PROGRESS" account creations by specifying such criteria in our ListCreateAccountStatus call. As for the other two possible [account creation states](https://github.com/aws/aws-sdk-go-v2/blob/service/organizations/v1.31.1/service/organizations/types/enums.go#L236): we assume that accounts marked "SUCCEEDED" are already returned by `ListAccounts()`, and that accounts marked "FAILED" are irrelevant.

This PR modifies almost all of AccountWatcher's unit tests such that goMock now expects a call to ListCreateAccountStatus. This PR also adds a few new unit test cases designed to simulate the race condition scenario mentioned in the ticket.

## Steps To Manually Test
1. Start the operator
1. Run the thing
1. Clean up the thing

Ref OSD-17066

### DIFF
--- a/pkg/awsclient/client.go
+++ b/pkg/awsclient/client.go
@@ -16,10 +16,11 @@ package awsclient
 import (
 	"context"
 	"fmt"
-	"github.com/aws/aws-sdk-go/service/account"
-	"github.com/aws/aws-sdk-go/service/account/accountiface"
 	"net/http"
 	"time"
+
+	"github.com/aws/aws-sdk-go/service/account"
+	"github.com/aws/aws-sdk-go/service/account/accountiface"
 
 	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/service/route53/route53iface"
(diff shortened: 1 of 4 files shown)


iamkirkbater:
Overall this LGTM - though I wonder if we _need_ to update the list type on the CRD? I only say this because when we have to update CRDs with AAO that's when we tend to have to do OLM Dances to get the deploys working. Otherwise LGTM.

I'm going to approve AND hold this, only so that it doesn't automatically merge the CRD change if that's something that needs to be addressed. Otherwise, feel free to remove the hold and deploy when ready :) 

/hold
/approve

abyrne55:
Thanks for the review @iamkirkbater! As much as I tried to avoid it (in the interest of keeping the PR small), `make container-generate` kept erroring-out until I added that list type. This seems to be in line with [the k8s docs](https://github.com/kubernetes/kube-openapi/blob/9e1beecbcb384a484fb5cbd41024f316d51fdfac/pkg/idl/doc.go#L21-L69)...
```
// ListType annotates a list to further describe its topology. It may
// have 3 possible values: "atomic", "map", or "set". Note that there is
// no default, and the generation step will fail if a list is found that
// is missing the tag.
```
...but it was also one of many fights I had with `make` while trying to get this PR to pass the CI tests, so it quite possible I'm misattributing the root cause here

abyrne55:
Seeing no further objections, I'm gonna let this merge
/unhold

PR #838, opened 2024-12-12
[OSD-18390] removed sampleVPCID var from Fedramp region intialization

reedcort:
# What is being added?
Fixing FedRAMP VPC cleanup by removing the sampleVPCID variable and relying on the AWS API to retrieve the VPC ID when it's time to clean up resources in a region.

## Steps To Manually Test
1. Start the operator
1. Run the thing
1. Clean up the thing

Ref OSD-18390


controllers/account/ec2.go
```diff
@@ -31,7 +31,6 @@ const T3INSTANCETYPE = "t3.micro"
 const T2INSTANCETYPE = "t2.micro"
 
 var sampleCIDR = "10.0.0.0/16"
-var sampleVPCID = ""
```
joshbranham, line 34: There is a comment on line 143 that can be removed/updated to reflect this var being removed

controllers/account/ec2.go
```diff
@@ -206,12 +204,11 @@ func (r *AccountReconciler) InitializeRegion(
 	successMsg := fmt.Sprintf("EC2 instance created and terminated successfully in region: %s", region)
 
 	// if vpcID exists then make sure we delete it
```
joshbranham, line 206: This could be updated to reflect the new behavior

controllers/account/ec2.go
```diff
@@ -628,8 +625,48 @@ func CreateEC2Instance(reqLogger logr.Logger, account *awsv1alpha1.Account, clie
 		}
 
 		// If fedramp, create subnet and set value for RunInstancesInput
-		if config.IsFedramp() {
-			subnetID, err := createSubnet(reqLogger, client, account, managedTags, customerTags, sampleCIDR, sampleVPCID)
+		if !config.IsFedramp() {
+			// Get a list of all VPCs with appropriate tag
+			result, err := client.DescribeVpcs(&ec2.DescribeVpcsInput{
```
joshbranham, line 630: Since we are doing this lookup twice now, might be easier to read as a standalone function to fetch the VPCs for a given config

controllers/account/ec2.go
```diff
@@ -628,8 +625,48 @@ func CreateEC2Instance(reqLogger logr.Logger, account *awsv1alpha1.Account, clie
...
+							aws.String(awsv1alpha1.ClusterClaimLinkNamespaceTagKey),
+						},
+					},
+				},
+			})
+			if err != nil {
+				reqLogger.Error(err, "Error while describing VPCs")
+				controllerutils.LogAwsError(reqLogger, "Error finding vpcID", nil, err)
+			}
+			subnetID, err := createSubnet(reqLogger, client, account, managedTags, customerTags, sampleCIDR, *result.Vpcs[0].VpcId)
```
joshbranham, line 669: We should first nil check these, I presume if no VPCs match then we get an empty slice back. For example, `if result != nil && len(*result.Vpcs) > 0 {}`

controllers/account/ec2.go
```diff
@@ -628,8 +625,48 @@ func CreateEC2Instance(reqLogger logr.Logger, account *awsv1alpha1.Account, clie
 		}
 
 		// If fedramp, create subnet and set value for RunInstancesInput
-		if config.IsFedramp() {
-			subnetID, err := createSubnet(reqLogger, client, account, managedTags, customerTags, sampleCIDR, sampleVPCID)
+		if !config.IsFedramp() {
+			// Get a list of all VPCs with appropriate tag
+			result, err := client.DescribeVpcs(&ec2.DescribeVpcsInput{
+				MaxResults: aws.Int64(5),
+				Filters: []*ec2.Filter{
```
joshbranham, line 632: Were you able to validate this query works for the fedramp VPC configuration?

joshbranham (changes requested):
Overall I agree with the approach to not rely on a global variable to know wether to cleanup or not a VPC.

Couple comments inline, additionally tests are failing due to new API calls being added.

controllers/account/ec2.go
reedcort: yeah I verified it

joshbranham (commented):
With my limited understanding of what the FedRamp process looks like, I think this looks good. When this merges, does it land in staging automatically? We should kick the tires on a FedRAMP account with the team there to make sure things get cleaned up and no unexpected side effects, and likely then promote this to prod after the code freeze?

/lgtm

reedcort:
yeah it lands in stage and int once its merged. I'll reach out to the FedRAMP team and have them create a new AWS account and have them validate that it works as expected.

PR #841, opened 2025-02-03
OSD-18390: Ensure Region Initialization Fails on Missing AMI

joshbranham:
Currently, when a region is being initialized, if we fail to lookup an AMI to use, or the instance type, the code would continue to execute and later fail on passing an empty string to AWS.

The function performing the initialization does return errors as well, but they are never read by the controller calling. Instead it spawns a goroutine for each region init, and uses channels to handle errors and info messages.

In the scenario where the AMI lookup fails, the error is sent back on the errors channel, and then that channel is closed, following by the goroutine continuing to run in the background until it eventually panics trying to send to the now closed error channel.

This causes the operator process to exit prematurely, leading to odd scenarios around cleaning up initialization VPCs that get leaked.

I also did a bit of code cleanup while I was in here.

## Steps To Manually Test
1. Start the operator with an invalid `ami-owner` and fedramp enabled
2. Create an `AccountClaim`, which will try to spawn a VPC and init
3. When the AMI is not found, it returns early, never creates the VPC, and marks the AccountClaim as failed

https://issues.redhat.com/browse/OSD-18390

### DIFF
--- a/controllers/account/ec2.go
+++ b/controllers/account/ec2.go
@@ -143,6 +143,28 @@ func (r *AccountReconciler) InitializeRegion(
 		return nil
 	}
 
+	// Attempt to gather data needed to launch the init EC2 instance. We do this here before FedRamp initialization,
+	// as they require a VPC to be created and if we hit an error here we want to bail before then to reduce possible
+	// cleanup.
+	instanceType, err := RetrieveAvailableMicroInstanceType(reqLogger, awsClient)
+	if err != nil {
+		determineTypesErr := fmt.Sprintf("Unable to determine available instance types in region: %s", region)
+		controllerutils.LogAwsError(reqLogger, determineTypesErr, nil, err)
+		ec2Errors <- regionInitializationError{ErrorMsg: determineTypesErr, Region: region}
+		return err
+	}
+	ami, err := RetrieveAmi(awsClient, amiOwner)
+	if err != nil {
+		retrieveAmiErr := fmt.Sprintf("Unable to find suitable AMI in region: %s", region)
+		controllerutils.LogAwsError(reqLogger, retrieveAmiErr, nil, err)
+		ec2Errors <- regionInitializationError{ErrorMsg: retrieveAmiErr, Region: region}
+		return err
+	}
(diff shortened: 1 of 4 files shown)


controllers/account/ec2.go
```diff
@@ -182,12 +185,14 @@ func (r *AccountReconciler) InitializeRegion(
 		determineTypesErr := fmt.Sprintf("Unable to determine available instance types in region: %s", region)
 		controllerutils.LogAwsError(reqLogger, determineTypesErr, nil, err)
 		ec2Errors <- regionInitializationError{ErrorMsg: determineTypesErr, Region: region}
+		return err
```
joshbranham, line 188: We also should return if we can't find an instance type for some reason

reedcort:
Tested it locally and everything looks good
/lgtm

controllers/account/ec2.go
```diff
@@ -377,18 +387,18 @@ func deleteFedrampInitializationResources(reqLogger logr.Logger, client awsclien
...
+func retrieveVpcs(reqLogger logr.Logger, client awsclient.Client, region string) (*ec2.DescribeVpcsOutput, error) {
 	// Make dry run to certify required authentication
 	_, err := client.DescribeVpcs(&ec2.DescribeVpcsInput{
 		DryRun: aws.Bool(true),
 	})
 
 	// If we receive AuthFailure, do not attempt to clean resources
-	if aerr, ok := err.(awserr.Error); ok {
+	var aerr awserr.Error
+	if errors.As(err, &aerr) {
```
iamkirkbater, line 398: Is there a specific reason that this is used over the previous code paradigm?

iamkirkbater (approved):
/approve

controllers/account/ec2.go
joshbranham: Primarily has to do with being able to assert error types even when errors are wrapped when using things like `fmt.Errorf("some error %w", err)`. https://go.dev/blog/go1.13-errors

joshbranham:
> /label tide/merge-method-squash

Well, because I edited this comment to fix the command it never picked it up and didn't squash :(

PR #851, opened 2025-04-15
Bump boilerplate 7.1.0

joshbranham:
# What is being added?
Bumping boilerplate to the latest version which includes linting changes. Each commit in this PR lays out the fixes for each newly introduced linting failure as part of `golangci-lint` v2.

The only fundamental change is changing case for error messages, so if we are string matching those in another system that could break. Otherwise everything is simple package naming and other things that could be auto fixed.


.golangci.yml
joshbranham, line 1: I am not sure why we have this file in repos. We run a _second_ linting target to use this afaict, which doubles the linting time. https://github.com/openshift/boilerplate/blob/119ea794fb0f2d0f931ab1cba04824bd470fe385/boilerplate/openshift/golang-osd-operator/standard.mk#L186-L189

For now I just migrated the file to appease the make target to succeed.

reedcort:
Tested locally by creating an account and accountclaim
/lgtm

PR #891, opened 2025-09-09
Bump to latest boilerplate and Golang 1.24.7

abyrne55:
This PR addresses [SREP-1765](https://issues.redhat.com/browse/SREP-1765), which calls for security-related dependency bumps


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/openshift/aws-account-operator
 
-go 1.24.4
+go 1.24.7
```
joshbranham, line 3: I think I did this in error last time, but this should be the minimum version of Go the code supports, then a `toolchain` directive should be used to force `go build` to install the version you want for say CVEs to be patched. I guess you could go either way, but if `1.24.7` is not available on the build/local system, I believe it will error on you.

PR #1049, opened 2026-07-13
fix: enforce namespace match on tenant-specified SecretRefs (ROSAENG-61301)

BATMAN-JD:
## Summary

- Fixes CVSS 7.7 confused-deputy vulnerability (ROSAENG-61301) where tenant-authored CRs (`AccountClaim`, `AWSFederatedAccountAccess`) could reference secrets in other namespaces via free-form `SecretRef` fields, allowing cross-namespace secret access through the operator's cluster-wide RBAC
- Adds `Validate()` methods on CR types to reject namespace mismatches early, plus defense-in-depth guards at every secret-access call site using the pattern `!= "" && != crNamespace` to preserve backwards compatibility with existing CRs
- Ensures deletion paths run before validation to prevent stuck finalizers on legacy CRs
- Switches prek golangci-lint hook to `language=system` to fix Go toolchain version mismatch

## Security Details

**Attack vector**: A tenant with `create` rights on `AccountClaim` or `AWSFederatedAccountAccess` could set `spec.awsCredentialSecret.namespace` (or similar fields) to `aws-account-operator`, causing the operator to load org-root AWS credentials and hand them back to the tenant.

**Fix approach**: Two layers of defense:
1. **Validation layer**: `Validate()` on CR types checks all SecretRef namespace fields match `metadata.namespace` — covers all code paths (BYOC, STS, non-CCS, fleet manager)
2. **Defense-in-depth**: Inline guards at every function that reads a tenant-specified secret (`getCCSClient`, `createIAMSecret`, `createIAMRoleSecret`, `addBYOCSecretFinalizer`, `removeBYOCSecretFinalizer`, `processFake`, FAA `GetClient`)

**Backwards compatibility**: All guards use `!= "" && != crNamespace` — empty namespace fields (common in existing CRs) pass through. Deletion finalizer cleanup returns `nil` instead of erroring for cross-namespace refs, preventing stuck resources.

## Test plan

- [x] Unit tests pass (`make test`)
- [x] API tests pass (`make test-apis`)
- [x] golangci-lint passes (zero issues)
- [x] `go build ./...` succeeds
- [x] New test cases for namespace mismatch on AccountClaim (BYOC, non-CCS, STS paths)
- [x] New test cases for AWSFederatedAccountAccess namespace validation
- [x] Existing test fixtures updated to use matching namespaces
- [ ] CI/PROW full test suite

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced namespace consistency for BYOC and AWS credential secret references; requests with mismatched namespaces now fail with clear namespace-mismatch errors.
  * Added validation guardrails to prevent incorrect secret provisioning/cleanup when an AccountClaim’s spec is invalid.
  * Improved AWS Federated Account Access reconciliation by validating earlier and performing more reliable deletion finalizer cleanup; validation failures now update CR status more consistently.
* **Tests**
  * Expanded unit test coverage for the new namespace-mismatch validation behavior in both AccountClaim and AWS Federated Account Access.

### DIFF
--- a/controllers/account/byoc.go
+++ b/controllers/account/byoc.go
@@ -297,6 +297,12 @@ func (r *AccountReconciler) getSTSClient(log logr.Logger, accountClaim *awsv1alp
 }
 
 func (r *AccountReconciler) getCCSClient(currentAcct *awsv1alpha1.Account, accountClaim *awsv1alpha1.AccountClaim) (awsclient.Client, error) {
+	// Defense-in-depth: block cross-namespace secret references
+	if accountClaim.Spec.BYOCSecretRef.Namespace != "" && accountClaim.Spec.BYOCSecretRef.Namespace != accountClaim.Namespace {
+		return nil, fmt.Errorf("BYOCSecretRef namespace %q does not match AccountClaim namespace %q",
+			accountClaim.Spec.BYOCSecretRef.Namespace, accountClaim.Namespace)
+	}
+
 	awsRegion := config.GetDefaultRegion()
 
 	// Get credentials
--- a/controllers/account/byoc.go
+++ b/controllers/account/byoc.go
@@ -297,18 +297,20 @@ func (r *AccountReconciler) getSTSClient(log logr.Logger, accountClaim *awsv1alp
 }
 
 func (r *AccountReconciler) getCCSClient(currentAcct *awsv1alpha1.Account, accountClaim *awsv1alpha1.AccountClaim) (awsclient.Client, error) {
-	// Defense-in-depth: block cross-namespace secret references
-	if accountClaim.Spec.BYOCSecretRef.Namespace != "" && accountClaim.Spec.BYOCSecretRef.Namespace != accountClaim.Namespace {
-		return nil, fmt.Errorf("BYOCSecretRef namespace %q does not match AccountClaim namespace %q",
-			accountClaim.Spec.BYOCSecretRef.Namespace, accountClaim.Namespace)
+	if err := accountClaim.Validate(); err != nil {
+		return nil, err
+	}
+
+	secretNamespace := accountClaim.Spec.BYOCSecretRef.Namespace
+	if secretNamespace == "" {
+		secretNamespace = accountClaim.Namespace
 	}
 
 	awsRegion := config.GetDefaultRegion()
 
-	// Get credentials
 	ccsAWSClient, err := r.awsClientBuilder.GetClient(controllerName, r.Client, awsclient.NewAwsClientInput{
 		SecretName: accountClaim.Spec.BYOCSecretRef.Name,
-		NameSpace:  accountClaim.Spec.BYOCSecretRef.Namespace,
+		NameSpace:  secretNamespace,
 		AwsRegion:  awsRegion,
 	})
 	if err != nil {
--- a/controllers/accountclaim/accountclaim_controller.go
+++ b/controllers/accountclaim/accountclaim_controller.go
@@ -143,14 +143,6 @@ func generateInlinePolicy(accountID string) (string, error) {
 	return string(policyJSON), nil
 }
 
-// validateSecretRefNamespace is a defense-in-depth check that prevents cross-namespace secret access.
-func validateSecretRefNamespace(ref awsv1alpha1.SecretRef, crNamespace string) error {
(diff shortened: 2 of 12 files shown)


controllers/account/byoc.go
```diff
@@ -297,6 +297,12 @@ func (r *AccountReconciler) getSTSClient(log logr.Logger, accountClaim *awsv1alp
 }
 
 func (r *AccountReconciler) getCCSClient(currentAcct *awsv1alpha1.Account, accountClaim *awsv1alpha1.AccountClaim) (awsclient.Client, error) {
+	// Defense-in-depth: block cross-namespace secret references
+	if accountClaim.Spec.BYOCSecretRef.Namespace != "" && accountClaim.Spec.BYOCSecretRef.Namespace != accountClaim.Namespace {
```
AlexSmithGH: If we've reached this stage without setting the `Namespace` spec on the BYOC secret ref, I wonder if it's better that we just default to the `accountClaim.Namespace`, given that it is also not an empty string.

controllers/accountclaim/accountclaim_controller.go
```diff
@@ -143,6 +143,14 @@ func generateInlinePolicy(accountID string) (string, error) {
 	return string(policyJSON), nil
 }
 
+// validateSecretRefNamespace is a defense-in-depth check that prevents cross-namespace secret access.
+func validateSecretRefNamespace(ref awsv1alpha1.SecretRef, crNamespace string) error {
+	if ref.Namespace != "" && ref.Namespace != crNamespace {
+		return fmt.Errorf("secret ref namespace %q does not match CR namespace %q", ref.Namespace, crNamespace)
+	}
+	return nil
+}
```
AlexSmithGH: Unless I'm missing something in the code, this is handled by `accountClaim.Validate()`. I believe this function is redundant. Anywhere we execute an action on a secret, we should just ensure `accountClaim.Validate()` has been issued before proceeding.

AlexSmithGH (changes requested):
A validation function already exists on `accountClaim`, we should just call that whenever we're accessing the secret instead of using dedicated logic in each controller.

controllers/account/byoc.go
AlexSmithGH: Also, we can just use `accountClaim.Validate()` here as well.

BATMAN-JD: Makes sense, I will update

controllers/accountclaim/accountclaim_controller.go
BATMAN-JD: This is a product of my prompt in plan mode as I wanted to make sure I was addressing the Jira thoughtfully and removing any security risks. This does look redundant, especially with the above requested changes now. This will clean up a lot of additions too

controllers/accountclaim/accountclaim_controller.go
```diff
@@ -225,7 +225,11 @@ func (r *AccountClaimReconciler) Reconcile(ctx context.Context, request ctrl.Req
 	}
 
 	if accountClaim.DeletionTimestamp != nil {
-		if accountClaim.Spec.FleetManagerConfig.TrustedARN != "" {
+		// Skip fleet-manager secret cleanup for invalid specs,
+		// but always proceed to handleAccountClaimDeletion to remove finalizers.
+		if err := accountClaim.Validate(); err != nil {
+			reqLogger.Info("skipping fleet-manager secret cleanup due to invalid spec during deletion", "error", err.Error())
+		} else if accountClaim.Spec.FleetManagerConfig.TrustedARN != "" {
```
AlexSmithGH: nit: Golang wants to be left aligned for idiomatic reading. If checking if the `accountClaim.Validate` is intended to early return and skip out of the rest of this function, it should just do that and not predicate on an `else` clause to do so. 

```suggestion
		if err := accountClaim.Validate(); err != nil {
			reqLogger.Info("skipping fleet-manager secret cleanup due to invalid spec during deletion", "error", err.Error())
			return reconcile.Result{}, r.handleAccountClaimDeletion(reqLogger, accountClaim)
		}
		
		if accountClaim.Spec.FleetManagerConfig.TrustedARN != "" {
```

BATMAN-JD: Appreciate the feedback. Yes must cleaner

PR #1053, opened 2026-07-22
fix: reduce memory allocation by deferring expensive operations in reconcile loops

BATMAN-JD:
## Summary

- **Account controller**: Move cheap state checks (IsFailed, IsInitializingRegions, IsReadyUnclaimedAndHasClaimLink, creation timeout, account limit) before ConfigMap fetch and AWS client creation. Every Account reconcile was instantiating 9 AWS service clients (IAM, EC2, Organizations, Route53, S3, STS, Support, ServiceQuotas, Account) even for accounts in terminal states that exit immediately.
- **Account controller**: Eliminate double ConfigMap fetch — `IsPayerAccount()` internally fetched the ConfigMap, then the reconciler fetched it again. Replaced with `IsPayerAccountFromConfigMap()` that takes the already-fetched ConfigMap.
- **AccountClaim controller**: Move `claimIsSatisfied` check before ConfigMap fetch and AWS work, with `DeletionTimestamp == nil` and `!BYOC` guards so deletion and BYOC reconciliation paths still work. BYOC claims are excluded because `handleBYOCAccountClaim` performs idempotent reconciliation (finalizer, support role ARN, IAM secret) on every pass — the original code never evaluated `claimIsSatisfied` for BYOC claims.
- **Boilerplate sync**: UBI base image bump (`9.8-1784092902` → `9.8-1784705586`), OWNERS_ALIASES update, and agentic SDLC pipeline refactored from inline to centralized `pipelineRef` in boilerplate repo.

Addresses ballooning memory consumption and OOM kills observed in staging/integration where the operator was allocating ~9 AWS HTTP clients per reconcile for accounts that didn't need them.

## Test plan

- [x] `go build ./...` — compiles cleanly
- [x] `golangci-lint run` — 0 issues
- [x] `go test ./...` — all packages pass
- [x] New unit tests for `IsPayerAccountFromConfigMap` (5 cases: match, no-match, nil configmap, missing field, whitespace)
- [x] CI/Prow: validate, lint, test, images, coverage, prek — all passed

## Follow-up

A separate Jira card tracks threading `context.Context` through ~198 `context.TODO()` call sites across the codebase. Four `//nolint:contextcheck` directives were added here to avoid a blast-radius refactor in a memory-optimization PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved account reconciliation by handling failed, initializing, unclaimed, and stalled accounts earlier.
  * Added safeguards to prevent protected payer accounts from being modified.
  * Improved account creation limit enforcement and reduced unnecessary processing.
  * Corrected claim completion behavior for deleting and BYOC account claims.

* **Chores**
  * Updated automated pull request validation pipeline configuration.
  * Refreshed project ownership aliases.


controllers/account/account_controller.go
```diff
@@ -121,10 +121,63 @@ func (r *AccountReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+
+	// Log accounts that have failed and don't attempt to reconcile them
+	if currentAcctInstance.IsFailed() && !currentAcctInstance.IsPendingDeletion() {
+		reqLogger.Info(fmt.Sprintf("Account %s is failed. Ignoring.", currentAcctInstance.Name))
+		return reconcile.Result{}, nil
+	}
+
+	// Detect accounts for which we kicked off asynchronous region initialization
+	if currentAcctInstance.IsInitializingRegions() && !currentAcctInstance.IsPendingDeletion() {
+		return r.handleAccountInitializingRegions(reqLogger, currentAcctInstance) //nolint:contextcheck // pre-existing function signature
```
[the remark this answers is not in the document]
BATMAN-JD: https://redhat.atlassian.net/browse/ROSAENG-62411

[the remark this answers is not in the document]
BATMAN-JD: created a Jira for separate work to take care of all ctx work. Out of scope for this PR

controllers/account/account_controller.go
```diff
@@ -121,10 +121,63 @@ func (r *AccountReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+	}
+
+	// Update account Status.Claimed to true if the account is ready and the claim link is not empty
+	if !currentAcctInstance.IsPendingDeletion() && !currentAcctInstance.IsBYOC() && currentAcctInstance.IsReadyUnclaimedAndHasClaimLink() {
+		return reconcile.Result{}, ClaimAccount(r, currentAcctInstance) //nolint:contextcheck // pre-existing function signature
+	}
+
+	// see if in creating for longer than default wait time
+	if !currentAcctInstance.IsPendingDeletion() && currentAcctInstance.IsCreating() && utils.CreationConditionOlderThan(*currentAcctInstance, createPendTime) {
```
dakotalongRH: One quick question - prior to this change we had a check here for !currentAcctInstance.IsBYOC(). The check was in the else (non-BYOC) branch, so BYOC accounts never hit it. Now it'll apply to all non-deleted accounts. Is this intentional?

BATMAN-JD: Yes I was worried what Claude was doing here too. But reran 3x to confirm functionality.
`Yes, intentional. The original !IsBYOC() exclusion was an artifact of the code living in the else branch after newBYOCAccount(), not a deliberate scoping decision. I traced the history — it was introduced in PR #316, which was a pure code organization refactor (moving checks into if/else blocks), not a behavioral change. newBYOCAccount() is IsBYOC() && (!HasState() || !IsClaimed()), so BYOC accounts that are IsCreating() (which sets state) would have HasState() == true — meaning newBYOCAccount() returns false and they'd fall into the else branch and hit this timeout anyway. The only BYOC accounts that skipped it were ones with no state yet, which by definition can't be IsCreating(). Net behavior is unchanged.`

BATMAN-JD: I have decided to put back in. May be out of scope, and would not hurt anything to leave in

BATMAN-JD:
Container test exited with code 137, reason OOMKilled

will try later...

PR #1055, opened 2026-07-28
fix: treat CloseAccount rate limit as transient, not a reuse failure

AlexSmithGH:
When close-on-release is enabled, deleting an AccountClaim triggers a CloseAccount call against the AWS Organizations API. That API is quota- limited, and closeAndDeleteAccount correctly defers with an exponential backoff, returning ErrRateLimited.

However, handleAccountClaimDeletion treated any non-conflict error from the finalizer as a reuse failure and flipped the account to state=Failed with a permanent Failed=True condition (never cleared due to UpdateConditionNever). A routine quota throttle therefore left accounts looking permanently failed and produced an optimistic-lock write race each reconcile.

Short-circuit on errors.Is(err, ErrRateLimited) alongside the existing IsConflict check: log and requeue so the finalizer retries once the backoff window elapses, without marking the account Failed.

Adds a reconcile-level test that pre-seeds an unexpired rate-limit annotation and asserts the claim keeps its finalizer and the account is not marked Failed.

# What is being added?

## Steps To Manually Test
1. Start the operator
1. Run the thing
1. Clean up the thing

Ref OSD-0000

## Summary by CodeRabbit

* **Bug Fixes**
  * Account reuse now automatically retries when AWS rate limits account closure or reset operations.
  * Account claims remain pending during rate limiting instead of being incorrectly marked as failed.
  * Improved handling ensures account status and failure indicators accurately reflect the temporary condition.

### DIFF
--- a/controllers/accountclaim/accountclaim_controller.go
+++ b/controllers/accountclaim/accountclaim_controller.go
@@ -3,6 +3,7 @@ package accountclaim
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


geowa4 (approved):
/lgtm

controllers/accountclaim/accountclaim_controller.go
```diff
@@ -719,6 +720,17 @@ func (r *AccountClaimReconciler) handleAccountClaimDeletion(reqLogger logr.Logge
 				return fmt.Errorf("account CR modified during reset: %w", err)
 			}
 
+			// A CloseAccount rate limit is expected and transient: when close-on-release
+			// is enabled and we exceed the AWS Organizations CloseAccount quota, the close
```
geowa4, line 724: Is this why this only happens to n test environments? Is our production volume that much lower for this pathway?

BATMAN-JD: From what I can tell, may be way off, but the ACCOUNT_LIMIT is so much higher in prod. 2100 vs 50. So these failed accounts that brick have an impact on the total accounts much sooner than it would in prod. Could be why we see it here

PR #1059, opened 2026-07-30
fix: prevent pool burst creation, zombie churn, and auto-cleanup

BATMAN-JD:
## Problem

The pool controller had several issues causing staging instability:

1. **Zombie churn cycle**: Accounts that hit the AWS account limit were left in NoState forever, requeueing every 5 minutes. The pool counted these as "unclaimed" so it thought the pool was satisfied, but they would never provision — blocking new healthy accounts from being created.

2. **Burst creation**: The pool created all needed Account CRs in rapid succession via the ownership watch re-reconcile, causing burst CreateAccount API calls that overwhelmed the account controller.

3. **Pending deletion miscounting**: Accounts pending deletion were still counted as "unclaimed", inflating the pool satisfaction check and preventing replacement accounts from being created.

4. **958 zombie Account CRs in staging**: Failed, pool-owned Account CRs with no AWS account ID accumulated since 2022. They are irrecoverable but inflate the Account CR list, adding overhead to every pool reconcile.

### Commit: cc94dba4 — Prevent pool burst creation and zombie churn cycle
- Pool satisfaction now uses `IsPendingFirstProcessing()` instead of counting all unclaimed accounts, excluding stuck/zombie CRs
- Pool satisfaction check uses `available + progressing + pending` instead of raw unclaimed count
- 30-second delay between Account CR creations prevents burst API calls

### Commit: 01c5c339 — Use Condition instead of failing accounts blocked by AWS limit
- Accounts blocked by the AWS limit get an observable `AccountPending` Condition (reason: `AWSAccountLimitReached`) instead of silently requeueing
- **Sets a Condition, not a State** — this is the load-bearing safety decision. Setting State would strand the account forever (never picked up for creation once limit clears, never reaped). Keeping State empty preserves the recovery path
- The account controller requeues every 5 minutes and will proceed normally once headroom opens up

### Commit: d0595247 — Exclude accounts pending deletion from pending count
- Accounts with a deletion timestamp are no longer counted as "unclaimed" in pool satisfaction
- Ensures the pool controller creates replacement accounts promptly during reuse cleanup

### Commit: 612989a2 — Garbage-collect zombie Account CRs on reconcile
- On reconcile, Failed Account CRs with no AWS account ID and pool ownership are automatically deleted
- Finalizers are stripped first to prevent the account controller's deletion path from trying STS assume role into a non-existent account
- On deploy to staging, this will automatically clean up all ~958 existing zombie CRs

### Pool-level limit gate (in cc94dba4)
- Pool controller checks `accountCount >= limit` before creating new Account CRs
- If at the limit, pauses and requeues after 5 minutes instead of creating CRs that would sit in NoState
- This is the missing cap that prevents unbounded CR creation when the org limit is exhausted

## Testing

- Unit tests added for pool satisfaction logic, pending deletion exclusion, zombie GC, and limit gate
- Zombie GC tests: verifies deletion of zombies (Failed + no AWS ID + pool-owned) and preservation of real failures (Failed + has AWS ID)
- Pool tests: NoState pending accounts satisfy pool, Failed accounts do not, mixed states work correctly, limit-reached blocks creation

## Follow-up work

1. **Leaked account cleanup** (separate PR): ~217 Failed accounts WITH AWS account IDs need investigation — these represent real AWS accounts that failed during setup but still exist in the organization
2. **Context threading** ([ROSAENG-62411](https://redhat.atlassian.net/browse/ROSAENG-62411)): Thread `ctx` through helper functions to replace `context.TODO()` — the zombie GC code uses a `//nolint:contextcheck` for `removeFinalizer` which doesn't accept context yet
3. **Pool controller label selectors**: Currently lists ALL Account CRs every reconcile with no label selector — should filter by pool ownership at the API level


controllers/account/account_controller.go
```diff
@@ -125,6 +125,23 @@ func (r *AccountReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 
 	// Log accounts that have failed and don't attempt to reconcile them
 	if currentAcctInstance.IsFailed() && !currentAcctInstance.IsPendingDeletion() {
+		// Garbage-collect zombie CRs: Failed, no AWS account, pool-owned.
+		// These were created by the pool controller but never provisioned in AWS
+		// (e.g. account limit was reached). They are irrecoverable and inflate
+		// the Account CR list, adding overhead to every pool reconcile.
+		if !currentAcctInstance.HasAwsAccountID() && currentAcctInstance.IsOwnedByAccountPool() {
```
AlexSmithGH, line 132: I'd consider wrapping this in a feature flag to enable at will in staging and not impact our other environments. I'm not totally sure if in all environments if this set of conditionals on an account CR should always lead to use removing finalizers and deleting the account.

AlexSmithGH (commented):
/hold
/lgtm

